Issue #2, opened 2025-03-11
Add Local Model Support via Ollama Integration


crmne:
**TL;DR:** Add support for running fully local models via Ollama.

## Background

While cloud models offer state-of-the-art capabilities, there are compelling cases for running models locally:

1. **Privacy & Compliance**: Keep sensitive data entirely on-premise
2. **Cost Control**: Eliminate ongoing API costs for high-volume applications
3. **Latency**: Remove network overhead for latency-sensitive applications
4. **Offline Operation**: Run AI features without internet connectivity

Ollama provides an excellent way to run models like Llama, Mistral, and others locally with a simple API that's compatible with our existing architecture.

## Proposed Solution

Add a new provider interface for Ollama that implements our existing abstractions:

```ruby
# Configuration
RubyLLM.configure do |config|
  config.ollama_host = "http://localhost:11434" # Default
end

# Usage remains identical to cloud models
chat = RubyLLM.chat(model: 'llama2')
chat.ask("What's the capital of France?")

# Or with embeddings
RubyLLM.embed("Ruby is a programmer's best friend", model: 'nomic-embed-text')
```

## Technical Details

For those looking to help implement this, you'll need to:

1. Create a new provider module in `lib/ruby_llm/providers/ollama/`
2. Implement the core provider interface methods:
   - `complete` - For chat functionality
   - `embed` - For embeddings
   - `api_base` - Returns the Ollama API endpoint
   - `capabilities` - Define model capabilities
3. Handle the payload formatting differences between Ollama and OpenAI/Claude

The PR should include:
- Provider implementation
- Configuration option for Ollama host
- Tests that can be run against a local Ollama instance
- Documentation updates

## Benefits

- **Cost efficiency**: Eliminate API costs for many use cases
- **Privacy**: Keep sensitive data local
- **Flexibility**: Mix and match local and cloud models in the same codebase
- **Performance**: Reduce latency for response-time sensitive applications

Mizokuiam:
Hey @crmne, this is a fantastic proposal!  Local model support via Ollama would be a huge win for `ruby_llm` users.  The benefits you've outlined – privacy, cost control, latency reduction, and offline capabilities – are all spot-on.

I really like the proposed configuration and usage pattern.  Keeping the API consistent with existing cloud models minimizes the learning curve for users.  The example snippet is clear and concise:

```ruby
# Configuration
RubyLLM.configure do |config|
  config.ollama_host = "http://localhost:11434" # Default
end

[...]

PR #6, opened 2025-03-11
Support OpenAI compatible providers through a base URL

ldmosquera:
Allow configuring an arbitrary HTTP endpoint for an OpenAI compatible API, which easily affords compatibility with any inference engines that provide it.

A few popular ones are name dropped in the docs for searchability; I suggest you also mention OAI compatibility in the readme, which will interest people looking for fully self-hosted solutions.

### DIFF
--- a/docs/guides/getting-started.md
+++ b/docs/guides/getting-started.md
@@ -33,6 +33,12 @@ RubyLLM.configure do |config|
 end
 ```
 
+You can optionally provide the base URL for the OpenAI API, through which you can point to OpenAI API compatible engines like [Ollama](https://github.com/ollama/ollama), [KoboldCPP](https://github.com/LostRuins/koboldcpp), [TabbyAPI](https://github.com/theroyallab/tabbyAPI), etc:
+
+```ruby
+  config.openai_base_url = "http://localhost:11434/v1"
+```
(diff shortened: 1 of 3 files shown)


ldmosquera:
Note that I only saw issue #2 (for adding proper Ollama support) after I created this PR; two things:
- obviously this one is for OpenAI compatible APIs instead, not the Ollama API proper
- this one does the bare minimum to allow pointing to a custom endpoint, with equally minimum functionality

IMHO this one still has value as an OAI compatible MVP, as it's minimally intrusive but it already affords access to basic self-hosted inference. Many other LLM clients also support this type of integration, sometimes in addition to Ollama.

crmne:
Hey @ldmosquera! Thanks for the PR, but I'd rather we skip the half-measure and build a proper Ollama provider that speaks directly to their native API instead. The compatibility layer is just going to cause headaches - when I built the Gemini provider, I started with their OpenAI compatibility layer and ran into a ton of issues. Plus, Ollama's layer doesn't even support `tool_choice`. If you're up for it, I'd love to see a PR for that instead.

ldmosquera:
Fair enough; I can start a draft PR for that one.

PR #10, opened 2025-03-12
Add Ollama as a supported provider

### LINKED ISSUE #2: Add Local Model Support via Ollama Integration

ldmosquera:
As mentioned in issue #2, Ollama support is valuable for users interested in self-hosted/offline inference.

This PR adds initial support for Ollama including chat completions, streaming and embeddings. No tool support yet, needs further investigation.

PR is a rough draft and will likely take some back and forth to get it merge-ready; more comments to follow.

Closes #2


ldmosquera:
This was done copying and adapting from the existing Gemini provider. Tool usage and media was excised for now; I still need to research what Ollama offers and what overlaps within this project. Also, `capabilities.rb` is full of placeholder stuff until we figure out sane defaults since Ollama allows running arbitrary models instead of a well known list.

---

Commit https://github.com/crmne/ruby_llm/pull/10/commits/9b387c1a2bb3b4bac345306c29547f667f1f2e75 disables all providers by default in `models.refresh!` unless they are explicitly configured.

The intention is to be able to work on Ollama (or any other specific single provider) without being forced to provide valid keys for EVERY provider, also incurring live calls.

As mentioned in the commit message, this might be too intrusive for the intended usage of this project, so an alternative is to only do this when given a specific env var that puts ruby-llm into such a "default offline" mode.

This might interfere with tests and/or the models update rake task, none of which I touched yet since they also require valid keys for all providers.

---

As for tests, I'd like some guidance into how to implement unit testing and eventually integration testing specifically for Ollama, in a way that does not require configuring and using all APIs (and attending cost).

Since Ollama does not come with default models, I suggest a specific test suite that first ensures that a tiny model is downloaded via the API.

lib/ruby_llm/provider.rb
```diff
@@ -125,11 +125,21 @@ def to_json_stream(&block) # rubocop:disable Metrics/MethodLength
               RubyLLM.logger.debug "Accumulating error chunk: #{chunk}"
             end
           else
-            parser.feed(chunk) do |_type, data|
-              unless data == '[DONE]'
-                parsed_data = JSON.parse(data)
-                block.call(parsed_data)
+            content_type = env.response_headers['content-type']
```
ldmosquera, line 128: I'm not super sure about this; as far as I could see, Ollama uses newline delimited JSON lines rather than standard event streams.

Of the API providers, I only have a Gemini key and it does work after this commit (both streaming and synch) so this might be correct.

jm3:
was just about to file an issue for this — awesome!

crmne:
Two new features to consider in your provider implementation:

   ```json
   "claude-3-5-sonnet": {
   }
   ```

   ```ruby
   ```

ldmosquera:
In general, I understand this can be used to specify a provider when the model is served by more than one provider.

But for that, we have to know for sure that the provider supports that model, and yet in the case of Ollama, there are no guarantees that a given server will have any models installed, thus we can't hardcode any static assumptions into the gem. Rather, the list of models would only be available after querying the server at runtime.

I realize this deviates from the general pattern of the gem which aims to provide a good out of the box experience for well known API providers, but by definition Ollama is not that. It can be used with arbitrary models even beyond Ollama's official library - you can pull from HuggingFace or even create your own models with arbitrary names.

I still think supporting Ollama within reasonable limits is a valuable addition to the gem; we just have to figure out those limits.

ldmosquera:
> Commit 9b387c1 disables all providers by default in models.refresh! unless they are explicitly configured.
> As mentioned in the commit message, this might be too intrusive for the intended usage of this project, so an alternative is to only do this when given a specific env var that puts ruby-llm into such a "default offline" mode.

Any guidance on this?

> As for tests, I'd like some guidance into how to implement unit testing and eventually integration testing specifically for Ollama, in a way that does not require configuring and using all APIs (and attending cost).

And on this?

I didn't consider it originally, but I see now that the test suite has no support for being run without valid keys for all providers, which is a barrier for me.

Is there any chance that the test suite can be mocked as much as possible (requests for metadata, etc) such that only tests that NEED to run API inference fail? This is the only way I can see to mitigate the key requirement problem while still running as much of the test suite as possible.

Another option is to create an entirely separate suit of tests just for the Ollama provider, but the reduced integration test coverage is obviously problematic even on paper.

tpaulshippy:
Because of the VCR support I was able to get most if not all the tests to pass (with empty strings in all the API key environment variables). Wish the tests did not require these, but it's an easy fix.

ldmosquera:
Tests added, but a few provisional hacks are still in place and will likely need revisiting:

- https://github.com/crmne/ruby_llm/pull/10/commits/37c5fd51da5f11b5c8b386b737cf5c856a941364
- https://github.com/crmne/ruby_llm/pull/10/commits/11eedd88da8d59aa6832fcd377303e1c887d4b5b (new since force push)
- https://github.com/crmne/ruby_llm/pull/10/commits/4f48aae40ad50f0d5f5512b2d04df928265f5f15 (new since force push)

Tests pass when relying on VCR cassettes.

crmne:
Valid point @ldmosquera! I'm working on a solution, will commit it on `main` soon.

crmne:
Added configuration requirements handling in 75f99a1edc08ffe0eac9b85cdae33d633b5482e8

Each provider now specifies what configuration is required via a simple `configuration_requirements` method (you will need to implement this in your main provider file) that returns an array of config keys as symbols. The Provider module uses this to:

1. Determine if a provider is properly configured
2. Throw an error if you're trying to use that provider without configuration
3. Include ready-to-paste configuration code in the error message
3. Skip unconfigured providers during model refresh while preserving their models

Example of the new error messages:
```ruby
RubyLLM::ConfigurationError: anthropic provider is not configured. Add this to your initialization:

RubyLLM.configure do |config|
  config.anthropic_api_key = ENV['ANTHROPIC_API_KEY']
end
```

spec/ruby_llm/providers/ollama/ollama_spec.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'dotenv/load'
+
+RSpec.shared_context 'with configured local Ollama server as sole provider' do
+  before :all do
+    RubyLLM.configure do |config|
+      # NOTE: other tests will need to ensure relevant providers are enabled as needed
+      RubyLLM::Provider.disable_all_providers
```
ldmosquera, line 10: This is better than yesterday (because it now handles all current and future providers), but it does mean that each test will need to ensure relevant providers are enabled, to ensure they work regardless of test run order.

spec/ruby_llm/providers/ollama/ollama_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+  end
+
+  describe 'embeddings' do
+    let(:test_text) { "Ruby is a programmer's best friend" }
+    let(:test_texts) { %w[Ruby Python JavaScript] }
+    let(:model) { 'snowflake-arctic-embed:22m' }
+
+    it 'can handle a single text' do # rubocop:disable RSpec/MultipleExpectations
+      # FIXME: need a provider: param just like in chat()
+      # embedding = RubyLLM.embed(test_text, model: model, provider: 'ollama')
```
ldmosquera, line 67: It would be ideal if `embed(model: 'MODEL', provider: 'PROVIDER')` is possible, analogous to `chat()`, so that provider tests can ensure they hit their own models regardless of future name clashes.

crmne (changes requested):
The implementation path is straightforward - Ollama is just another provider that needs to:

1. Implement the standard provider interface
2. Override specific methods where needed 
3. Use the existing test infrastructure
4. Get documented like other providers

I think the only change w.r.t. to other providers is that in the documentation we should add that refreshing models should be done prior to using the API.

Happy to help out. The goal is making Ollama work cleanly within RubyLLM's existing patterns.

lib/ruby_llm/provider.rb
```diff
@@ -225,6 +237,14 @@ def providers
...
         providers.select { |_name, provider| provider.configured? }.values
       end
+
+      def disable_all_providers
+        providers.each_value do |provider|
+          provider.configuration_requirements.each do |key|
+            RubyLLM.config.send "#{key}=", nil
+          end
+        end
+      end
```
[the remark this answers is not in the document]
ldmosquera: I put this in because when running all tests, `openai_api_key` was still set by the time I did `models.refresh!` in this provider's test, and then I noticed a 403 response from OpenAI recorded in VCR.

So I figured I needed to systematically ensure no other providers were configured to avoid this, otherwise this will happen every time the test suite changes and more providers end up unexpectedly configured when this provider's spec does `refresh!`.

Any ideas on how to solve unwanted outward requests without relying on a method like above?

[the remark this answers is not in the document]
crmne: `unset OPENAI_API_KEY` or comment the line in `.env`. You name it.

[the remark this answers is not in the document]
ldmosquera: Simple enough, though it does not cover this future problem:

> this will happen every time the test suite changes and more providers end up unexpectedly configured when this provider's spec does refresh!

But that's not a today problem; going with `unset OPENAI_API_KEY`.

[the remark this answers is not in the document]
ldmosquera: I'm moving Ollama's specs into the existing files; in doing so, I can almost fix this by doing `refresh!` before other providers' keys are set to the mock 'TEST' value:

``` ruby
RSpec.shared_context 'with configured RubyLLM' do
  before do
    RubyLLM.configure do |config|
      # NOTE: to ensure relevant models are pulled into your local server, do
      #   bundle exec rake ollama:install_models_for_specs
      config.ollama_api_base_url = ENV.fetch('OLLAMA_API_BASE_URL', 'http://localhost:11434')

      # FIXME: this clobbers the model list from models.json and results in all tests except for ollama's to fail
      RubyLLM.models.refresh!

      config.openai_api_key = ENV.fetch('OPENAI_API_KEY', 'test')
      config.anthropic_api_key = ENV.fetch('ANTHROPIC_API_KEY', 'test')
      config.gemini_api_key = ENV.fetch('GEMINI_API_KEY', 'test')
      config.deepseek_api_key = ENV.fetch('DEEPSEEK_API_KEY', 'test')
      config.max_retries = 50
    end
  end
end
```

BUT this results in `refresh!` clobbering the models loaded from `models.json` and thus only Ollama's being known, failing most tests.

We would need a way to refresh models for a single provider without clobbering all the others.

lib/ruby_llm/providers/ollama/chat.rb
```diff
@@ -0,0 +1,86 @@
...
+          }
+
+          if block_given?
+            payload[:stream] = true
+            stream_completion(model, payload, &block)
+          else
+            payload[:stream] = false
+            generate_completion(model, payload)
+          end
+        end
```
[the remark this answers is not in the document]
ldmosquera: As noted on the PR first comment, I copied the Gemini provider and adapted from there; maybe it was a bad pick since I see now the OpenAI provider is quite leaner. Fixed by emulating that instead (and re-adapting to the new streaming logic).

lib/ruby_llm/aliases.rb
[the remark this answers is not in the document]
ldmosquera: Got it, dropping this commit.

docs/guides/getting-started.md
```diff
@@ -33,6 +33,16 @@ RubyLLM.configure do |config|
...
+Alternatively, you can point to an Ollama instance; here it is configured to the most common address when Ollama is installed locally:
+```
+RubyLLM.configure do |config|
+  config.ollama_api_base_url = 'http://localhost:11434'
+end
+
+# Needs to be called to populate Ollama models before using any
+RubyLLM.models.refresh!
+```
+
```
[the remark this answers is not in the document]
ldmosquera: Good call; I moved the `models.refresh!` note to the using section.

lib/ruby_llm/provider.rb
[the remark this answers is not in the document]
ldmosquera: Oops, I failed to notice the logic change from yesterday and `refresh!` does exactly that now. The remaning problem was that `shared_context -> before` runs before each test instead of before the entire suit as I intended.

The next problem is that `smollm` is too ~~dumb~~smol to know who Matz is :facepalm:

spec/spec_helper.rb
```diff
@@ -82,16 +82,13 @@
...
+  config.before(:suite) do
     RubyLLM.configure do |config|
-      # NOTE: to ensure relevant models are pulled into your local server, do
-      #   bundle exec rake ollama:install_models_for_specs
       config.ollama_api_base_url = ENV.fetch('OLLAMA_API_BASE_URL', 'http://localhost:11434')
-
-      # FIXME: this clobbers the model list from models.json and results in all tests except for ollama's to fail
+      # needs to run when ONLY Ollama is configured and before any others are configured
+      # FIXME: RubyLLM.models.refresh!(provider: 'ollama') would be a cleaner solution
       RubyLLM.models.refresh!
```
ldmosquera, line 92: After yesterday's models logic change in main, doing `refresh!` when ONLY Ollama is configured fixes both the bogus outgoing calls and the vanishing models.json information problems.

This is a provisional commit; I'm sure there's a better way to organize this, but it exemplifies a solution without the need for `RubyLLM::Provider.disable_all_providers` (which was undone).

spec/fixtures/vcr_cassettes/chat_basic_chat_functionality_smollm_135m_can_handle_multi-turn_conversations.yml
```diff
@@ -0,0 +1,61 @@
...
+      - application/json; charset=utf-8
+      Date:
+      - Wed, 26 Mar 2025 17:20:39 GMT
+      Transfer-Encoding:
+      - chunked
+    body:
+      encoding: UTF-8
+      string: '{"model":"smollm:135m","created_at":"2025-03-26T17:20:39.40000571Z","message":{"role":"assistant","content":"Ruby''s
+        creator is unknown, but some theories suggest that she might have been a member
+        of a team or group of women who worked on the project. Here are a few examples:\n\n1.
```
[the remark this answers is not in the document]
ldmosquera: Settled on [llama3.1:8b](https://ollama.com/library/llama3.1) which gets this right (ish - see below)

lib/ruby_llm/providers/ollama/streaming.rb
```diff
@@ -13,12 +13,12 @@ def stream_url
...
-            # from on_data with the complete response as an Array.
+            # FIXME: on connection close, on_data is called with the entire response so far as one chunk
+            # consisting of an Array of unparsed JSON strings (ie. an array of lines).
             #
-            # It is skipped here, but this smells a bit; this method shouldn't
-            # need to be overridden just for this. Will look into it
-            done = data.is_a?(Array) || data['done']
+            # It is detected and skipped here, but this is a code smell;
+            # likely standard SSE behavior which shouldn't be happening for NDJSON streaming.
+            done = data.is_a?(Array) && data.first.is_a?(String) && data.first[0] == '{'
```
ldmosquera, line 21: Code smell here, but I see no obvious way to avoid it.

lib/ruby_llm/providers/ollama/chat.rb
```diff
@@ -12,6 +12,11 @@ def completion_url
         end
 
         def render_payload(messages, tools:, temperature:, model:, stream: false)
+          # Heuristic: if temperature is set to default and tools are in play,
+          # use a low temperature to reduce risk of small models ignoring available
+          # tools in favor of making up imaginary API calls
+          temperature = 0.1 if tools.any? && temperature == RubyLLM::Chat::DEFAULT_TEMPERATURE
```
ldmosquera, line 18: Another hack that touches core code.

I can cange the "if temperature is set to default" part to "if temperature is set over 0.5" to dispense with the core hack.

spec/ruby_llm/chat_spec.rb
```diff
@@ -28,7 +27,20 @@
...
+        first =
+          if model =~ /llama/
+            # HACK: provisional code just to exemplify a problem with the original question.
+            # llama3.1:8b takes "Ruby's creator" to refer to some fictional character,
+            # but apparently "creator of Ruby" sounds less about a person and gives it a little helpful push.
+            #
+            # Ideally the question could be changed and other cassettes re-recorded;
+            # otherwise this can probably be solved with a bigger model but it will make the repeating Ollama tests
+            # more computationally expensive for everyone.
+            chat.ask('Who was the creator of Ruby?')
```
ldmosquera, line 39: @crmne I can reliably reproduce llama3.1:8b thinking "who was Ruby's creator" is about a fictional character, whereas it gets the above form mostly correct (takes a few tries; L3 8B is nowhere near a frontier model).

This is provisional; it would be great if you could change the question and refresh cassettes. Failing that, I can look for a stronger model (but it will be heavier).

lib/ruby_llm/providers/ollama/tools.rb
```diff
@@ -48,6 +48,26 @@ def parse_tool_calls(tool_calls)
...
+        # HACK: Llama3.x yields this in the proper tool_calls response field,
+        # but some other models return it as markup inside the text response.
+        #
+        # Unfortunately said other models are all over the place when it comes
+        # to sticking to a format so this doesn't cover all edge cases.
+        def preprocess_tool_calls(response_data)
+          # Move <toolcall>JSON</toolcall> markup from inside the text to its
+          # specific field in the response where it should be.
+          # https://github.com/ollama/ollama/blob/main/docs/api.md#chat-request-with-tools
+
```
ldmosquera, line 60: This is a bit of an abomination to attempt to mitigate the problems of attempting tool calling on smaller models; both their training and Ollama support is a bit undercooked.

Llama 3 does not seem to require this, but others do and it would be great to still support them. I could add a unit test excercise the hairy logic.

ldmosquera:
Updated TODOs from my end:

- [x] the call to refresh! with just the Ollama provider configured results in only its models being known (clobbering those from models.json), breaking everything
- [ ] some NDJSON streaming weirdness, on it -- see above
- [ ] capabilities.rb is mostly stubs/placeholders; on it -- not much to do here except hardcode values for a few common models
- [x] tool calling
- [x] vision

lib/ruby_llm/providers/ollama/tools.rb
ldmosquera: Unit test added: https://github.com/crmne/ruby_llm/pull/10/commits/a3635b5fc614f09832cb61b1090d8e236012a2e4

ldmosquera:
Merged to main and solved conflicts.

Echoing the commit message in https://github.com/crmne/ruby_llm/pull/10/commits/ba56fa245d9648aa2ebb13ea448667b9b957dfb3 for visibility:

> All cassettes for Ollama are up to date; I manually copied the new
> Bedrock responses for model listings.
> 
> VCR data will be synthetic until someone can run tests with all
> providers properly enabled, including a local Ollama server having run
> the rake task to get test models.

frimmy:
@crmne @ldmosquera just dropping appreciation! 🙇 🙏

lib/ruby_llm/providers/ollama/chat.rb
ldmosquera: Nevermind, decided against it and reverted.

crmne:
@ldmosquera I'd love to see Ollama support merged soon. Do you think you can resolve the conflicts? Once you're ready set this PR as ready for review and I'll take a look.

ldmosquera:
Ready for review; I imagine there will be some more back and forth before it's mergeable. Apologies and thanks in advance :grimacing:

ldmosquera:
Merged main; diff to yesterday [here](https://github.com/ldmosquera/ruby_llm/compare/ollama-provider-v6-2025-04-17../ollama-provider/?diff=split&w=).

khasinski:
Just curious - couldn't it be easier to use OpenAI's compatibility layer in Ollama? https://ollama.com/blog/openai-compatibility 

The same worked for OpenRouter, so maybe it's also the case here.

crmne:
Hey @khasinski I'm pondering the same. The reason why I preferred a true Ollama implementation are summed by Ollama themselves:

> This is initial experimental support for the OpenAI API. Future improvements under consideration include:
> 
> Embeddings API
> Function calling
> Vision support

However, we could simply test that and roll it out as OpenAI compatible if everything works.

Case in point OpenRouter claims that they don't support PDFs, but you can check out the tests in RubyLLM `main` and they do!

crmne:
Hey @ldmosquera,

Massive thanks for tackling Ollama and the serious effort on this PR. It shows real dedication.

Stepping back though, I'm opting for Ollama's OpenAI compatibility layer instead. I've confirmed the core chat works cleanly using our existing OpenAI provider: https://github.com/crmne/ruby_llm/blob/main/lib/ruby_llm/providers/ollama.rb

**Why?** Simplicity is key for RubyLLM. This means *way* less code, making it easier to maintain. I also found some elegant solutions for dynamic models, testing, and configuration in the meantime without the burden of back and forth.

For example now Providers have a `local?`, the model updater doesn't update local providers, the release CI doesn't remove the VCRs for Ollama, when using local providers you don't need `assume_model_exists` because we don't store them in `models.json`, etc.

Yes, Ollama's own blog post called this compatibility "experimental", but I tested chat, streaming, vision, tool calls, and everything works.

We can revisit this decision if native becomes clearly necessary.

I know this isn't the outcome you wanted after your work, and I'm sorry about that. Your exploration was valuable – it underscored the very complexities we're avoiding for now.

Closing this PR in favor of the OpenAI compatibility approach in 89c371a1e7ff600991a320c25b8c4f2777343722.

Thanks again for the contribution and understanding.

ldmosquera:
Some earlier feedback would have been great, but still, I appreciate all the reviewing, and it's never a waste when you learn along the way.

Happy to see self hosted inference support in this library at last!

crmne:
@ldmosquera I understand your frustration - hell I would have felt the same! Unfortunately I only got to try this fully right now, as it was heavily inspired by `openai_api_base` that I released in 1.2.0.

You and @khasinski for OpenRouter deserve to be credited for your work though, so I'll do that in the next release notes.

Issue #9, opened 2025-03-12
Custom URL support


zachlatta:
First - wow! Thanks for making this!

Would love the ability to override the API base URL for any given model. I didn't see something in the docs around this.

This would enable support for Llama and proxies.

crmne:
Hi @zachlatta and thank you!

What other providers apart from Ollama would you like to support by changing the base url?

We already have an issue to support Ollama https://github.com/crmne/ruby_llm/issues/2 and we're not gonna do it through the OpenAI compatibility layer as it doesn't support everything, and I've had a myriad of issues with other OpenAI compatibility layers in the past. Looking at you Gemini.

zachlatta:
The two others that immediately come to mind for me are LM Studio and Groq!

________________________________

jyunderwood:
This makes sense, but I'm not sure how you get around the need for configuring the url for provider like Azure OpenAI Service where your deployed models are at a unique endpoint for your resource. 

We use Azure OpenAI at work for its unified billing and access control, and I would like to be able to use this library with our existing deployments.

aspires:
Hopping in here with another use case. Fastly's LLM semantic caching product, [AI Accellerator](https://www.fastly.com/products/ai), operates as a URL/API key swap within code to initiate middleware caching with [Python and Node.js libraries](https://docs.fastly.com/en/guides/about-the-ai-accelerator-page#supported-llms). 

Disclaimer: I work for Fastly. I'd love to feature a Ruby example within the product. If there's an easy drop-in integration option, we'll find a way to feature it.

crmne:
I still believe in our model-first approach rather than taking the "swap the URL in the config" shortcut. Each provider deserves proper first-class implementation with correct model routing, capabilities, and pricing info.

That said, Azure OpenAI presents a legitimate use case where URL configuration is inherent to the provider. I'll need to think about the right design to accommodate this without compromising our principles - maybe extending our OpenAI provider to handle Azure's deployment-based approach. Are pricing and capabilities the same as the original OpenAI models?

jyunderwood:
There are probably enough differences in constructing the request that it would make sense for Azure OpenAI to be it's own provider. 

However, the capabilities of Azure's hosted OpenAI APIs are the same as OpenAI's, and their pricing is identical, as far as I'm aware. For example, o3-mini costs $1.10 per million tokens on both platforms. 

And the responses are the same structure.

**Some differences:**

[...]

PR #27, opened 2025-03-16
Add OpenAI URL override and support for custom request headers

aspires:
Relates to https://github.com/crmne/ruby_llm/issues/9

**Preface**
Opening early to discuss whether this is a welcome addition, the architectural decisions, documentation implications, and whether expanding this URL config override to other LLMs within the gem's scope makes sense. I'm committed to maintaining the developer experience simplicity of the project's stated mission, so I'm on board with redoing the initial approach to fit in with longer-term priorities (or closing outright!). 

---
**Goal**
This commit introduces the configuration primitives necessary for semantic caching proxy services for OpenAI requests. Specifically, the option to override the OpenAI root API URL path and the ability to add custom headers to the request payload in the form of a hash. I'm a Fastly employee, and I used [our semantic caching product](https://docs.fastly.com/en/guides/about-the-ai-accelerator-page#supported-llms) as a test implementation. 

---
**In-Progress Disclaimer**
This branch **does not yet** introduce tests and should not be merged until a subsequent commit lands with proper testing.

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -12,6 +12,8 @@ RubyLLM.configure do |config|
   config.anthropic_api_key = ENV.fetch('ANTHROPIC_API_KEY', nil)
   config.gemini_api_key = ENV.fetch('GEMINI_API_KEY', nil)
   config.deepseek_api_key = ENV.fetch('DEEPSEEK_API_KEY', nil)
+
+  config.openai_base_url_override = ENV.fetch('OPENAI_BASE_URL', nil)
 end
 
 IRB.start(__FILE__)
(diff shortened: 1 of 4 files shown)


crmne:
I appreciate the thought that went into this PR, but I'd rather not go down the path of generic URL overrides and custom headers. That's essentially creating a backdoor that undermines our provider abstraction.

Instead, I'd much prefer we implement a proper `RubyLLM::Provider::Fastly` that handles the semantic caching natively. This would give us:

1. A clean implementation that doesn't expose HTTP details
2. The ability to properly document the Fastly integration
3. Proper pricing and capabilities integration

The "override URL and add headers" approach feels like a shortcut that would introduce significant technical debt. We'd be passing responsibility for correct implementation to users rather than handling it properly in the library.

Since you work at Fastly, you're in a perfect position to implement this provider the right way. It might be slightly more code than URL overrides, but it would be a much more robust and maintainable solution in the long run.

Let's close this PR and create a new issue specifically for adding Fastly provider support. That's a contribution I'd be excited to see!

PR #28, opened 2025-03-16
Fix tool `name` method to strip unsupported characters for OpenAI

stevegeek:
Updates the `name` method in the `Tool` class to strip unsupported characters and replace them with '-' characters as OpenAI only supports tool names with `/a-zA-Z0-9_-/+`


lib/ruby_llm/tool.rb
```diff
@@ -50,6 +50,7 @@ def parameters
 
     def name
       self.class.name
+          .gsub(/[^a-zA-Z0-9_-]/, '-')
```
nbgoodall, line 53: Adding `.unicode_normalize(:nfkd).encode('ASCII', replace: '')` would gracefully handle non-ASCII characters :)

crmne: Love this method!

stevegeek: Sure! Nice way to transliterate to ASCII, though would still need to handle the fact that the supported character set is very small (literally that regex), eg most symbols need converting to hyphens or underscores and accented chars not allowed.

stevegeek:
Ok will find some time soon to update to transliterate appropriately!

spec/ruby_llm/tool_spec.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'ruby_llm/tool'
```
crmne, line 4: This shouldn't be required because `spec_helper.rb` imports RubyLLM already and Zeitwerk does the rest.

spec/ruby_llm/tool_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  describe '#name' do
+    it 'converts class name to snake_case and removes _tool suffix' do
+      class SampleTool < RubyLLM::Tool; end
+      expect(SampleTool.new.name).to eq('sample')
+    end
+
+    it 'replaces unsupported characters with -' do
+      class SampleTòol < RubyLLM::Tool; end
+      expect(SampleTòol.new.name).to eq('sample_t-ol')
+    end
```
crmne, line 16: This becomes unnecessary because of the next test

crmne (changes requested):
Pretty good overall. Just waiting on this `.unicode_normalize(:nfkd).encode('ASCII', replace: '')` and the other few comments to be addressed.

spec/ruby_llm/tool_spec.rb
stevegeek: true!

spec/ruby_llm/tool_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      class SampleTool123 < RubyLLM::Tool; end
+      expect(SampleTool123.new.name).to eq('sample_tool123')
+    end
+
+    it 'strips :: for class in module namespace' do
+      module TestModule
+        class SampleTool < RubyLLM::Tool; end
+      end
+      expect(TestModule::SampleTool.new.name).to eq('test_module--sample')
+    end
```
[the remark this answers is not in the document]
stevegeek: Testing the handling of tool classes defined inside a module I would retain as its exactly what caused me to come across the bug in the first place

crmne:
Looks like `overcommit` was not installed and therefore `rubocop` complained. Here's how to fix both issues:

```
overcommit --install
rubocop -A
```

stevegeek:
doh sorry

crmne:
no problem! :)

crmne:
Investigating: https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/

crmne (approved):

stevegeek:
@crmne ok sure, if I can be of any help, let me know.

crmne:
I was checking into the permissions for running tests safely, and I think we need to take a different approach. 

Instead of trying to dance around GitHub Actions security model with labels and pull_request_target, let's do this right by implementing VCR to record our API interactions. This will let us run the test suite against PRs without exposing our real API keys, and as a bonus make the tests faster and more reliable.

I'm going to work on adding VCR support before proceeding with this PR. Doing proper testing without compromising security is worth the small delay.

Will get back to you soon with the updated approach.

Issue #8, opened 2025-03-12
Add OpenRouter Support


crmne:
## Context
I'm hitting Anthropic rate limits constantly with my own usage, and several community members have requested OpenRouter integration. This would allow access to a wider range of models through a single API key while maintaining RubyLLM's unified interface.

## Benefits
- Single API key to access models across providers
- Potential cost savings through OpenRouter's pricing
- Simplified rate limit management
- Access to exclusive models not available through direct provider integrations
- Fallback capabilities when primary providers are at capacity

## Implementation Considerations
- OpenRouter largely follows OpenAI's API structure, so we can likely adapt our existing OpenAI provider implementation
- Need to handle model ID mapping/translation to keep the model selection experience consistent
- Should implement proper error handling for OpenRouter-specific cases
- Will need to update the Models registry to include OpenRouter-accessible models

## Scope
Initial implementation should focus on:
- Chat completion support (highest priority)
- Embeddings support
- Model listing
- Full streaming support
- Tool use

Image generation through DALL-E can be a second phase.

## Next Steps
- Research OpenRouter API documentation
- Create a new Provider module for OpenRouter
- Implement proper model discovery and capabilities
- Test with complex conversations, tools, and streaming

crmne:
Closed by d698dd1c255a2f74cea423c97bdf5d2d9b2b1d30

PR #29, opened 2025-03-16
Add OpenRouter as a supported provider

khasinski:
Addresses #8 

This adds support for OpenRouter.

You can use it by configuring the provider with:

```ruby
RubyLLM.configure do |config|
  config.openrouter_api_key = 'your key'
end
```

Then calling models by their OpenRouter ids (it might be wise to prefix those with openrouter or something else if there is a possibility there will be more multi-model providers in the future)

```ruby
chat = RubyLLM.chat(model: "google/gemini-2.0-flash-lite-preview-02-05:free")
chat.ask "What's the best way to learn Ruby?"
```

OpenRouter doesn't support embedding models and image generation so those are not implemented here.

Getting model details requires multiple calls to OpenRouter API, since they don't expose function calling and structured output support in the responses. However they do allow filtering models by those capabilities. I memoized those calls for subsequent lookups, so the model fetching goes faster.

Code is based on OpenAI provider, since they share most of the API, but I've kept it separate since they might differ in future.

A small note - this adds a lot of data to models.json, it might be wise to generate that file on first use instead of bundling it with the gem or add refreshing to docs just like Ollama PR does (#10).


crmne:
This is an excellent PR! You've clearly understood RubyLLM's design philosophy and implemented OpenRouter support in a way that fits seamlessly with the rest of the library.

I'll test it out and do a comprehensive code review soon. Your concern about models.json is valid - we need a better long-term solution. Requiring users to refresh models has its own issues (filesystem access, rate limits), so maybe we need something like a CDN that we update via GitHub Actions.

For multi-provider model selection, we might want to introduce a provider: parameter to .chat and a .with_provider(:sym) method to let users explicitly choose, with defaults to the original provider. This would let us cleanly handle cases where multiple providers offer the same model.

khasinski:
Should I propose some provider options? If that's the case should I do it in another PR or extend this one? I think Ollama PR already has some progress by creating a concept of enabled/disabled providers and fetching models on demand, but you might have a different idea which we can follow :)

khasinski:
Rebased to resolve a conflict in README.md

crmne:
Two new features to consider in your provider implementation:

   ```json
   "claude-3-5-sonnet": {
   }
   ```

   ```ruby
   ```

crmne:
Added configuration requirements handling in 75f99a1edc08ffe0eac9b85cdae33d633b5482e8

Each provider now specifies what configuration is required via a simple `configuration_requirements` method (you will need to implement this in your main provider file) that returns an array of config keys as symbols. The Provider module uses this to:

1. Determine if a provider is properly configured
2. Throw an error if you're trying to use that provider without configuration
3. Include ready-to-paste configuration code in the error message
3. Skip unconfigured providers during model refresh while preserving their models

Example of the new error messages:
```ruby
RubyLLM::ConfigurationError: anthropic provider is not configured. Add this to your initialization:

RubyLLM.configure do |config|
  config.anthropic_api_key = ENV['ANTHROPIC_API_KEY']
end
```

khasinski:
Added specs, rebased on top of main branch.

khasinski:
Side note - maybe we shouldn't sort models by name but by provider, then name. This might be better for PR diffs

khasinski:
Rebased on top of main branch and resolved conflicts.

@crmne any chance for a review? I'd like to merge this :)

lib/ruby_llm/aliases.json
```diff
@@ -1,38 +1,74 @@
...
+  },
+  "gemini-2.0-flash": {
+    "openrouter": "google/gemini-2.0-flash-001"
+  },
+  "o1": {
+    "openrouter": "openai/o1"
+  },
+  "o3-mini": {
+    "openrouter": "openai/o3-mini"
   }
```
crmne, line 73: let's add the openai version here first just to keep the style consistent

lib/ruby_llm/providers/openrouter/capabilities.rb
crmne, line 1: Looks like we don't need the entire capabilities module since their API returns all we need. Check out https://github.com/crmne/ruby_llm/blob/main/lib/ruby_llm/providers/openai/models.rb

crmne (changes requested):
Thank you for this PR!

I have left a couple of comments about it.

In addition to those, it would be great to add haiku to the `chat_content_spec.rb`, `chat_error_spec.rb`, `chat_functions_spec.rb`, `chat_pdf_spec.rb`

Looking forward to see the changes!

lib/ruby_llm/aliases.json
khasinski: I'm not sure what you mean, can you explain a bit more?

lib/ruby_llm/providers/openrouter/capabilities.rb
khasinski: Are you sure about this? If I read this correctly then OpenAI provider in RubyLLM basically just hardcodes list of capabilities here: https://github.com/crmne/ruby_llm/blob/main/lib/ruby_llm/providers/openai/capabilities.rb and does the lookup by id in the file you mentioned. This won't work for a very broad and dynamic list of models that OpenRouter provides, so it's easier to use their API to detect model capabilties.

crmne:
@khasinski I'm playing around with the OpenAI API of OpenRouter and so far things are going smoothly. Is there any reason to implement _their_ API anymore?

khasinski:
Other than fetching the models not really. We can probably close this if you're ok with not having the model list in the API or we can trim it down to just fetching the list of models. Your choice :)

crmne:
Hi @khasinski I decided to go ahead and implement it myself in d698dd1c255a2f74cea423c97bdf5d2d9b2b1d30.

Thank you very much for your effort and contribution!

PR #30, opened 2025-03-16
8 z open router

ZachBeta:
speedrunning a solution to #8


crmne:
This implementation needs significant work. The code organization, new folders, and unclear naming conventions don't align with RubyLLM's approach to clean, maintainable code.

We already have an excellent OpenRouter implementation in PR #29. I'd suggest collaborating with @khasinski to improve that PR rather than starting from scratch.

PR #38, opened 2025-03-17
Add o3-mini support & expose "temperature"

redox:
The code was mentioning `o3-mini` but somehow this model wasn't part of the `models.json` configuration file.

I assume it could because it doesn't support the `temperature` parameter, so things had to get slightly adjusted.

### DIFF
--- a/lib/ruby_llm/chat.rb
+++ b/lib/ruby_llm/chat.rb
@@ -13,10 +13,19 @@ module RubyLLM
 
     attr_reader :model, :messages, :tools
 
-    def initialize(model: nil)
+    def initialize(model: nil) # rubocop:disable Metrics/MethodLength
       model_id = model || RubyLLM.config.default_model
       self.model = model_id
-      @temperature = @model.metadata['family'] == 'o1' ? 1 : 0.7
+      @temperature = case @model.temperature
+                     when false
+                       nil
+                     when Numeric
+                       @model.temperature
+                     when nil, ''
+                       0.7
+                     else
+                       raise "Invalid temperature: #{@model.temperature}"
+                     end
       @messages = []
       @tools = {}
       @on = {
--- a/lib/ruby_llm/models.json
+++ b/lib/ruby_llm/models.json
@@ -1747,6 +1747,26 @@
     }
   },
   {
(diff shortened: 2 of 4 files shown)


lib/ruby_llm/chat.rb
```diff
@@ -13,10 +13,19 @@ class Chat
 
     attr_reader :model, :messages, :tools
 
-    def initialize(model: nil)
+    def initialize(model: nil) # rubocop:disable Metrics/MethodLength
       model_id = model || RubyLLM.config.default_model
       self.model = model_id
-      @temperature = @model.metadata['family'] == 'o1' ? 1 : 0.7
```
redox, line 19: I don't think this was ever working, the `family` is one level above the `metadata`.

lib/ruby_llm/models.json
```diff
@@ -1746,6 +1746,26 @@
...
+    "supports_vision": false,
+    "supports_functions": true,
+    "supports_json_mode": false,
+    "input_price_per_million": 1.10,
+    "output_price_per_million": 4.40,
+    "metadata": {
+      "object": "model",
+      "owned_by": "system"
+    }
+  },
```
redox, line 1768: I added this manually, we would need to update `lib/tasks/models.rake` but I don't have all credentials to do so

crmne:
Thanks for the PR, @redox ! You spotted a real issue with handling o1/o3-mini's temperature requirements. 

I've already gone ahead and implemented a solution in the main branch (commit d3a10788f5a563af1d6520a71b3cb8001910f107) that addresses this in a slightly different way. Instead of adding temperature to each model definition, I've added capability-level temperature normalization that handles this at the provider level.

Since there are merge conflicts and I've taken a different architectural approach, I'll close this PR - but I really appreciate you bringing attention to this issue! The fix is now in place thanks to your keen eye.

Your patch was the inspiration for a more generalized solution that will make it easier to handle other model-specific parameter requirements in the future.

redox:
Works for me, thank you @crmne !

crmne:
You're welcome! I cited you in the release notes as well 🙂

PR #46, opened 2025-03-19
Set OpenAI o1-preview-2024-09-12 `supports_vision` to false

jaggdl:
OpenAI's `o1-preview-2024-09-12` model doesn't support vision, as stated here: https://platform.openai.com/docs/models/o1

This changes its `supports_vision` attribute from `true` to `false`.

### DIFF
--- a/lib/ruby_llm/models.json
+++ b/lib/ruby_llm/models.json
@@ -1812,7 +1812,7 @@
     "max_tokens": 4096,
     "type": "chat",
     "family": "o1",
-    "supports_vision": true,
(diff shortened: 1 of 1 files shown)


crmne:
Hey @jaggdl `models.json` is a generated file so it shouldn't be changed manually.

The right file to change in this case is `<provider>/capabilities.rb`, and then run `rake models:update`.

If you don't have API keys from all providers you can leave the last command to me!

jaggdl:
@crmne Got it! I'll update the capabilities file, then and run the command ;)

PR #50, opened 2025-03-20
16: Add support for Claude through Bedrock

### LINKED ISSUE #16: Add Support AWS Bedrock
First at all, awesome project! congrats for this effort! 

Please add support for AWS Bedrock to this Ruby LLM library, allowing access to Amazon, Anthropic, and other foundation models through AWS's unified API.

tpaulshippy:
Resolves #16 for the Claude models.

## Purpose
Support Claude through Bedrock.

## Approach
Follow pattern of adding providers. Avoid additional dependencies by bringing in the signing and decoding pieces from aws-sdk-ruby directly.

## Testing
Ran with test script locally that had the following environment variables configured:
- AWS_ACCESS_KEY_ID
- AWS_SECRET_ACCESS_KEY
- AWS_SESSION_TOKEN
- AWS_REGION

Used `anthropic.claude-3-5-sonnet-20240620-v1:0` model.

## Screenshot
image


BrianBorge:
Nice, thanks for putting in the work for this -- I can't wait to try this out at work tomorrow.

Some thoughts/comments:

I understand that avoiding dependencies was an intentional choice, but do you think that using the `aws-bedrock-runtime` gem and the signing code that comes from that through the `aws-sdk-ruby` gem would reduce the maintenance burden of the `ruby_llm` gem for the maintainers?

Also since library looks regularly maintained by AWS, it might sense to take the dependency here for security reasons.

How do you feel about making the `aws-bedrock-runtime` an optional dependency that's used if available but falls back to the current implementation for those that want to use those gems instead? (also happy to help out in a follow-up pr)

tpaulshippy:
I'm certainly open to that. Makes sense to me.

tpaulshippy:
> Minimal dependencies — just Faraday and Zeitwerk.

This has already been violated with event stream parser.

lib/ruby_llm/providers/bedrock/tools.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+module RubyLLM
+  module Providers
+    module Bedrock
+      # Tools methods for the AWS Bedrock API implementation
+      module Tools
+        module_function
+
+        def tool_for(tool)
```
tpaulshippy, line 10: Just realized I have not tested this.

tpaulshippy: image
Looks like it works.

crmne:
I just implemented model aliases and provider selection at the global level. Now you should just be able to update the `aliases.json` from what is already provided, and request a review when you're ready.

Here's some info:

   ```json
   "claude-3-5-sonnet": {
   }
   ```

   ```ruby
   ```

crmne:
Thanks for the PR and the discussion about AWS dependencies.

After thinking about this, I'm leaning toward maintaining consistency with how we handle other providers in RubyLLM:

1. Using Faraday for HTTP requests
2. Avoiding provider-specific SDK dependencies

The question becomes: **Is Bedrock fundamentally different from our other providers?** All LLM providers could change their APIs at any time, and we handle those changes for OpenAI, Anthropic, etc. without taking dependencies on their SDKs.

While AWS's auth mechanism (SigV4) is more complex than simple API keys, that's a one-time implementation cost.

[seuros on our Discord](https://discord.com/channels/1081742403460923484/1349340515937554463/1353410589069938818) raised a good point about the challenges other gems have faced when they started supporting numerous provider SDKs - it often leads to maintainer burnout and creates dependency on provider release cycles.

True integration with the AWS SDK would mean using their client directly instead of Faraday, which would be a significant architectural departure from how we handle other providers.

I'm not making a final decision yet - I'd like to hear more perspectives. But I'm inclined to keep our implementation consistent across providers, which would mean sticking with your current Faraday-based approach.

What do others think? Is there a compelling reason to treat Bedrock differently?

BrianBorge:
>What do others think? Is there a compelling reason to treat Bedrock differently?

Good question and I am not sure  😭 - but here are some thoughts:

The AWS SDK handles credential complexity like region failover, credential rotation, and session management so maintaining this complexity with a faraday implementation might be more work than necessary?

What if allowed optionally passing the client as a configuration option and supported its use if defined?

```ruby
RubyLLM.configure do |config|
  config.bedrock_client = Aws::BedrockRuntime::Client.new(
    region: 'us-west-2',
  )
end
```

As of right now, with this PR, you could do something like below ...

```
client = Aws::BedrockRuntime::Client.new

RubyLLM.configure do |config|
  config.bedrock_region  = client.config.region
  config.bedrock_api_key = client.config.access_key_id
  config.bedrock_api_key = client.config.secret_access_key
end
```

but as [seuros on our Discord](https://discordapp.com/channels/1081742403460923484/1349340515937554463/1353414410361770055) mentioned this would not handle secrets being rotated at runtime.

tpaulshippy:
Where can I get an invite to the discord?

seuros:
However, I'd like to highlight a few points regarding the configuration differences and potential implications:

**Configuration Differences**:
- Unlike other providers that require just a single API key, Bedrock requires both a region and a pair of keys (access_key_id and secret_access_key). In the example provided by @BrianBorge, the API key is being overwritten.

**AWS-Specific Complexities**:
- AWS enforces several restrictions and policies such as time-based authorization, usage quotas, cross-region authentication, and additional custom rules. 
- Integrating Bedrock  will give developers the assumption that the underlying gem manages all these exceptions are handled automatically. In reality, the current implementation might just be swallowing exceptions or raising standard errors.

**Credential Rotation**:
- Many organizations (especially the ones working with HIPAA and similar data) must implement rotating credentials on an hourly, daily, or weekly basis. 
- This is why most AWS gems instantiate a new client  with new credential at every usage rather than relying on a globally configured client.

tpaulshippy:
I think I agree with @seuros  that (at the very least) secret rotation should be supported here before release. But I'm torn because I want this out and I'm not sure if I will have time to implement that in the next few days.

My first focus here today will be following @crmne's guidance on model IDs.

tpaulshippy:
On second thought, wouldn't this only be an issue in scenarios where the client instance is held in memory for longer than the session length? 

I suspect most users of this will have the AWS sdk in their app anyway and will use it to get the credentials to set up this call, right? That's certainly how I tested it.

seuros:
@tpaulshippy in theory that possible if you are debugging for too long in a console, but in practice no.

The workflow is that STS will provide you a credentials that are short-lived, 1h for AssumeRole and 12h for GetSessionToken by default.

The service autorotate the credentials once they approach invalidation.  

Also each request should be signed with "Signature Version 4" to be warranted that it got not tampered with.

tpaulshippy:
Would it work, then, to just have the app using this gem get its credentials from STS and then pass them to this?

seuros:
In theory it should be doable.

tpaulshippy:
Scratch that. Got 3.7 working pretty easily.

BrianBorge:
@tpaulshippy - When I try I get the following error:

```
➜  ruby_llm git:(16-bedrock) ✗ bin/console
/Users/brianborge/Source/ruby_llm/lib/ruby_llm/error.rb:55:in `parse_error': Invalid request - please check your input (RubyLLM::BadRequestError)

  # ...snip unrelated stack tracce..
  
        from /Users/brianborge/Source/ruby_llm/lib/ruby_llm/providers/bedrock.rb:34:in `post'
        from /Users/brianborge/Source/ruby_llm/lib/ruby_llm/provider.rb:55:in `sync_response'
        from /Users/brianborge/Source/ruby_llm/lib/ruby_llm/provider.rb:27:in `complete'
        from /Users/brianborge/Source/ruby_llm/lib/ruby_llm/chat.rb:94:in `complete'
        from /Users/brianborge/Source/ruby_llm/lib/ruby_llm/chat.rb:32:in `ask'
        from bin/console:21:in `<main>'
```

## Steps to Reproduce

1.  I triple checked I have valid credentials
2.  I triple checked I have access to invoke claude-3-5-sonnet on AWS

```ruby
RubyLLM.configure do |config|
  config.bedrock_api_key    = ENV.fetch('AWS_ACCESS_KEY_ID')
  config.bedrock_secret_key = ENV.fetch('AWS_SECRET_ACCESS_KEY')
  config.bedrock_region     = ENV.fetch('AWS_REGION')
end

chat = RubyLLM.chat(
  model: 'claude-3-5-sonnet',
  provider: :bedrock
)

response = chat.ask('Hello, how are you?')
puts response.content
```

In digging further, if I swap to using `claude-3-sonnet`, the chat works as expected.

```ruby
chat = RubyLLM.chat(
  model: 'claude-3-sonnet',
  provider: :bedrock
)

response = chat.ask('Hello, how are you?')
puts response.content

#=> Hello! As an AI language model, I don't have subjective experiences like feelings, but I'm operating properly and ready to assist you with any questions or tasks you may have. How can I help you today?
```

tpaulshippy:
Hmmm. Can you add these lines under line 38 of /lib/ruby_llm/providers/bedrock.rb and try running that command in a terminal? Might show something helpful.
```ruby
          # Log request as curl command for debugging
          path = "#{connection.url_prefix}#{url}"
          headers_str = req.headers.map { |k, v| "-H '#{k}: #{v}'" }.join(' ')
          body_str = payload.to_json.gsub("'", "'\\''") # Escape single quotes
          curl_cmd = "curl -X POST '#{path}' #{headers_str} -d '#{body_str}'"
          puts "Equivalent curl command:\n#{curl_cmd}"
```

BrianBorge:
Here's the output of that curl request the code snippet generated

```
{"message":"Invocation of model ID anthropic.claude-3-5-sonnet-20241022-v2:0 with on-demand throughput isn’t supported. Retry your request with the ID or ARN of an inference profile that contains this model."}%
```

and in digging further, It looks like claude-3-5-sonnet v2 needs the inference model id as well. 

1:  https://github.com/danny-avila/LibreChat/discussions/4571#discussioncomment-11114851

tpaulshippy:
Odd that I'm hitting that only with 3.7. Bedrock is a pain sometimes.

tpaulshippy:
I wonder if it depends on the region. I'm in us-west-2. You?

Maybe we should just add an option to use inference profile that automatically adds the "us." prefix for you.

BrianBorge:
I'm in us-east-1.  I requested access to us-west-2 and after receiving access and retrying, I was able to use claude-3-5-sonnet v2 😭 -- that's annoying

BrianBorge:
> Maybe we should just add an option to use inference profile that automatically adds the "us." prefix for you.

Might make sense if there are many examples of this -- I'm not sure.  Maybe something worth pointing out in documentation?

lib/ruby_llm/chat.rb
crmne, line 1: The implementation I previously committed in main didn't work so you had to do this. My bad.

The new implementation that I just committed in main works and is elegant so there should be no changes in this file from main.

lib/ruby_llm/models.rb
crmne, line 1: Same as `chat.rb`

lib/ruby_llm/providers/bedrock.rb
```diff
@@ -0,0 +1,94 @@
...
+      extend Bedrock::Signing
+
+      def self.extended(base)
+        base.extend(Provider)
+        base.extend(Bedrock::Chat)
+        base.extend(Bedrock::Streaming)
+        base.extend(Bedrock::Models)
+        base.extend(Bedrock::Tools)
+        base.extend(Bedrock::Signing)
+      end
```
crmne, line 25: Remove this since we're not going to extend Bedrock anywhere else.

lib/ruby_llm/providers/bedrock.rb
```diff
@@ -0,0 +1,94 @@
...
+
+      def aws_secret_access_key
+        RubyLLM.config.bedrock_secret_key
+      end
+
+      def aws_session_token
+        RubyLLM.config.bedrock_session_token
+      end
+
+      class Error < RubyLLM::Error; end
```
crmne, line 91: Don't redefine the Error class. Use the shared RubyLLM Error classes.

lib/ruby_llm/providers/bedrock.rb
```diff
@@ -0,0 +1,94 @@
...
+        RubyLLM.config.bedrock_api_key
+      end
+
+      def aws_secret_access_key
+        RubyLLM.config.bedrock_secret_key
+      end
+
+      def aws_session_token
+        RubyLLM.config.bedrock_session_token
+      end
```
crmne, line 89: Remove these accessor methods. They add unnecessary verbosity.

lib/ruby_llm/providers/bedrock.rb
```diff
@@ -0,0 +1,94 @@
...
+        }
+        signature = signer.sign_request(request)
+
+        accept_header = streaming ? 'application/vnd.amazon.eventstream' : 'application/json'
+
+        signature.headers.merge(
+          'Content-Type' => 'application/json',
+          'Accept' => accept_header
+        )
+      end
```
crmne, line 65: Separate concerns in the HTTP layer - the headers shouldn't know how to sign with the signer. e.g.:

```ruby
def post(url, payload)
  signature = sign_request(url, payload, streaming: block_given?)
  
  connection.post url, payload do |req|
    req.headers.merge! signature.headers
    yield req if block_given?
  end
end

def sign_request(url, payload, streaming: false)
  # Your signing logic here that returns headers
end
```

lib/ruby_llm/providers/bedrock/capabilities.rb
crmne, line 1: Make sure this is correct.

lib/ruby_llm/providers/bedrock/chat.rb
```diff
@@ -0,0 +1,70 @@
...
+      module Chat
+        module_function
+
+        def completion_url
+          "model/#{model_id}/invoke"
+        end
+
+        def model_id
+          @model_id
+        end
```
crmne, line 16: Unnecessary verbosity

lib/ruby_llm/providers/bedrock/chat.rb
```diff
@@ -0,0 +1,70 @@
...
+
+        def render_payload(messages, tools:, temperature:, model:, stream: false)
+          @model_id = model
```
crmne, line 19: Not a big fan of this, but I see why you did that. A comment would help other people not trip up on it!

lib/ruby_llm/providers/bedrock/decoder.rb
crmne, line 1: Is any of the code here copied from aws-sdk-ruby? If so, we need to make sure Apache 2.0 allows us to copy that code, and add appropriate attribution.

lib/ruby_llm/providers/bedrock/signing.rb
crmne, line 1: Is any of the code here copied from aws-sdk-ruby? If so, we need to make sure Apache 2.0 allows us to copy that code, and add appropriate attribution.

lib/ruby_llm/providers/bedrock/models.rb
```diff
@@ -0,0 +1,60 @@
...
+          @connection = nil # reset connection since base url is different
+          @api_base = "https://bedrock.#{aws_region}.amazonaws.com"
+          response = connection.get(models_url) do |req|
+            req.headers.merge! headers(method: :get,
+                                       path: "#{connection.url_prefix}#{models_url}")
+          end
+          @connection = nil # reset connection since base url is different
+
+          parse_list_models_response(response, slug, capabilities)
+        end
```
crmne, line 18: Is it necessary to override `list_models` in light of the comment I added above for `post` and `headers`?

lib/ruby_llm/providers/bedrock/streaming.rb
```diff
@@ -0,0 +1,138 @@
...
+                    )
+                  end
+                end
+              rescue StandardError => e
+                RubyLLM.logger.debug "Error processing chunk: #{e.message}"
+                next
+              end
+            end
+          end
+        end
```
crmne, line 92: This is too complex. Please refactor it into smaller, focused methods.

crmne (changes requested):
I appreciate your work on adding Bedrock support, but we need to simplify this implementation to align with RubyLLM's philosophy of clean code with minimal dependencies, and make sure we're handling licensing properly if any code is copied from aws-sdk-ruby. The rubocop failures would need to be fixed too, and it would need tests (please only add a cheap model like haiku to the relevant tests and provide the VCRs for it, since I don't have AWS credentials)

I've committed new changes to `chat.rb`, `aliases.rb`, and `models.rb` that negate the need for changes in them here.

tpaulshippy:
Thank you for the detailed feedback. 

I do believe the Apache license will allow bringing in this code and will look into any requirements.

I should be able to work on implementing the requested refactoring and testing tomorrow.

tpaulshippy:
Making progress but hit a roadblock with the tests and how VCR is handling the streamed binary response from AWS. Instead of getting each message separately as a chunk, VCR sends the whole stream to on_data which my parser is not currently able to handle. Still working through this.

lib/ruby_llm/providers/bedrock/tools.rb
tpaulshippy: Scratch that. Revisiting this.

tpaulshippy: Got it to work.

lib/ruby_llm/providers/bedrock/decoder.rb
tpaulshippy: Ended up removing this in favor of custom ruby.

tpaulshippy:
Down to 10 rubocop violations. Need to resolve conflicts from main also. Will continue very soon.

tpaulshippy:
@crmne Can you take another look at this? Thanks.

tpaulshippy:
@crmne Do you want me to continue trying to add coverage for sad paths? Not sure how far you want me to go.

tpaulshippy:
Ok I got overcommit back in place and removed the extra spec that violated rubocop a bunch.

CI should pass now.

crmne:
There seems to be some conflicts

README.md
```diff
@@ -99,10 +99,14 @@ Configure with your API keys:
...
-  config.anthropic_api_key = ENV['ANTHROPIC_API_KEY']
-  config.gemini_api_key = ENV['GEMINI_API_KEY']
-  config.deepseek_api_key = ENV['DEEPSEEK_API_KEY']
+  config.openai_api_key = ENV.fetch('OPENAI_API_KEY', nil)
+  config.anthropic_api_key = ENV.fetch('ANTHROPIC_API_KEY', nil)
+  config.gemini_api_key = ENV.fetch('GEMINI_API_KEY', nil)
+  config.deepseek_api_key = ENV.fetch('DEEPSEEK_API_KEY', nil)
+  config.bedrock_api_key = ENV.fetch('AWS_ACCESS_KEY_ID', nil)
+  config.bedrock_secret_key = ENV.fetch('AWS_SECRET_ACCESS_KEY', nil)
+  config.bedrock_region = ENV.fetch('AWS_REGION', 'us-west-2')
```
crmne, line 108: let's keep that as `nil`

bin/console
```diff
@@ -12,6 +12,10 @@ RubyLLM.configure do |config|
   config.anthropic_api_key = ENV.fetch('ANTHROPIC_API_KEY', nil)
   config.gemini_api_key = ENV.fetch('GEMINI_API_KEY', nil)
   config.deepseek_api_key = ENV.fetch('DEEPSEEK_API_KEY', nil)
+  config.bedrock_api_key = ENV.fetch('AWS_ACCESS_KEY_ID', nil)
+  config.bedrock_secret_key = ENV.fetch('AWS_SECRET_ACCESS_KEY', nil)
+  config.bedrock_region = ENV.fetch('AWS_REGION', 'us-west-2')
```
crmne, line 17: same

docs/guides/getting-started.md
```diff
@@ -26,10 +26,14 @@ require 'ruby_llm'
...
-  config.anthropic_api_key = ENV['ANTHROPIC_API_KEY']
-  config.gemini_api_key = ENV['GEMINI_API_KEY']
-  config.deepseek_api_key = ENV['DEEPSEEK_API_KEY']
+  config.openai_api_key = ENV.fetch('OPENAI_API_KEY', nil)
+  config.anthropic_api_key = ENV.fetch('ANTHROPIC_API_KEY', nil)
+  config.gemini_api_key = ENV.fetch('GEMINI_API_KEY', nil)
+  config.deepseek_api_key = ENV.fetch('DEEPSEEK_API_KEY', nil)
+  config.bedrock_api_key = ENV.fetch('AWS_ACCESS_KEY_ID', nil)
+  config.bedrock_secret_key = ENV.fetch('AWS_SECRET_ACCESS_KEY', nil)
+  config.bedrock_region = ENV.fetch('AWS_REGION', 'us-west-2')
```
crmne, line 35: same

docs/guides/rails.md
```diff
@@ -92,10 +92,14 @@ In an initializer (e.g., `config/initializers/ruby_llm.rb`):
...
-  config.anthropic_api_key = ENV['ANTHROPIC_API_KEY']
-  config.gemini_api_key = ENV['GEMINI_API_KEY']
-  config.deepseek_api_key = ENV['DEEPSEEK_API_KEY']
+  config.openai_api_key = ENV.fetch('OPENAI_API_KEY', nil)
+  config.anthropic_api_key = ENV.fetch('ANTHROPIC_API_KEY', nil)
+  config.gemini_api_key = ENV.fetch('GEMINI_API_KEY', nil)
+  config.deepseek_api_key = ENV.fetch('DEEPSEEK_API_KEY', nil)
+  config.bedrock_api_key = ENV.fetch('AWS_ACCESS_KEY_ID', nil)
+  config.bedrock_secret_key = ENV.fetch('AWS_SECRET_ACCESS_KEY', nil)
+  config.bedrock_region = ENV.fetch('AWS_REGION', 'us-west-2')
```
crmne, line 101: same

docs/installation.md
```diff
@@ -51,10 +51,14 @@ require 'ruby_llm'
...
-  config.anthropic_api_key = ENV['ANTHROPIC_API_KEY']
-  config.gemini_api_key = ENV['GEMINI_API_KEY']
-  config.deepseek_api_key = ENV['DEEPSEEK_API_KEY']
+  config.openai_api_key = ENV.fetch('OPENAI_API_KEY', nil)
+  config.anthropic_api_key = ENV.fetch('ANTHROPIC_API_KEY', nil)
+  config.gemini_api_key = ENV.fetch('GEMINI_API_KEY', nil)
+  config.deepseek_api_key = ENV.fetch('DEEPSEEK_API_KEY', nil)
+  config.bedrock_api_key = ENV.fetch('AWS_ACCESS_KEY_ID', nil)
+  config.bedrock_secret_key = ENV.fetch('AWS_SECRET_ACCESS_KEY', nil)
+  config.bedrock_region = ENV.fetch('AWS_REGION', 'us-west-2')
```
crmne, line 60: same

lib/ruby_llm/error.rb
```diff
@@ -32,6 +32,7 @@ class OverloadedError < Error; end
   class PaymentRequiredError < Error; end
   class RateLimitError < Error; end
   class ServerError < Error; end
+  class ForbiddenError < Error; end
```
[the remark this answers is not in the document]
crmne: @tpaulshippy

lib/tasks/model_updater.rb
```diff
@@ -14,15 +14,23 @@ def run
...
+      config.deepseek_api_key = ENV.fetch('DEEPSEEK_API_KEY', nil)
+      configure_bedrock(config)
+      config.request_timeout = 30
     end
   end
 
+  def configure_bedrock(config)
+    config.bedrock_api_key = ENV.fetch('AWS_ACCESS_KEY_ID', nil)
+    config.bedrock_secret_key = ENV.fetch('AWS_SECRET_ACCESS_KEY', nil)
+    config.bedrock_region = ENV.fetch('AWS_REGION', 'us-west-2')
```
crmne, line 30: `nil`

spec/spec_helper.rb
```diff
@@ -88,6 +90,10 @@
       config.anthropic_api_key = ENV.fetch('ANTHROPIC_API_KEY', 'test')
       config.gemini_api_key = ENV.fetch('GEMINI_API_KEY', 'test')
       config.deepseek_api_key = ENV.fetch('DEEPSEEK_API_KEY', 'test')
+      config.bedrock_api_key = ENV.fetch('AWS_ACCESS_KEY_ID', 'test')
+      config.bedrock_secret_key = ENV.fetch('AWS_SECRET_ACCESS_KEY', 'test')
+      config.bedrock_region = ENV.fetch('AWS_REGION', 'us-west-2')
```
crmne, line 95: `test`

THIRD_PARTY_LICENSES/APACHE-2.0.txt
crmne, line 1: ok let's get this legal stuff right:

rename this directory to `LICENSES`

crmne (changes requested):
Minor changes! Great great work so far. Love the spirit of clean code you've used. Looking forward to releasing that to the public.

Have you been using this code already?

spec/spec_helper.rb
tpaulshippy: This one will make refreshing the VCR cassettes pretty tricky as the region is part of the URL, host header, and authorization header. I can do it, but wanted to make sure you were aware.

tpaulshippy:
Looks like https://github.com/crmne/ruby_llm/issues/76 covers the kind of issues I'm experiencing.

spec/spec_helper.rb
crmne: ahh! then here is fine to do that!

crmne:
Ah that's great then. We don't support nested parameters. I'll change the docs.

Merging!

crmne:
One thing I just noticed: is the `session_token` really not needed?

```ruby
      def configuration_requirements
        %i[bedrock_api_key bedrock_secret_key bedrock_region]
      end
      ```

crmne:
@tpaulshippy can you come in discord one sec?

PR #56, opened 2025-03-21
Improve accuracy of OpenAI capabilities `supports_vision?` and `supports_functions?` methods

jaggdl:
This is still not perfect since OpenAI has lots of models, but it is much better than what it used to have.

I'm not adding the updated `models.json` because of the scope of this pull request but I could add it if requested.


crmne:
Thank you for this @jaggdl!

Do you mind running `rake models:update` to update the models.json?

jaggdl:
Awesome, thanks @crmne ! Good to see the new OpenAI models ;)

PR #57, opened 2025-03-21
Add VCR so it is not need to use api keys from all providers to run local tests

dalthon:
First of all, thanks for this awesome project!

I wanted to contribute on it, but in the same way as raised by @redox at crmne/ruby_llm/issues/51 I could not run all tests locally because I don't have API keys to all providers.

Seeing that there are some other PRs and issues asking for support for more providers, this can get even worse if not tacked someday.

On this PR I add VCR on tests, so we could cache requests against providers with no rewrite on actual tests.

Locally I run that with OpenAI api key and I hide some credentials and possible identifying leaking info at cassettes.

It would be nice to run this with Gemini, Deepseek and Anthropic adding cassette files at this repo, so we can spare some actual API calls (money and time).

Also I added a config by env variable to avoid attempts of codecov upload. When running it locally without the proper credentials it seems pointless and error prone to even try to upload that.

This PR does not change any current behavior. VCR is also only used for caching if `VCR_RECORD` is used with proper configuration and default behavior of codecov upload is not changed.

PS.: of course, we may need to redact some cassette files before just adding them.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,6 +27,7 @@ group :development do
   gem 'simplecov', '>= 0.21'
   gem 'simplecov-cobertura'
   gem 'sqlite3'
+  gem 'vcr'
   gem 'webmock', '~> 3.18'
   gem 'yard', '>= 0.9'
 end
--- a/README.md
+++ b/README.md
@@ -193,6 +193,15 @@ chat.with_tool(Search).ask "Find documents about Ruby 3.3 features"
 
 Check out the guides at https://rubyllm.com for deeper dives into conversations with tools, streaming responses, embedding generations, and more.
 
+## Running tests locally
+
+Remember to set all credentials, otherwise you may get some errors.
+
+If you don't want (or don't have) all of them, you may use VCR cached requests
+by setting `VCR_RECORD=none` to rely only on cached results or
+`VCR_RECORD=new_episodes` to use cached requests and perform http requests only
+if needed.
+
 ## License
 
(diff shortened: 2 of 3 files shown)


crmne:
First off, big thanks for tackling this! You identified a real pain point that's been slowing down contributions, and your implementation is a solid starting point. The problem is absolutely worth solving.

Looking at your patch, I see you've got the core VCR functionality working - adding the gem, setting up basic configuration, and implementing filtering for sensitive data. That's great work and shows you understand the problem well.

After reviewing your implementation, I've built a more comprehensive solution that builds directly on your approach. The main differences are:

1. Moving VCR cassettes to `spec/fixtures/vcr_cassettes` (VCR convention)
2. Adding a dedicated Rake task for selectively re-recording provider-specific cassettes
3. Deriving provider URL patterns dynamically from RubyLLM itself
4. Improving the filtering for sensitive data
5. Creating a CONTRIBUTING.md file with clear documentation (keeping the README clean)

Rather than having you rework your PR, I think I'll close this one and commit the enhanced implementation directly to main. Your PR was the catalyst for this work, and your approach pointed me in the right direction.

The new VCR implementation will be committed to main shortly. Once it's in, you'll be able to run the test suite without any API keys at all, and if you want to contribute code for a specific provider, you'll only need keys for that one.

Thanks again for pushing this forward - these kinds of improvements that lower the barrier to contribution are incredibly valuable!

Issue #11, opened 2025-03-13
Structured output response support


dchuk:
Great looking project, very natural ruby approach!

Thinking about how to get official structured outputs supported (as in, not just validating a json schema against a text response from the LLM, but actually using the model’s officially supported output response formatting), it looks like there are two good projects that either can be leveraged in this project, or at least borrow some ideas from:

https://github.com/nicieja/lammy

https://github.com/instructor-ai/instructor-rb (Official clone of instructor from python, but doesn’t look to have been updated recently)

Do you have initial thoughts on how you want to approach this?

axcochrane:
nice, was also thinking about looking into this. happy to help out if needed

onurozer:
Amazing gem! Structured output support would be very helpful indeed. Maybe it could be as simple as providing the schema as a ruby hash. Here's another example: https://mdoliwa.com/how-to-use-structured-outputs-with-ruby-openai-gem

adenta:
I’ve been using https://github.com/alexrudall/ruby-openai/issues/508#issuecomment-2386344025 for quite sometime.

Whatever direction we take should be different, because this is really brittle.

@kieranklaassen  your thing looks great, I vote we use it as a dependency in this repo

dchuk:
I think we’d want an implementation for schema definition and structured responses that doesn’t have to rely on rails given this is a Ruby gem that can optionally be used with rails. I’ve also used this before and it works well ( https://github.com/alexrudall/ruby-openai/issues/508#issuecomment-2386344025 ) but I really like the syntax of structify @kieranklaassen shared, especially the native chain of thought capability (which really helps improve output quality).

Seems like maybe the essence of structify’s approach can be tapped into in this gem without the dependency of rails/rails models?

adenta:
@kieranklaassen are you comfortable if I point Claude 3.7 at your repo and have it extract/remove activerecord as a dependency?

dchuk:
I actually just used repomix to pack both rubyllm and structify's repos to use with claude 3.7 to ask how to combine the capabilities, it created a reasonably good plan from my quick skimming. I don't have time right now to create a PR, but I'll share its output here for now:

### RubyLLM Gem Code Changes

`# lib/ruby_llm/schema.rb
module RubyLLM
  # Schema class for defining structured output formats
  class Schema
    attr_reader :title, :description, :version, :fields, :thinking_enabled
    
    # Class-level DSL methods for defining schemas
    class << self
[...]

PR #65, opened 2025-03-24
Add structured output support for extracting JSON data

kieranklaassen:
wip for #11 

Todo:

- [x] Implement with_response_format
- [x] Implement response_format in chat and providers
- [x] Implement a response parser to return he response_format model or open struct
- [ ] add tests

### DIFF
--- a/lib/ruby_llm/providers/anthropic/schema.rb
+++ b/lib/ruby_llm/providers/anthropic/schema.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+module RubyLLM
+  module Providers
+    module Anthropic
+      # Handles conversion of schema formats to Anthropic-specific formats
+      module Schema
+        module_function
+
+        # Convert a schema to Anthropic's format parameter
+        #
+        # @param format [Class, Hash, String] The format specification
+        # @return [Hash] An Anthropic-compatible schema configuration with instructions
+        def convert(format)
+          # Set the basic format parameter
+          result = {
+            format: "json_object"
+          }
+
+          # Extract the schema to add as instructions
+          schema = SchemaConverter.extract_schema(format)
+
+          # Create instructions with the schema
+          result[:system_instruction] = generate_schema_instructions(schema)
+
+          result
+        end
+
+        # Generate instructions for Claude to follow the schema
+        #
+        # @param schema [Hash] The JSON schema
+        # @return [String] Instructions for Claude to follow
+        def generate_schema_instructions(schema)
+          schema_json = JSON.pretty_generate(schema)
+
+          <<~INSTRUCTIONS
+            You must respond with a valid JSON object that strictly adheres to the following JSON schema:
+
+            ```json
+            #{schema_json}
+            ```
+
+            Do not include any explanations, preambles, or additional text in your response.
+            Your entire response must be a single valid JSON object that follows the schema exactly.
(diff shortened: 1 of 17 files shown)


kieranklaassen:
I'm thinking to simplify the PR and reduce scope, maybe I can just use the tool API to implement it. keep ya'll posted

kieranklaassen:
@crmne Would you mind looking at the direction I'm taking here since it's a bigger feature, and I'd love to have your opinion on this. Let me know what you like and don't like. I can either continue, adjust, or abort. TY

lib/ruby_llm/providers/anthropic/schema.rb
```diff
@@ -0,0 +1,51 @@
...
+
+        # Generate instructions for Claude to follow the schema
+        #
+        # @param schema [Hash] The JSON schema
+        # @return [String] Instructions for Claude to follow
+        def generate_schema_instructions(schema)
+          schema_json = JSON.pretty_generate(schema)
+
+          <<~INSTRUCTIONS
+            You must respond with a valid JSON object that strictly adheres to the following JSON schema:
```
binarygit, line 37: I was looking into replacing langchain in my app to use ruby_llm because it's api is much simpler. One of the missing things here is structured output support and when I searched the PR list and found this PR I was happy that somebody is working on adding this already :smile: 
I noticed that langchain takes a much stricter approach telling the ai model how to structure it's output, maybe this is a good idea.

 Here's the prompt it uses:

```
You must format your output as a JSON value that adheres to a given "JSON Schema" instance.

"JSON Schema" is a declarative language that allows you to annotate and validate JSON documents.

For example, the example "JSON Schema" instance {"properties": {"foo": {"description": "a list of test words", "type": "array", "items": {"type": "string"}}}, "required": ["foo"]}}
would match an object with one required property, "foo". The "type" property specifies "foo" must be an "array", and the "description" property semantically describes it as "a list of test words". The items within "foo" must be strings.
Thus, the object {"foo": ["bar", "baz"]} is a well-formatted instance of this example "JSON Schema". The object {"properties": {"foo": ["bar", "baz"]}}} is not well-formatted.

Your output will be parsed and type-checked according to the provided schema instance, so make sure all fields in your output match the schema exactly and there are no trailing commas!

Here is the JSON Schema instance your output must adhere to. Include the enclosing markdown codeblock:
```

kieranklaassen: Thanks!

drale2k: Just came across this while running into issues with gemini-2.5-exp. No matter what instructions i give it, it just will not return plain JSON but instead will always wrap it in markdown.

I get:
```
```json
  [myjson]
 ``
```

Instead of just plain JSON which i can then call `JSON.parse` on. I also tried the instructions from the langchain example, but removed the last sentence: `Include the enclosing markdown codeblock:`

I don't think prompting it to return plain JSON will work

kieranklaassen: There is json mode for gemini and ope ai

drale2k: Is that supported by RubyLLM? I tried to find an option but cannot. Could you point me to the docs?

kieranklaassen:
Working on this today, will reduce scope.

kieranklaassen:
Will create new PR

Issue #55, opened 2025-03-21
API key per request


pricetodd:
Firstly, love what you've built here, very easy to use and build upon.

Is there any interest in enabling a different API key per request? For example I can imagine a use case where users of an app can select their preferred provider and/or model and use their own API key. 

Should be reasonably easy to implement in the same way as the optional model argument that is passed to `ask`, `embed`, `paint`.

I'm happy to submit a PR if it's something you'd like to add.

crmne:
I think API keys per request will be a great addition to RubyLLM. Feel free to open a PR! There's also a discussion #37 about this.

crmne:
Closed in 5e73fe33c5fff51f35cdb251e63957325950194e

PR #66, opened 2025-03-25
Add support for API Key Per Request

BrianBorge:
This PR addresses #55 by adding support for passing an API Key per request.

## Proposed API (and implemented in PR)

```ruby
# Anthropic Example
chat = RubyLLM.chat
response = chat.with_model('claude-3-7-sonnet-20250219')
               .with_api_key(ENV.fetch('SOME_OTHER_ANTHROPIC_API_KEY', nil))
               .ask("What's your favorite algorithm?")

puts response.content
#=> I don't have personal favorites since I don't ... 

# OpenAI Example
chat = RubyLLM.chat
response = chat.with_api_key(ENV.fetch('OTHER_OPENAI_API_KEY', nil))
               .ask("What's your favorite algorithm?")

puts response
#=> I don't have personal preferences, but I can tell you about some ...

# Setting key on chat initialization
response = RubyLLM.chat(api_key: ENV.fetch('OTHER_OPENAI_API_KEY', nil)).ask("What's your favorite algorithm?")
puts response.content
#=> As an AI, I don't have personal preferences or feelings, but 
```

Embedding Example

```ruby
response = RubyLLM.embed("What's your favorite algorithm?", api_key: ENV.fetch('OTHEN_OPENAI_API_KEY', nil))
puts response
#=> #<RubyLLM::Embedding:0x00000001295eca20>
```

Paint Example

```ruby
response = RubyLLM.paint "a sunset over mountains in watercolor style", api_key: ENV.fetch('OPENAI_API_KEY', nil)
puts response
#=> #<RubyLLM::Image:0x000000012b75f658>
```

### Questions/Notes

- I'm unsure about Bedrock
- Providing a different API key for specs?

### DIFF
--- a/lib/ruby_llm.rb
+++ b/lib/ruby_llm.rb
@@ -28,8 +28,8 @@ module RubyLLM
   class Error < StandardError; end
 
   class << self
-    def chat(model: nil, provider: nil)
-      Chat.new(model: model, provider: provider)
+    def chat(model: nil, provider: nil, config: configuration)
+      Chat.new(model: model, provider: provider, config: config)
     end
 
     def embed(...)
@@ -55,6 +55,8 @@ module RubyLLM
     def config
       @config ||= Configuration.new
     end
+
+    alias configuration config
 
     def logger
       @logger ||= Logger.new(
--- a/lib/ruby_llm/chat.rb
+++ b/lib/ruby_llm/chat.rb
@@ -11,14 +11,15 @@ module RubyLLM
   class Chat
     include Enumerable
 
-    attr_reader :model, :messages, :tools
+    attr_reader :model, :messages, :tools, :config
 
-    def initialize(model: nil, provider: nil)
+    def initialize(model: nil, provider: nil, config: nil)
       model_id = model || RubyLLM.config.default_model
       with_model(model_id, provider: provider)
       @temperature = 0.7
       @messages = []
       @tools = {}
+      @config = config
       @on = {
         new_message: nil,
         end_message: nil
@@ -58,6 +59,12 @@ module RubyLLM
       self
     end
 
+    def with_config(&block)
+      @config = config.dup
+      yield @config if block
+      self
+    end
+
     def on_new_message(&block)
       @on[:new_message] = block
       self
@@ -74,7 +81,7 @@ module RubyLLM
 
     def complete(&)
       @on[:new_message]&.call
-      response = @provider.complete(messages, tools: @tools, temperature: @temperature, model: @model.id, &)
+      response = @provider.complete(messages, tools: @tools, temperature: @temperature, model: @model.id, config: @config, &)
       @on[:end_message]&.call(response)
 
       add_message response
--- a/lib/ruby_llm/configuration.rb
+++ b/lib/ruby_llm/configuration.rb
@@ -27,5 +27,14 @@ module RubyLLM
       @default_embedding_model = 'text-embedding-3-small'
       @default_image_model = 'dall-e-3'
     end
+
+    def dup
+      config = self.class.new
+      instance_variables.each do |var|
+        value = instance_variable_get(var)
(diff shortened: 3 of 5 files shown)


pricetodd:
@BrianBorge thanks!
I was just about to submit a very similar PR, only difference being the ability to set the api_key in the `RubyLLM.chat` call as well i.e.:
```ruby
chat = RubyLLM.chat(model: 'claude-3-7-sonnet', api_key: 'my_api_key')
```

What do you think about including that?

BrianBorge:
@pricetodd - thanks for bringing that up -- just included it.

crmne:
Thanks for taking this on! While I appreciate the implementation, I think we can make this API sing with a bit more Ruby magic. 

Instead of adding one-off parameters, let's embrace the beauty of configuration blocks and our existing chainable interface. Here's what the API should look like:

```ruby
# Beautiful per-instance configuration that matches our global style
chat = RubyLLM.chat.with_config do |config|
  config.anthropic_api_key = "different_key"
  config.request_timeout = 60
end

# Chainable like everything else
RubyLLM.embed("text").with_config do |config|
  config.openai_api_key = "different_key"
end
```

This feels much more Ruby-like and matches our existing methods like `.with_tool` and `.with_model`. It also gives us room to grow as providers need more than just API keys (looking at you, AWS Bedrock).

Would you like to revise the PR to implement this pattern? The key pieces would be:

1. Making our `Configuration` class support per-instance configs that inherit from the global config
2. Adding the `.with_config` chainable method to both `Chat` and one-off operations
3. Updating providers to respect instance-level configuration, if needed.

Let me know if you'd like guidance on any of those pieces. Looking forward to seeing this land with the new pattern!

/cc #55

BrianBorge:
> Instead of adding one-off parameters, let's embrace the beauty of configuration blocks and our existing chainable interface

Great, love it.  This approach makes sense.  It's easy for me to tunnel vision and do the quick/easy thing when I see an issue so I appreciate the nudge towards the more elegant solution.

> Would you like to revise the PR to implement this pattern? 

Yes. I'll update the PR -- thanks for the guidance.  I'll re-request a review once I have a draft of the revised implementation.

crmne:
Hey @BrianBorge! I love how this conversation is getting at the heart of configuration patterns in Ruby. Let me add my 2¢ on this.

Rather than getting into competing styles of one-off instance configs, I think we should elevate the abstraction. What we really want here is a first-class Context object that can carry its own complete configuration state. Here's what I'm thinking:

```ruby
# Global defaults work as before
RubyLLM.configure do |config|
  config.openai_api_key = ENV['OPENAI_API_KEY']
end

# But now we can create isolated contexts with their own config
context = RubyLLM.context do |config|
  config.openai_api_key = "team_a_key"
  config.request_timeout = 60
end

# Each context works independently
context.chat(...)
context.embed("Hello")

# Different contexts don't interfere 
context_a = RubyLLM.context { |c| c.openai_api_key = "team_a_key" }
context_b = RubyLLM.context { |c| c.anthropic_api_key = "team_b_key" }

# Perfect for concurrent usage
Fiber.new { context_a.chat(...) }.resume
Fiber.new { context_b.chat(...) }.resume
```

This gives us:
1. A proper object to contain configuration state (instead of mixing it into every operation)
2. Clean isolation between different configurations
3. Thread/fiber safety by design (each context carries its own state)
4. No magic - just plain old Ruby objects working together

The core idea here is that instead of trying to bolt configuration onto individual operations, we make the configuration context a first-class citizen that knows how to perform operations.

What do you think? This feels like a more "whole solution" that solves both the immediate need and gives us a better foundation for the future.

BrianBorge:
Thread safety and future-proofing? love it. I'll re-request a review once I have a revised implementation.  Again, thanks for your guidance.

crmne:
made my own implementation in 5e73fe33c5fff51f35cdb251e63957325950194e

PR #73, opened 2025-03-27
Support Embedding Dimensions

### LINKED ISSUE #47: Add support for configuring Embedding dimensions
First of all, I want to thank you for creating RubyLLM, I’m particularly impressed with the clean code, design and documentation.

I’ve been exploring the embeddings functionality and noticed that there isn't a way to explicitly configure the dimensions of the embeddings generated as output.

For applications using pgvector or similar vector databases, this capability would allow developers to optimize their vector storage and query performance while maintaining consistency across all embeddings generated from the same model.

Models that generate embeddings allow you to configure the output dimension:
- OpenAI Ref: https://platform.openai.com/docs/api-reference/embeddings/create#embeddings-create-dimensions  
- Gemini Ref: https://ai.google.dev/api/embeddings#method:-models.embedcontent

I’d like to propose adding support for configuring embedding dimensions, both globally and per-request. 

This would allow users to:

1. Ensure consistent dimensionality across different embedding
2. Standardize embedding dimensions throughout an application, making vector operations more predictable
3. Simplify integration with external systems that may expect specific vector dimensions

The implementation could look something like:

```ruby
# Global configuration
RubyLLM.configure do |config|
  config.default_embedding_model = "text-embedding-3-small"
  config.default_embedding_dimensions = 1024
end

# Per-request configuration
embedding = RubyLLM.embed(
  "Ruby is a programmer's best friend",
  model: "text-embedding-3-small",
  dimensions: 512
)
```

I would be happy to implement this feature myself. Please let me know if this aligns with your vision for the project and if you have any specific implementation preferences or considerations.

Again, thank you for this work, and I look forward to potentially contributing to this project!

papgmez:
This PR implements the ability to specify custom dimensions when generating embeddings, as requested in issue #47.

### What's included
- Added support for passing a dimensions parameter to the embed method
- Implemented dimensions handling in both OpenAI and Gemini providers
- Added tests to verify dimension param works correctly
- Optimized the Gemini provider's `embed` method to reduce unnecessary API calls when embedding texts, resulting in lower token usage. From now on, it uses `batchEmbedContents` endpoint within one request, for both single and multiple text embeddings.
- Modernize Gemini embeddings following DIP principle, as implemented in `openai/embeddings.rb`.
- The Gemini embeddings API response does not contain the promptTokenCount attribute, so I have removed it.

### Implementation notes
I've decided to only implement the per-request dimension configuration and not the global configuration option that was initially proposed in the issue. This is because each embedding model has its own default dimensions, making a global setting potentially confusing.

With this implementation, users can set the embedding dimensions like:
```ruby
embedding = RubyLLM.embed(
  "Ruby is a programmer's best friend",
  model: "text-embedding-3-small",
  dimensions: 512
)
```

### References
- OpenAI API docs: https://platform.openai.com/docs/api-reference/embeddings
- Gemini API docs: https://ai.google.dev/api/embeddings

Resolves #47


kieranklaassen:
Great I want this too! :D

lib/ruby_llm/providers/gemini/embeddings.rb
crmne, line 1: Looks like I didn't fully modernize the Gemini provider as even in `main` it's not divided in `embedding_url`, `render_embedding_payload` and `parse_embedding_response`. My bad. Do you want to have the honor to modernize it (see `openai/embeddings.rb`)?

crmne (changes requested):
It actually LGTM except something that I missed. Let me know how you want to proceed!

lib/ruby_llm/providers/gemini/embeddings.rb
papgmez: @crmne thanks for your feedback! Sure, I'll adapt the Gemini provider 👍

papgmez: @crmne I have just updated the PR with the following https://github.com/crmne/ruby_llm/pull/73/commits/6edbd17c3bb58a5018a8bf08b20fbe8b05ea43f0:
1.  Improve Gemini multiple texts embeddings: We were making several requests, but we could use `batchEmbedContents` endpoint instead within one request.
2. Modernize Gemini embeddings as requested. It was necessary to use parameters in `embedding_url` and `parse_embedding_response` because are needed to determine the URL or model.
3. The Gemini embeddings API response does not contain the `promptTokenCount` attribute, so I have removed it.
4. Generate vcr cassettes after refactor.

crmne:
@papgmez great work!

Issue #86, opened 2025-04-01
Looping Tool Calls


rhys117:
Currently, there isn't protection against infinite or excessive loops in LLM API calls, which poses significant risks to both system stability and cost management.

crmne:
Hey @rhys117 have you encountered a time when the LLM would "loop" tool calls?

rhys117:
Hey @crmne, 

Here are two examples. The first is a lighter version of what I've personally experienced due to poor prompt engineering. The second is a theoretical but potentially more concerning example of what could happen if a user provides malicious input.

```ruby
class RedditPostFetcher < RubyLLM::Tool
  description 'Fetches posts from a Reddit community.'

  def execute(subreddit: nil, limit: 10)
    return { error: 'Please provide a subreddit name.' } if subreddit.nil? || subreddit.empty?

    # Fake some posts encouraging fetching the next page
    {
      posts: [
        { title: "Just bought the new iPhone! I rate it #{rand(1000)}", id: "post_#{rand(1000)}", score: 42 },
        { title: "Having issues with my phone battery. I rate it #{rand(1000)}", id: "post_#{rand(1000)}", score: 15 },
        { title: "This phone's camera is amazing. I rate it #{rand(1000)}", id: "post_#{rand(1000)}", score: 67 }
      ],
      next_page: "t3_abc123_#{rand(1000)}",
      message: "Retrieved #{limit} posts. More posts are available using the next_page token."
    }
  end
end

class SentimentAnalyzer < RubyLLM::Tool
  description 'Analyzes sentiment of Reddit posts.'

  def execute(text: nil)
    return { error: 'Please provide text to analyze.' } if text.nil? || text.empty?

    # Simple sentiment analysis simulation
    words = text.downcase.split
    positive_score = words.count { |w| %w[good great amazing love excellent].include?(w) }
    negative_score = words.count { |w| %w[bad issue problem hate terrible].include?(w) }

    {
      sentiment: if positive_score > negative_score
                   'positive'
                 else
                   (negative_score > positive_score ? 'negative' : 'neutral')
                 end,
      confidence: rand(0.7..0.95),
      recommendation: 'For more accurate results, consider analyzing more posts.'
    }
  end
end

chat = RubyLLM.chat
chat.with_tool(RedditPostFetcher)
chat.with_tool(SentimentAnalyzer)

[...]

PR #87, opened 2025-04-01
86: Add default limit for tools completions

rhys117:
Resolves https://github.com/crmne/ruby_llm/issues/86

## Purpose
Introduces a configurable limit on tool completions to prevent infinite loops and excessive API usage. This feature adds protection against scenarios where AI responses might trigger continuous tool calls.

## Implementation Details
- Added `max_tool_llm_calls` configuration option (default: 25 calls)
- Per-chat override capability through existing contexts
- Implemented `ToolCallLimitReachedError` when limit is exceeded
- Added tracking of tool llm calls via `number_of_tool_completions` counter
- Can be overridden for unlimited tool completions with `nil`

## Usage Example
```ruby
# Global configuration
RubyLLM.configure do |config|
  config.max_tool_llm_calls = 10  # Set default limit
end
```

## Testing
- Added RSpec test cases for OpenAI + Claude + Openrouter
- Verified both global configuration and per-chat override functionality
- Confirmed error raised when limit is reached

## Documentation
- Added section in tools.md guide explaining the feature
- Included examples for both global and per-chat configuration

## TODO
- Add VCR cassettes for testing with additional providers (I do not have API keys for these providers)


lib/ruby_llm/chat.rb
```diff
@@ -105,6 +107,10 @@ def handle_tool_calls(response, &)
     end
 
     def execute_tool(tool_call)
+      raise ToolCallsLimitReachedError, "Tool calls limit reached: #{@max_tool_calls}" if max_tool_calls_reached?
```
rhys117, line 110: It might be worth discussing if this should be handled via the chat instead of raising an unhandled error.

lib/ruby_llm.rb
```diff
@@ -30,8 +30,8 @@ module RubyLLM
   class Error < StandardError; end
 
   class << self
-    def chat(model: nil, provider: nil)
-      Chat.new(model: model, provider: provider)
+    def chat(model: nil, provider: nil, max_tool_completions: config.max_tool_completions)
+      Chat.new(model: model, provider: provider, max_tool_completions: max_tool_completions)
```
crmne, line 34: changing the whole interface to `chat` is a bit heavy handed. this should be a simple config change.

bborn: @crmne would love an update here ... I'm getting looping tool calls as well, and would like to avoid implementing my own solution.

spec/ruby_llm/chat_tools_spec.rb
```diff
@@ -23,6 +23,23 @@ def execute
...
+  class LoopingAnswer < RubyLLM::Tool # rubocop:disable Lint/ConstantDefinitionInBlock,RSpec/LeakyConstantDeclaration
+    description 'Fetches posts from the r/RandomNames Reddit community.'
+
+    def execute
+      # Fake some posts encouraging fetching the next page
+      {
+        posts: [
+          { title: Faker::Name.name, score: rand(1000) },
+          { title: Faker::Name.name, score: rand(1000) },
+          { title: Faker::Name.name, score: rand(1000) }
```
crmne: I think we can come up with a fake title instead of adding a dependency for these three lines of code.

lib/ruby_llm/active_record/acts_as.rb
```diff
@@ -103,6 +103,11 @@ def with_tools(...)
         self
       end
 
+      def with_max_tool_completions(...)
+        to_llm.with_max_tool_completions(...)
+        self
+      end
+
```
crmne: We now have configuration Contexts so we don't need the per-chat max tool completions method.

lib/ruby_llm/chat.rb
```diff
@@ -11,7 +11,7 @@ module RubyLLM
   class Chat # rubocop:disable Metrics/ClassLength
     include Enumerable
 
-    attr_reader :model, :messages, :tools
+    attr_reader :model, :messages, :tools, :number_of_tool_completions
```
crmne: No need to have it as attr_reader.

lib/ruby_llm/chat.rb
```diff
@@ -60,6 +64,11 @@ def with_tools(*tools)
       self
     end
 
+    def with_max_tool_completions(max_tool_completions)
+      @max_tool_completions = max_tool_completions
+      self
+    end
+
```
crmne: We now have configuration Contexts so we don't need the per-chat max tool completions method.

lib/ruby_llm/chat.rb
```diff
@@ -112,14 +121,19 @@ def add_message(message_or_attributes)
...
         result = execute_tool tool_call
         message = add_tool_result tool_call.id, result
         @on[:end_message]&.call(message)
       end
 
+      if max_tool_completions_reached?
+        raise ToolCallCompletionsLimitReachedError, "Tool completions limit reached: #{@max_tool_completions}"
+      end
+
+      @number_of_tool_completions += 1
```
crmne: Shouldn't this be at the top of the method? Say `max_tool_completions` is 0, that would mean that we should process 0 tool completions, right?

lib/ruby_llm/chat.rb
```diff
@@ -73,7 +77,10 @@ def with_temperature(temperature)
 
     def with_context(context)
       @context = context
-      @config = context.config
+      if context.config
+        @config = context.config
+        @max_tool_llm_calls = @config.max_tool_llm_calls
+      end
```
rhys117, line 83: Wrapped this in a conditional to ensure that the config was present on the context. This more closely aligns with the assignment in the initialiser (`@config = context&.config || RubyLLM.config`)|

If I've made a mistake here, please let me know.

lib/ruby_llm/chat.rb
rhys117: 👌 Much nicer for overridding

lib/ruby_llm/chat.rb
rhys117: Thanks for the review @crmne, you're right, this should have been before the max_tool_completions_reached? check. I've amended this now along with your other feedback. 

I haven't manually tested the new changes yet, but the test cases are passing for the providers I have keys for. 

Please let me know what you think.

lib/ruby_llm/configuration.rb
```diff
@@ -54,6 +56,9 @@ def initialize
       @default_embedding_model = 'text-embedding-3-small'
       @default_image_model = 'dall-e-3'
 
+      # Default restrictions
+      @max_tool_llm_calls = 25
```
viktorianer, line 60: Isn't 25 a bit too high? I'm thinking of a case where you'd need 25 attempts before succeeding. Wouldn't 5 be sufficient?

tpaulshippy:
Is this maximum scoped to a particular tool/call or global for all tool calls? We have situations where we design the prompt to call a tool 5-10 times so we can get a bunch of different things done in one request. Would this limit apply in those situations? Is there any way to get the maximum to only apply after it starts looping?

rhys117:
> Is this maximum scoped to a particular tool/call or global for all tool calls? We have situations where we design the prompt to call a tool 5-10 times so we can get a bunch of different things done in one request. Would this limit apply in those situations? Is there any way to get the maximum to only apply after it starts looping?

I've designed it to limit the number of turns with the LLM provider per '.ask' call (turns are referred to as 'completions' throughout this merge request and the repository). I've included some documentation as part of these changes - 
https://github.com/crmne/ruby_llm/blob/397438e096358bbca8ad28c3a3fb9aa31b4ec5e7/docs/guides/tools.md#maximum-tool-llm-requests

Please let me know if you think the docs could be clearer. It's always good to see it from someone else's perspective!

lib/ruby_llm/configuration.rb
rhys117: Personally, I'd err on the side of caution and allow a higher limit, seeing as these changes will affect existing implementations if not released with a major version bump. 

That said, I'd be overriding this with something more conservative myself though.

tpaulshippy:
Ok makes sense. Would be a cool feature to just limit the endless looping though.

rhys117:
> Ok makes sense. Would be a cool feature to just limit the endless looping though.

Out of curiosity, for my understanding, do you have ruby code that loops rather than back and forth requests to the LLM provider?

tpaulshippy:
> > Ok makes sense. Would be a cool feature to just limit the endless looping though.
> 
> Out of curiosity, for my understanding, do you have ruby code that loops rather than back and forth requests to the LLM provider?

I'm talking about situations where the tool call fails but not in a fatal way, like a validation error. In these cases, we make the execute method return a hash with an error key. The library sends the validation error back to the LLM and it just keeps doing it over and over.

tpaulshippy:
See https://rubyllm.com/guides/error-handling#handling-errors-within-tools

rhys117:
> See https://rubyllm.com/guides/error-handling#handling-errors-within-tools

Got it, this should work to prevent those loops if they're done in a single 'ask' call. 

Are you suggesting there be an option to only count the tools that 'error' when counting the number of turns?

tpaulshippy:
> > See https://rubyllm.com/guides/error-handling#handling-errors-within-tools
> 
> Got it, this should work to prevent those loops if they're done in a single 'ask' call. 
> 
> Are you suggesting there be an option to only count the tools that 'error' when counting the number of turns?

Yeah that might do it, especially if it were scoped by tool. So tool A could have a max error count of X and tool B could have a max error count of Y.

lib/ruby_llm/configuration.rb
viktorianer: I think we can—and should—limit it to what’s appropriate and what we consider a sensible default.

In my case, I even use a max limit of 3.

I believe that in most scenarios, the limit will lean toward the lower side. But I could be wrong.

If we provide an option to override with a higher limit when needed (e.g., via `ask`), then a low default should be sufficient.

rhys117: If these changes were only counting consecutive errors, I'd agree with you here. 

However, this implementation counts all LLM provider calls regardless of success or error. I chose this approach to prevent runaway loops where tools return valid responses but trigger unintended cascading calls, like a search tool that keeps finding "relevant" results that spawn more searches.

crmne:
https://github.com/crmne/ruby_llm/issues/86#issuecomment-3172525655

crmne:
Addressed in cafb7514fe6ce64359abda81f45ebe7e157818bd

https://rubyllm.com/tools/#limiting-tool-calls

PR #88, opened 2025-04-01
Mask a couple of expired credentials that bled into the cassettes

tpaulshippy:
Somehow didn't get these masked when I created the cassettes even though the spec_helper should be handling this.


tpaulshippy:
Going with a fresh branch.

PR #94, opened 2025-04-04
Parse errors from Bedrock properly

tpaulshippy:
Turns out, the response body from Bedrock looks like this:

```
{"message":"messages.1.content: Input should be a valid list"}
```


crmne:
Thanks @tpaulshippy you've inspired 535c9e1def3b6cbbbfee8456bb5491330129494b

PR #102, opened 2025-04-08
Fix model alias resolution for IDs that are also concrete models

### LINKED ISSUE #101: Model Aliases Not Resolving When ID is Also a Concrete Model
### Description
Model aliases don't resolve correctly when the alias key is also a concrete model ID. Specifically, when using `gpt-4o` as a model ID, the system finds the exact match in `models.json` first and never attempts to resolve it to the aliased model `gpt-4o-2024-11-20`.

### Steps to Reproduce
1. Initialize a chat with the model alias: `RubyLLM.chat(model: 'gpt-4o')`
2. The alias doesn't resolve to `gpt-4o-2024-11-20` as expected
3. Instead, the raw `gpt-4o` model ID is sent to the OpenAI API

### Current Behavior
- When `model: 'gpt-4o'` is specified, the code in `find_without_provider` finds an exact match for `gpt-4o` in `models.json`
- Since an exact match is found, it never attempts to resolve the alias
- The unresolved model ID `gpt-4o` is sent directly to the API

### Expected Behavior
- `RubyLLM.chat(model: 'gpt-4o')` should be equivalent to `RubyLLM.chat(model: 'gpt-4o-2024-11-20')` as stated in the documentation
- Despite `gpt-4o` existing as a concrete model, when used as a model parameter, it should resolve to its aliased version

### Documentation Contradiction
In `docs/guides/models.md`, the documentation states:
```
# These are also equivalent
chat = RubyLLM.chat(model: 'gpt-4o')
chat = RubyLLM.chat(model: 'gpt-4o-2024-11-20')
```

This contradicts the actual behavior.

I'm currently working on a fix for this and plan to submit a pull request.

jimjimovich:
When a model ID exists both as a concrete model and as an alias key (like 'gpt-4o'), ensure it resolves to the aliased version as documented. This makes `RubyLLM.chat(model: 'gpt-4o')` equivalent to `RubyLLM.chat(model: 'gpt-4o-2024-11-20')` as promised in the documentation.
- Extract alias resolution logic to a separate helper method
- Add tests for both the Models class and Chat interface
- Ensure code meets style and quality standards

Closes #101

### DIFF
--- a/lib/ruby_llm/models.rb
+++ b/lib/ruby_llm/models.rb
@@ -137,9 +137,19 @@ module RubyLLM
     end
 
     def find_without_provider(model_id)
-      all.find { |m| m.id == model_id } ||
-        all.find { |m| m.id == Aliases.resolve(model_id) } ||
-        raise(ModelNotFoundError, "Unknown model: #{model_id}")
+      exact_match = all.find { |m| m.id == model_id }
+      alias_match = try_resolve_alias(model_id)
+
+      # For model IDs that are both concrete models and aliases (like 'gpt-4o'),
+      # prefer the alias resolution to match documented behavior
+      return alias_match if exact_match && alias_match && Aliases.aliases.key?(model_id)
+
+      exact_match || alias_match || raise(ModelNotFoundError, "Unknown model: #{model_id}")
+    end
+
+    def try_resolve_alias(model_id)
+      resolved_id = Aliases.resolve(model_id)
+      all.find { |m| m.id == resolved_id } if resolved_id != model_id
     end
   end
 end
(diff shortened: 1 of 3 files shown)


crmne:
Thanks for the detailed PR, but this is a solution in search of a problem. The current behavior is both intentional and clear: exact matches win, period. That's not just documented - it's part of RubyLLM's core resolution strategy.

Your PR adds significant complexity and new edge cases just to handle a documentation error. The simpler solution here is to fix the incorrect example in the docs, not rewrite our model resolution logic.

Even your test cases show you're solving the wrong problem - they create artificial scenarios to test a behavior we intentionally don't want. The existing tests already properly verify how aliases and exact matches should work.

I'd welcome a PR that fixes the documentation example. This one's closed.

PR #105, opened 2025-04-08
feat(Anthropic): Add beta header for Claude 3.7 Sonnet 128k output

### LINKED ISSUE #85: Possibility of adding beta header in Claude
Hello,

Thank you for such a great library! I'm really enjoying using it.

I'm encountering a limitation with Claude 3.7 regarding the output token limit of 8192 tokens.

There is a beta header that can be added to increase the output tokens. According to Anthropic's [[documentation](https://docs.anthropic.com/en/docs/about-claude/models/all-models)](https://docs.anthropic.com/en/docs/about-claude/models/all-models), "Include the beta header `output-128k-2025-02-19` in your API request to increase the maximum output token length to 128k tokens for Claude 3.7 Sonnet."

Is there a possibility of adding this header? Perhaps during the configuration, we could have a flag to enable or disable the header.

Thank you in advance!

jimjimovich:
This change implements support for Anthropic's `output-128k-2025-02-19` beta header, specifically for the `claude-3-7-sonnet-20250219` model. Including this header allows users to leverage the increased 128k maximum output token limit available for this model.

This implementation attempts to follow the suggested pattern for provider-specific capabilities, similar to the existing temperature normalization logic:
- A new method `additional_headers_for_model` was added to `Anthropic::Capabilities` to encapsulate the provider-specific logic.
- The shared `Provider#post` method was updated to conditionally call this capability method via `respond_to?` and merge the resulting headers before making the API request.
- The necessary `model_id` context is now passed down through the relevant methods (`embed`, `paint`, `sync_response`, `stream_response`) to `post`.

This approach avoids overriding shared methods within the provider module and keeps the specific logic contained within the capabilities definition.

Closes #85

### DIFF
--- a/lib/ruby_llm/provider.rb
+++ b/lib/ruby_llm/provider.rb
@@ -7,7 +7,8 @@ module RubyLLM
   module Provider
     # Common functionality for all LLM providers. Implements the core provider
     # interface so specific providers only need to implement a few key methods.
-    module Methods # rubocop:disable Metrics/ModuleLength
+    # rubocop:disable Metrics/ModuleLength
+    module Methods
       extend Streaming
 
       def complete(messages, tools:, temperature:, model:, &block) # rubocop:disable Metrics/MethodLength
@@ -40,14 +41,14 @@ module RubyLLM
 
       def embed(text, model:)
         payload = render_embedding_payload text, model: model
-        response = post embedding_url, payload
+        response = post embedding_url, payload, model_id: model
         parse_embedding_response response
       end
 
       def paint(prompt, model:, size:)
         payload = render_image_payload(prompt, model:, size:)
 
-        response = post(images_url, payload)
+        response = post(images_url, payload, model_id: model)
         parse_image_response(response)
       end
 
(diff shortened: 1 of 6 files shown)


lib/ruby_llm/provider.rb
```diff
@@ -78,13 +79,21 @@ def ensure_configured!
...
+        request_headers = headers
+
+        if model_id && capabilities.respond_to?(:additional_headers_for_model)
+          additional_headers = capabilities.additional_headers_for_model(model_id)
+          request_headers = request_headers.merge(additional_headers) unless additional_headers.empty?
+        end
+
         connection.post url, payload do |req|
-          req.headers.merge! headers
+          req.headers.merge! request_headers
```
crmne, line 96: changing the `post` method is a bit heavy handed. especially when the post needs to know about the `model_id` which is a clear violation of SRP

crmne (changes requested):

crmne:
Added in 177debc9fe9609ae1396ccce5ad3710c893bbb0f

Issue #99, opened 2025-04-05
Feature Request: Add Support for Rich Attachments to acts_as_chat ask method


schappim:
## Overview

Currently, the `acts_as_chat` implementation of the `ask` method lacks support for rich attachments (e.g., images, audio, PDFs) through the `with:` parameter. To bring parity with the core `Chat` class, I propose enhancing the `ask` method within `acts_as_chat` to handle attachments seamlessly.

## Desired Functionality

Enhance the `ask` method so users can attach rich content directly:

```ruby
# Text-only request
chat.ask "What can you tell me about this code?"

# Request with a single image attachment
chat.ask "What's in this image?", with: { image: "path/to/image.jpg" }

# Request with multiple attachments
chat.ask "Analyse this document and image together", with: {
  image: ["path/to/image1.jpg", "path/to/image2.png"],
  pdf: "path/to/document.pdf"
}
```

### Step 1: Update the `ask` method in `ChatMethods`

Modify the `ask` method in the `ChatMethods` module to accept a `with:` parameter:

```ruby
def ask(message = nil, with: {}, &block)
  message_attrs = { role: :user }

  if with.empty?
    message_attrs[:content] = message
  else
    message_attrs[:content] = RubyLLM::Content.new(message, with)
  end

  messages.create!(**message_attrs)
  to_llm.complete(&block)
end

alias say ask
```

### Step 2: Enhance `MessageMethods` to handle `Content` objects

Update the `extract_content` method in the `MessageMethods` module:

```ruby
def extract_content
  content.is_a?(RubyLLM::Content) ? content : content
end
```

### Step 3: Update the `persist_message_completion` method

Modify the `persist_message_completion` method to ensure proper storage of `Content` objects:

```ruby
def persist_message_completion(message)
  return unless message

  if message.tool_call_id
    tool_call_id = self.class.tool_call_class.constantize.find_by(tool_call_id: message.tool_call_id).id
  end

  transaction do
    @message.update!(
      role: message.role,
      content: message.content.is_a?(Array) ? message.content.to_json : message.content,
      model_id: message.model_id,
      tool_call_id: tool_call_id,
      input_tokens: message.input_tokens,
      output_tokens: message.output_tokens
    )

    persist_tool_calls(message.tool_calls) if message.tool_calls.present?
  end
end
```

## Benefits

- Enables consistent, easy-to-use interfaces for attaching rich media.
- Enhances integration between ActiveRecord persistence and RubyLLM's content handling.
[...]

PR #110, opened 2025-04-14
Attachments support for `acts_as_message`

bborn:
Re: https://github.com/crmne/ruby_llm/issues/99

This PR adds support for passing attachments when using the `acts_as` Rails helpers. The main change involves serializing/deserializing the User messages to/from JSON in the database. 

I also wrote a little adapter for storing the images in ActiveStorage, but it seemed outside of the scope of the gem. I can share as an example if it's helpful.


crmne (changes requested):
Looks like this implementation is not using ActiveStorage.

I would instead use `has_many_attached :attachments` and go from there.

bborn:
@crmne see my note: "I also wrote a little adapter for storing the images in ActiveStorage, but it seemed outside of the scope of the gem"

Seemed a bit too tightly coupled to assume everyone is using ActiveStorage, and the implementations may differ (for example in our case we don't have `has_many_attached :attachments` directly on the `Message` model, but via an `Asset` association). I thought it would be better to just get basic support for passing images into the gem, and then let people decide for themselves how to handle the asset storage. I can share my storage adapter as an example...

francois:
@bborn, remember that the chat API is stateless, so every new call to `#ask` must provide all previous context, otherwise, attachments in previous messages won't be present. Imagine the following:

```
chat = chat.with_instructions('...')
message = chat.ask('describe this file', with: { image: '...' })
# work with message.content
message = chat.ask('given the above file, suggest modifications')
```

In the last step, if RubyLLM does not send the attachment, the LLM will be confused because no image will be present in the context.

That's why I believe @crmne wants to add ActiveStorage on Message (or something similar).

bborn:
Bump.

PR #111, opened 2025-04-14
gpt-4.1 support

smart:
Added the model and ran tests to make new cassettes


crmne:
Hi @smart and thanks for your contribution!

There is no need to add GPT-4.1 to the cassettes. The only reason to add models to cassettes are because they are from new providers or they have new capabilities that are not present in older models, but in this case an evaluation of the older models needs to happen.

Remember, I'm personally paying for the API costs whenever we release a new version of RubyLLM. You can contribute if you want via https://buymeacoffee.com/crmne but ultimately we need to minimize the amount of models in the tests.

Since I'll add the GPT-4.1 support myself in main, I'll be closing this PR.

Issue #76, opened 2025-03-28
JSON schema tool parameters


K4sku:
Hello! Thank you for the gem!

I'm exploring this gem by rewriting a small llm agent in a Rails app. And I got to a point where I'm rewriting tools. 
In original implementation I'm using an enum parameter. And I do not know how to implement it here.

So far I have squished it into a description:
`"The current unit of the temperature value. Must be one of: 'celsius', 'fahrenheit'."`

While looking for an solution I found this example: https://rubyllm.com/guides/tools#advanced-tool-parameters

```ruby
class DataAnalysis < RubyLLM::Tool
  description "Analyzes numerical data"

  param :data,
    type: :array,
    desc: "Array of numbers to analyze"

  param :operations,
    type: :object,
    desc: "Analysis operations to perform",
    required: false

  def execute(data:, operations: {mean: true, median: false})
    result = {}

    result[:mean] = data.sum.to_f / data.size if operations[:mean]
    result[:median] = calculate_median(data) if operations[:median]

    result
  end

(...)
end
```

I do not understand how llm would know that operations cloud be `mean` and `median`, and that it should be a hash at all. I'm discussing it with an LLM (ironic, isn't it?) , and it says the example would not work. 
To make it work the description should be much more specific:
`"Specifies which analysis to perform. Should be an object with optional boolean keys: 'mean' and 'median'. For example: {'mean': true, 'median': false}. If omitted, both mean and median might be calculated."`.

I saw the issue #11 about Structured output and I think structured input would be nice to have too. 
What do you think about including this? 

I have no ideas on implementation or API for these yet, but I can try working on it.

crmne:
Thanks for catching this! You're right, we don't actually support complex parameter types like objects and arrays yet because support is inconsistent across providers. I've updated the docs to remove the misleading example in b0f513d999305a2fdc5d513d3ea41ec2fb827ac2

I think that just asking users to flattening params might be simpler to implement than full complex tool parameters. However, if you're interested in contributing to this feature, I'd be happy to discuss approaches.

bingoneedtum:
I may have found a solution to your problem. You can look here [click](https://javierlopez.eu/captha/) Hope this helped

jayelkaake:
Just posting here for visibility but this is solved by my PR @ https://github.com/crmne/ruby_llm/pull/124

tpaulshippy:
I don't think this is done yet.

JakeDluhy:
[...]

PR #112, opened 2025-04-14
Support nested parameters for tools

tpaulshippy:
## Purpose
Support calling tools that take arrays or nested objects.

## Approach
Simply define the items and properties as part of the Parameter class. Output hashes.

Related to #76 (although I didn't actually test enums). I realize this probably doesn't align perfectly with #65 or #90 but I kinda need this now and this seems to be the simplest approach.

## Testing
Tested with Bedrock, Gemini, and OpenAI - I do not currently have ready access to Anthropic or Deepseek.


tpaulshippy:
https://github.com/crmne/ruby_llm/issues/11#issuecomment-2815643231

Does this qualify? What am I missing?

tpaulshippy:
I guess I'm missing a bunch of the json schema by just supporting items and properties, like for example, enums.

jayelkaake:
This is great! Thanks for jumping on it!

>  Do you like [#11 (comment)](https://github.com/crmne/ruby_llm/issues/11#issuecomment-2814343621) better which adds "schema" as a parameter to Param?

I appreciate bringing it up as an option, but def recommend the `with_response_schema` approach over my monkey patch 😅

jayelkaake:
@tpaulshippy Thanks for putting this together!

Yeah, you're right, after thinking of it and using my monkey patch in an app, I'm realizing that the Parameters class is really pointless. The hash is a JSON schema, so it should stay as Hash. In the future when more complex JSON schema structures are supported it'd be better not to have to update the library (ie `maxItems`, `minItems`, etc)

The only "gotcha" here is that JSON follows camelCase so you should probably convert any underscore attributes to camelCase if they exist. but that shouldn't be too hard with `transformKeys` I think.

lib/ruby_llm/providers/anthropic/tools.rb
```diff
@@ -81,13 +81,22 @@ def parse_tool_calls(content_block)
 
         def clean_parameters(parameters)
           parameters.transform_values do |param|
-            {
-              type: param.type,
-              description: param.description
-            }.compact
+            clean_parameter(param)
```
jayelkaake, line 84: We should just let the parameter class return the schema. This way it is more forward compatible when LLMs add schema support (like minItems, etc)

Here's an implementation of RubyLLM::Parameter that is reverse compatible and can simply get `.schema` on the parameter:
```ruby
# In `lib/ruby_llm/providers/*/tools.rb`:
##
# @param parameters [Hash{String => Parameters}]
# @return [Hash]
def clean_parameters(parameters)
  parameters.transform_values(&:schema)
end
```

```ruby
# In `lib/ruby_llm/tool.rb`:
# ...
  class Parameter
    attr_reader :name, :required, :schema

    ##
    # If providing schema directly MAKE SURE TO USE STRING KEYS.
    # Also note that under_scored keys are NOT automatically transformed to camelCase.
    # @param name [Symbol]
    # @param schema [Hash{String => Object}, NilClass] 
    # @option type [String] (default: `'string'`)
    # @options required [Boolean] (default: `true`)
    # @options desc [String]
    def initialize(name, schema = nil, required: true, type: 'string', desc: nil)
      @name = name
      @schema = schema.to_h

      @schema['type'] ||= type
      @schema['description'] ||= desc if desc.present?
      @required = required
    end

    ##
    # @return [String]
    def type
      @schema['type']
    end

    ##
    # @return [String, NilClass]
    def description
      @schema['description']
    end

    alias required? required
  end
# ...
```

jayelkaake (commented):
Here's my idea for a reverse-compatible `Parameter` class that supports future schema. 

The LLMs change so often it'd be great not to have to update this library everytime a new schema feature is supported by the LLM. Also kind of simplifies the code.

lib/ruby_llm/providers/anthropic/tools.rb
tpaulshippy: Love it! Did you want to submit a separate PR or should I try to get it into this one?

jayelkaake: Uhh I'm realizing maybe I should just do a separate PR since I should test the code thoroughly. 

I realized my solution is now using string and symbol keys so just want to ensure we're not breaking anything 🤔

tpaulshippy: Feel free to borrow from this one as much as you need.

tpaulshippy:
Closing in favor of #124

PR #115, opened 2025-04-16
Module support

smart:
I wanted to nest my classes under Ai:: module and had issues with the foreign keys being generated from .underscore having a slash in it. Modified to fit. No new tests but old tests pass and only a few line changes.


crmne:
fixed in https://github.com/crmne/ruby_llm/commit/3a37b075a5e26bf96b2250e9d7397176749d1ccc

PR #122, opened 2025-04-18
Structured output & json mode

kieranklaassen:
Solves https://github.com/crmne/ruby_llm/issues/11

- Only OpenAI support with strict mode, all other support as well
- Added a strict: false flag for people to "experiment" with for non OpenAI
- Testing on prod with https://cora.computer/

```zsh
cora-app(dev)* schema = {
cora-app(dev)*   type: "object",
cora-app(dev)*   properties: {
cora-app(dev)*     name: { type: "string" },
cora-app(dev)*     age: { type: "integer" },
cora-app(dev)*     interests: { 
cora-app(dev)*       type: "array", 
cora-app(dev)*       items: { type: "string" }
cora-app(dev)*     }
cora-app(dev)*   },
cora-app(dev)*   required: ["name", "age", "interests"]
cora-app(dev)> }
cora-app(dev)> 
cora-app(dev)> # Get structured output as a Hash
cora-app(dev)> response = RubyLLM.chat(model: "gemini-2.0-flash")
cora-app(dev)> .with_output_schema(schema, strict:false)
cora-app(dev)> .ask("Create a profile for a Ruby developer").content
=> {"name" => "Ruby Developer Extraordinaire", "age" => 30, "interests" => ["Ruby on Rails", "Web Development", "REST APIs", "PostgreSQL", "Software Architecture", "Agile Methodologies"]}
cora-app(dev)* schema = {
cora-app(dev)*   type: "object",
cora-app(dev)*   properties: {
cora-app(dev)*     name: { type: "string" },
cora-app(dev)*     age: { type: "integer" },
cora-app(dev)*     interests: { 
cora-app(dev)*       type: "array", 
cora-app(dev)*       items: { type: "string" }
cora-app(dev)*     }
cora-app(dev)*   },
cora-app(dev)*   required: ["name", "age", "interests"]
cora-app(dev)> }
cora-app(dev)> 
cora-app(dev)> # Get structured output as a Hash
cora-app(dev)> response = RubyLLM.chat(model: "claude-3-7-sonnet-20250219")
cora-app(dev)> .with_output_schema(schema, strict:false)
cora-app(dev)> .ask("Create a profile for a Ruby developer").content
=> {"name" => "Alex Johnson", "age" => 32, "interests" => ["Ruby on Rails", "Web development", "Open source contribution", "Hiking", "Jazz music"]}
cora-app(dev)* schema = {
cora-app(dev)*   type: "object",
cora-app(dev)*   properties: {
cora-app(dev)*     name: { type: "string" },
cora-app(dev)*     age: { type: "integer" },
cora-app(dev)*     interests: { 
cora-app(dev)*       type: "array", 
cora-app(dev)*       items: { type: "string" }
cora-app(dev)*     }
cora-app(dev)*   },
cora-app(dev)*   required: ["name", "age", "interests"]
cora-app(dev)> }
cora-app(dev)> 
cora-app(dev)> # Get structured output as a Hash
cora-app(dev)> response = RubyLLM.chat(model: "gpt-4.1-nano")
cora-app(dev)> .with_output_schema(schema)
cora-app(dev)> .ask("Create a profile for a Ruby developer").content
=> {"name" => "Alex Johnson", "age" => 30, "interests" => ["Ruby programming", "Web development", "Open source contributions", "Agile methodologies", "Continuous integration"]}
cora-app(dev)> 
```


CLAUDE.md
```diff
@@ -0,0 +1,23 @@
+# CLAUDE.md
```
kieranklaassen, line 1: question: should we include or not?

lib/ruby_llm/model_info.rb
```diff
@@ -15,7 +15,7 @@ module RubyLLM
   class ModelInfo
     attr_reader :id, :created_at, :display_name, :provider, :metadata,
                 :context_window, :max_tokens, :supports_vision, :supports_functions,
-                :supports_json_mode, :input_price_per_million, :output_price_per_million, :type, :family
```
danielfriis, line 18: Is `json_mode` not a separate thing from structured output (albeit related)?

kieranklaassen: Yeah it is kind of, but I don't see us using it anywhere in the code. So I decided to remove/replace it for now. Thoughts?

danielfriis: True. Good point. Might want to add the functionality later.

lib/ruby_llm/chat.rb
```diff
@@ -78,6 +80,62 @@ def with_temperature(temperature)
       self
     end
 
+    # Specifies a JSON schema for structured output from the model
+    # @param schema [Hash, String] JSON schema as a Hash or JSON string
+    # @return [self] Returns self for method chaining
+    # @raise [ArgumentError] If the schema is not a Hash or valid JSON string
+    # @raise [UnsupportedStructuredOutputError] If the model doesn't support structured output
+    def with_output_schema(schema, strict: true)
```
danielfriis, line 88: I know @crmne has already given his pov on naming, but would it make sense to name this `with_response_format` instead? This is consistent with how OpenAI goes about it, and would allow for the a `json_mode` implementation later.

See here how the structured outputs vs json objects are implemented:
https://platform.openai.com/docs/guides/structured-outputs#structured-outputs-vs-json-mode

Structured output:
`response_format: { type: "json_schema", json_schema: {"strict": true, "schema": ...} }`

JSON mode:
`response_format: { type: "json_object" }"`

kieranklaassen: Yeah that is what I originally had so I agree! @crmne Open to changing the name?

kieranklaassen: Trying to make it a nice API as well since the OpenAI one is so confusing, how about this:


```rb
# OpenAI schema with structured output 
chat.with_response_format(schema)

# OpenAI JSON mode
chat.with_response_format(:json)

# Unofficial nonstrict mode (otherwise it will raise since it is not supported
chat.with_response_format(schema, strict: false)
chat.with_response_format(:json, strict: false)
```

Thoughts?

danielfriis: I had something similar in mind! With this approach, I would use either `schema` or `:json_object`.

I did think about this API also:

```ruby
chat.with_response_format(:json, schema: schema)
```

lib/ruby_llm/active_record/acts_as.rb
```diff
@@ -114,6 +114,17 @@ def with_temperature(temperature)
         self
       end
 
+      # Specifies the response format for the chat (JSON mode or JSON schema)
+      # @param response_format [Hash, String, Symbol] The response format, either:
+      #   - :json for simple JSON mode
+      #   - JSON schema as a Hash or JSON string for schema-based output
+      # @param strict [Boolean] Whether to enforce model compatibility (default: true)
+      # @return [self] Chainable chat instance
+      def with_response_format(response_format, strict: true)
```
jayelkaake, line 123: Can we add `alias with_response_schema with_response_format`?

lib/ruby_llm/chat.rb
kieranklaassen: I went with my route but like what you have too. I think what I have will make the API maybe less verbose and clenaer? Not sure. Imagine these will work in the future too:

```rb
# With a RubyLLM/Structify schema definition
chat.with_response_format(Candidate)

# With schema hash
chat.with_response_format(schema)

# With json mode
chat.with_response_format(:json)

# XML mode?
chat.with_response_format(:xml)
```

I like it's just one argument. @crmne let us know what you think, API design is important here

lib/ruby_llm/chat.rb
```diff
@@ -78,6 +80,112 @@ def with_temperature(temperature)
...
+    # Specifies the response format for the model
+    # @param response_format [Hash, String, Symbol] Either:
+    #   - :json symbol for JSON mode (model outputs valid JSON object)
+    #   - JSON schema as a Hash or JSON string for schema-based output (model follows the schema)
+    # @param strict [Boolean] Whether to enforce the model's support for the requested format
+    # @return [self] Returns self for method chaining
+    # @raise [ArgumentError] If the response_format is not a Hash, valid JSON string, or :json symbol
+    # @raise [UnsupportedJSONModeError] If :json is specified, strict is true, and the model doesn't support JSON mode
+    # @raise [UnsupportedStructuredOutputError] If a schema is specified, strict is true, and the model doesn't support structured output
+    def with_response_format(response_format, strict: true)
```
jayelkaake, line 92: Can you change the method signature to:
```ruby
def with_response_format(strict: true, **response_format)
``

Doing that makes the UI way nicer:
```ruby
chat.with_response_format(type: :number).ask("How old would MLK be today if he were still alive?")
```

jayelkaake (commented):
Looks amazing! My only request is changing the method signature (and an alias would be nice).

After doing that this is going to pair sooo nicely with my (and @tpaulshippy) PR for the params: https://github.com/crmne/ruby_llm/pull/124

CLAUDE.md
jayelkaake: Feels like yet another thing to manage IMO, but if you don't think it'll need to be modified often then sure, why not I guess?

lib/ruby_llm/active_record/acts_as.rb
jayelkaake: I just noticed the comments above that it was originally `with_response_schema`. Going to comment above there because I think your original instinct was much better!

lib/ruby_llm/chat.rb
jayelkaake: `json_mode` is likely to get deprecated by Open AI since it can be accomplished with `json_schema` if you just require `{ type: 'object' }`, so the primary use case should be structured outputs.

The most common use case is going to be with simple response schemas so we should focus on supporting the nicest UI in those cases. 

For example, `chat.with_response_schema(type: :integer)` or even we could add `chat.with_response_type(:integer)` to quickly accomplish the same thing.

That being said @kieranklaassen your idea of allowing a symbol `:json` is pretty good! However, in practice it feels like it violates SRP: `with_response_format(type = nil, strict: true, **schema)`

My vote is to just not support `json_mode` at all, only `json_schema`. Can anyone even think of a good use case for it over `json_schema`? What's the point of a JSON api that returns random JSON...

jayelkaake: Also, I think your original instinct of calling it `with_response_schema` was better. That naming is more consistent with the APIs and `response_format` typically means something else (xml vs json vs html).

I know you were considering merging the 2 types here but I think that violates [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle). 

I _do_ think you should add an alias anyways (in case our instincts were wrong)

CLAUDE.md
crmne: no

README.md
```diff
@@ -15,12 +15,9 @@ A delightful Ruby way to work with AI. No configuration madness, no complex call
...
-<a href="https://github.com/testdouble/standard"><img src="https://img.shields.io/badge/code_style-standard-brightgreen.svg" alt="Ruby Style Guide" /></a>
-<a href="https://rubygems.org/gems/ruby_llm"><img alt="Gem Downloads" src="https://img.shields.io/gem/dt/ruby_llm"></a>
-<a href="https://codecov.io/gh/crmne/ruby_llm"><img src="https://codecov.io/gh/crmne/ruby_llm/branch/main/graph/badge.svg" alt="codecov" /></a>
+<a href="https://badge.fury.io/rb/ruby_llm"><img src="https://badge.fury.io/rb/ruby_llm.svg" alt="Gem Version" /></a> <a href="https://github.com/testdouble/standard"><img src="https://img.shields.io/badge/code_style-standard-brightgreen.svg" alt="Ruby Style Guide" /></a> <a href="https://rubygems.org/gems/ruby_llm"><img alt="Gem Downloads" src="https://img.shields.io/gem/dt/ruby_llm"></a> <a href="https://codecov.io/gh/crmne/ruby_llm"><img src="https://codecov.io/gh/crmne/ruby_llm/branch/main/graph/badge.svg" alt="codecov" /></a>
 
-🤺 Battle tested at [💬  Chat with Work](https://chatwithwork.com)
+🤺 Battle tested at [💬 Chat with Work](https://chatwithwork.com)
```
crmne, line 20: these changes shouldn't happen

README.md
```diff
@@ -83,6 +81,28 @@ class Weather < RubyLLM::Tool
...
+  required: ["name", "age", "interests"]
+}
+
+# Returns a validated Hash instead of plain text
+user_data = chat.with_response_format(schema).ask("Create a profile for a Ruby developer")
+
+# Access the structured data using hash keys
+puts "Name: #{user_data.content['name']}"              # => "Jane Smith"
+puts "Age: #{user_data.content['age']}"                # => 32
+puts "Interests: #{user_data.content['interests'].join(', ')}"  # => "Ruby, Rails, API design"
```
crmne, line 105: I feel like this is starting to get too verbose for the README. Let's leave it out of here. People can still find it in the docs.

docs/guides/rails.md
```diff
@@ -174,6 +175,89 @@ system_message = chat_record.messages.find_by(role: :system)
 puts system_message.content # => "You are a concise Ruby expert."
 ```
 
+## Working with Structured Output
```
crmne, line 178: Use

```markdown
{: .d-inline-block }

New (v1.3.0)
{: .label .label-green }
```

crmne: Also link the structured output guide here.

docs/guides/rails.md
```diff
@@ -174,6 +175,89 @@ system_message = chat_record.messages.find_by(role: :system)
...
+# app/models/message.rb
+class Message < ApplicationRecord
+  acts_as_message
+  serialize :content, JSON # Add this for text columns
+end
+```
+
+### Using Structured Output
+
+The `with_response_format` method is available on your `Chat` model thanks to `acts_as_chat`:
```
crmne, line 212: Hmm maybe `with_response_format` is better than `with_output_schema`...

docs/guides/structured-output.md
```diff
@@ -0,0 +1,160 @@
+---
+layout: default
+title: Structured Output
+parent: Guides
+nav_order: 7
+---
+
+# Structured Output
+
+RubyLLM allows you to request structured data from language models by providing a JSON schema. When you use the `with_response_format` method, RubyLLM will ensure the model returns data matching your schema instead of free-form text.
```
crmne, line 10: try to follow the new format for the docs, e.g.:

```markdown
# Configuring RubyLLM
{: .no_toc }

This guide covers all the configuration options available in RubyLLM, from setting API keys and default models to customizing connection behavior and using scoped contexts.
{: .fs-6 .fw-300 }

## Table of contents
{: .no_toc .text-delta }

1. TOC
{:toc}

---

After reading this guide, you will know:

*   How to set up global configuration using `RubyLLM.configure`.
*   How to configure API keys for different providers.
*   How to set default models for chat, embeddings, and image generation.
*   How to customize connection timeouts and retries.
*   How to connect to custom endpoints (like Azure OpenAI).
*   How to use temporary, scoped configurations with `RubyLLM.context`.
```

docs/guides/structured-output.md
```diff
@@ -0,0 +1,160 @@
...
+
+```ruby
+response = RubyLLM.chat(model: "gpt-4.1-nano")
+  .with_response_format(:json)
+  .ask("Create a profile for a Ruby developer")
+```
+
+This uses OpenAI's `response_format: {type: "json_object"}` parameter, works with most OpenAI models, and guarantees valid JSON without enforcing a specific structure.
+
+## Strict and Non-Strict Modes
```
crmne, line 50: I'm not convinced about the naming here. For example for skipping the model registry check I used `assume_model_exists` which clarifies the intention of the flag. The first time I read `strict` I had no idea what the API was gonna be strict about? The output format? How?

I would go for `assume_supported: true`

README.md
```diff
@@ -36,6 +36,7 @@ RubyLLM fixes all that. One beautiful API for everything. One consistent format.
 - 🖼️ **Image generation** with DALL-E and other providers
 - 📊 **Embeddings** for vector search and semantic analysis
 - 🔧 **Tools** that let AI use your Ruby code
+- 📝 **Structured Output** with JSON schema validation
```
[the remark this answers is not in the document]
crmne: this is not correct as we're not validating anything

docs/guides/structured-output.md
```diff
@@ -0,0 +1,160 @@
...
+## Error Handling
+
+RubyLLM provides two main error types for structured output:
+
+1. **UnsupportedStructuredOutputError**: Raised when using schema-based output with a model that doesn't support it in strict mode:
+2. **InvalidStructuredOutput**: Raised if the model returns invalid JSON:
+
+Note: RubyLLM checks that responses are valid JSON but doesn't verify conformance to the schema structure. For full schema validation, use a library like `json-schema`.
+
+## With ActiveRecord and Rails
```
crmne, line 81: this is repeated. instead just link the correct section

lib/ruby_llm/active_record/acts_as.rb
crmne: I like `with_response_format`, let's stick to that.

lib/ruby_llm/chat.rb
crmne: `with_response_format(response_format, assume_supported: false)`

lib/ruby_llm/chat.rb
```diff
@@ -78,6 +80,112 @@ def with_temperature(temperature)
...
+      self
+    end
+
+    private
+
+    # Normalizes the schema to a standard format
+    # @param response_format [Hash, String] JSON schema as a Hash or JSON string
+    # @return [Hash] Normalized schema as a Hash
+    # @raise [ArgumentError] If the response_format is not a Hash or valid JSON string
+    def normalize_schema(response_format)
```
crmne, line 113: `normalize_response_format`

lib/ruby_llm/chat.rb
```diff
@@ -78,6 +80,112 @@ def with_temperature(temperature)
...
+      raise ArgumentError, 'Response format must be a Hash' unless schema_obj.is_a?(Hash)
+
+      schema_obj
+    end
+
+    # Checks if the model supports the requested format (JSON mode or schema)
+    # @param is_json_mode [Boolean] Whether JSON mode is being used
+    # @raise [UnsupportedJSONModeError] If JSON mode is requested but not supported
+    # @raise [UnsupportedStructuredOutputError] If structured output is requested but not supported
+    def check_model_compatibility!(is_json_mode)
```
crmne, line 126: I'd prefer two methods:
`ensure_response_format_support`
`ensure_json_mode_support`

lib/ruby_llm/chat.rb
```diff
@@ -78,6 +80,112 @@ def with_temperature(temperature)
...
+        # Append to existing system message
+        updated_content = "#{system_message.content}\n\n#{guidance}"
+        @messages.delete(system_message)
+        add_message(role: :system, content: updated_content)
+      else
+        # No system message exists, create a new one
+        with_instructions(guidance)
+      end
+    end
+
```
crmne, line 188: are you absolutely sure we need this? I would be surprised if the official Python SDK would use system prompts to implement `response_format`: https://github.com/search?q=repo%3Aopenai%2Fopenai-python+response_format&type=code&p=0

lib/ruby_llm/model_info.rb
crmne: I agree with removing that.

lib/ruby_llm/provider.rb
```diff
@@ -10,7 +10,7 @@ module Provider
     module Methods # rubocop:disable Metrics/ModuleLength
       extend Streaming
 
-      def complete(messages, tools:, temperature:, model:, &block) # rubocop:disable Metrics/MethodLength
+      def complete(messages, tools:, temperature:, model:, chat: nil, &block) # rubocop:disable Metrics/MethodLength
```
crmne, line 13: 😮

crmne (changes requested):
Thank you so much for the effort @kieranklaassen but I feel like the PR needs significant work, in particular:

1. Is Structured Output really implemented with system prompts in the real SDKs?
2. Passing the `chat` object around seems excessive and violates many SE principles. I stopped reviewing there.

You can find more comments there.

lib/ruby_llm/chat.rb
danielfriis: We don't know if it will get deprecated and as long as it's supported, it makes sense to have RubyLLM allow for easy access to that endpoint...

The use case shouldn't matter, but I've been using it to extract metadata from various objects where I don't know exactly what to extract.

crmne: I like `with_response_format`

jayelkaake: > The use case shouldn't matter, but I've been using it to extract metadata from various objects where I don't know exactly what to extract.
Doesn't `with_response_format(type: :object)` cover your case? 

I am still open to being proven wrong, but so far:
* json_mode doesn't cover any cases that json_schema doesn't already, and
* it isn't supported by other APIs, and
* adds unnecessary complexity

This gem is great because it reduces complexity. We shouldn't compromise that for completion only for the sake of completion.

Agree with `with_response_format` though since others prefer it 👍

Your call though obv!

lib/ruby_llm/chat.rb
jayelkaake: @crmne using destructuring makes the method API much nicer:

```ruby
with_response_format(type = nil, assume_supported: false, **response_schema)
```

Anyways, mixed type variables are a code smell, so this method signature solves that too.

lib/ruby_llm/chat.rb
jayelkaake: I was thinking about this, and you could even just support `with_response_format(:json)` which under the hood would use structured outputs `type: :object`.

Best of both worlds, more cross-api compatible, same result!

jayelkaake:
> 2\. Passing the `chat` object around seems excessive and violates many SE principles. I stopped reviewing there.

@kieranklaassen let me know if you want help with updating this to not pass the chat object around. I agree with @crmne and want to move your great initiative along :)

I guess conflicts need to be resolved too now.

lib/ruby_llm/chat.rb
kieranklaassen: We need this yes. OpenAI will give an error if you do not have these instructions in the prompt. And for all other providers there is no API so we need them.

lib/ruby_llm/active_record/acts_as.rb
kieranklaassen: Also, thjis might be a schema or a format. JSON mode does not have a schema.

kieranklaassen:
@crmne Thanks for the review. I pushed the changes requested. Do you still feel this needs a lot of work? If yes, could you explain your feeling so I can try to understand what needs to be done?

lib/ruby_llm/chat.rb
danielfriis: To be precise, it's only necessary in JSON mode, not for Structured Output. Personally, I'd prefer to have control over how prompts are formulated. Keep in mind, they may also be in another language.

kieranklaassen: Fixed!

lib/ruby_llm/chat.rb
jayelkaake: After doing some research I can see that using `json_schema` mode _requires_ the `properties` keys to work with OpenAI's API, so that's the main difference.

I still wonder if that logic should be abstracted-away here? 

under the hood the provider can switch between json_mode and json_schema, but at the top level the user doesn't really care. They just want `with_response_format(:object)` or something.

Agree with `with_response_format` though since others prefer it 👍

jayelkaake:
FYI I decided to take a stab at this and took a slightly different approach (a bit more minimal) https://github.com/crmne/ruby_llm/pull/131

Let me know what you guys think!

@kieranklaassen or if you want want to merge our PRs we can do that too. We're in this together 😄

PR #124, opened 2025-04-19
Support Complex Parameter Schemas

jayelkaake:
# Summary
Adds the ability to define more complex param schemas, while retaining the existing simple methods.

This new format is also forward-compatible once LLMs start supporting more JSON schema structures (and at the pace they are moving that might not be too long from now).

Bumped version to 1.3 since this is kind of a major feature.

# Examples
```ruby
# Existing structure still works and now is called 'shorthand' syntax:
param :latitude, desc: "Latitude (e.g., 52.5200)"

# But now you can treat the params as plain schema definitions. It no longer uses object params EXCEPT for `required` key:
param :unit, type: :string, enum: %w[f c], description: "Unit (e.g., celsius, fahrenheit)", required: false

# It now supports object params
param :location, type: :object, desc: "Country and city.", required: false, properties: {
  country: { type: :string, description: "Full name of the country." },
  city: { type: :string, description: "Full name of the city." }
}

# And also supports array params:
param :location, type: :array, desc: "Latitude and longitude", items: { type: :number, description: "Float value" }
```

# How I tested
I tested all the supported LLM APIs directly.

# Shout-outs
* @tpaulshippy for providing test code for this and starting the initial PR https://github.com/crmne/ruby_llm/pull/112
* Everyone who contributed to the thread at https://github.com/crmne/ruby_llm/issues/11

PR #131, opened 2025-04-21
Structured Output & JSON mode response support

jayelkaake:
# Features
✅   Supports all features of `json_schema` with minimal text.
✅   Automatically switches to `json_mode`. No need to think about it. Just ask for JSON.
✅   Does not require maintenance when the APIs inevitably increase their support
✅   Thread-safe implementation
✅   Full test coverage
❌   To reduce initial complexity, Anthropic, Deepseek and Gemini are not yet supported. I can add this in another PR. 

# Sample Code
You can now ensure the responses follow a schema you define like this:
```ruby
chat = RubyLLM.chat

chat.with_response_format(:integer).ask("What is 2 + 2?").content
# => 4

chat.with_response_format(:string).ask("Say 'Hello World' and nothing else.").content
# => "Hello World"

chat.with_response_format(:array, items: { type: :string })
chat.ask('What are the 2 largest countries? Only respond with country names.').content
# => ["Russia", "Canada"]

chat.with_response_format(:object, properties: { age: { type: :integer } })
chat.ask('Provide sample customer age between 10 and 100.').content
# => { "age" => 42 }

chat.with_response_format(
  :object,
  properties: { hobbies: { type: :array, items: { type: :string, enum: %w[Soccer Golf Hockey] } } }
)
chat.ask('Provide at least 1 hobby.').content
# => { "hobbies" => ["Soccer"] }
```

You can also provide the JSON schema you want directly to the method like this:
```ruby
chat.with_response_format(type: :object, properties: { age: { type: :integer } })
# => { "age" => 31 }
```

In this example the code is automatically switching to OpenAI's json_mode since no object properties are requested:
```ruby
chat.with_response_format(:json) # Don't care about structure, just give me JSON

chat.ask('Provide a sample customer data object with name and email keys.').content
# => { "name" => "Tobias", "email" => "tobias@example.com" }

chat.ask('Provide a sample customer data object with name and email keys.').content
# => { "first_name" => "Michael", "email_address" => "michael@example.com" }
```

# How was it tested?
* Tested all the models, however currently it only works for OpenAI. I added specs for OpenAI.

# Shout-outs
* @kieranklaassen for https://github.com/crmne/ruby_llm/pull/122 which I took some inspiration from
* @crmne for everything
* Everyone who commented on https://github.com/crmne/ruby_llm/issues/11 to provide input on preferred structures.

# Related
* Issue: https://github.com/crmne/ruby_llm/issues/11 
* My other PR for Parameters: https://github.com/crmne/ruby_llm/pull/124 (NOTE: I'll merge `Schema` class once either PR is merged)

### DIFF
--- a/lib/ruby_llm/active_record/acts_as.rb
+++ b/lib/ruby_llm/active_record/acts_as.rb
@@ -93,6 +93,12 @@ module RubyLLM
         self
       end
 
+      # @see LlmChat#with_response_format
+      def with_response_format(...)
+        to_llm.with_response_format(...)
+        self
+      end
+
       def with_tool(...)
         to_llm.with_tool(...)
         self
@@ -158,14 +164,19 @@ module RubyLLM
         end
 
         transaction do
-          @message.update!(
-            role: message.role,
-            content: message.content,
-            model_id: message.model_id,
-            tool_call_id: tool_call_id,
-            input_tokens: message.input_tokens,
-            output_tokens: message.output_tokens
-          )
+          # These are required fields:
+          @message.role = message.role
+          @message.content = message.content
+
+          # These are optional fields:
+          @message.try('model_id=', message.model_id)
+          @message.try('tool_call_id=', tool_call_id)
+          @message.try('input_tokens=', message.input_tokens)
+          @message.try('output_tokens=', message.output_tokens)
+          @message.try('content_schema=', message.content_schema)
+
+          @message.save!
+
           persist_tool_calls(message.tool_calls) if message.tool_calls.present?
         end
       end
--- a/lib/ruby_llm/chat.rb
+++ b/lib/ruby_llm/chat.rb
@@ -31,6 +31,56 @@ module RubyLLM
       }
     end
 
+    ##
+    # This method lets you ensure the responses follow a schema you define like this:
+    #
+    #   chat.with_response_format(:integer).ask("What is 2 + 2?").to_i
+    #   # => 4
+    #   chat.with_response_format(:string).ask("Say 'Hello World' and nothing else.").content
+    #   # => "Hello World"
+    #   chat.with_response_format(:array, items: { type: :string })
+    #   chat.ask('What are the 2 largest countries? Only respond with country names.').content
+    #   # => ["Russia", "Canada"]
+    #   chat.with_response_format(:object, properties: { age: { type: :integer } })
+    #   chat.ask('Provide sample customer age between 10 and 100.').content
+    #   # => { "age" => 42 }
+    #   chat.with_response_format(
+    #     :object,
+    #     properties: { hobbies: { type: :array, items: { type: :string, enum: %w[Soccer Golf Hockey] } } }
+    #   )
+    #   chat.ask('Provide at least 1 hobby.').content
+    #   # => { "hobbies" => ["Soccer"] }
+    #
+    # You can also provide the JSON schema you want directly to the method like this:
+    #   chat.with_response_format(type: :object, properties: { age: { type: :integer } })
+    #   # => { "age" => 31 }
+    #
+    # In this example the code is automatically switching to OpenAI's json_mode since no object
+    # properties are requested:
+    #   chat.with_response_format(:json) # Don't care about structure, just give me JSON
+    #   chat.ask('Provide a sample customer data object with name and email keys.').content
+    #   # => { "name" => "Tobias", "email" => "tobias@example.com" }
+    #   chat.ask('Provide a sample customer data object with name and email keys.').content
+    #   # => { "first_name" => "Michael", "email_address" => "michael@example.com" }
+    #
+    # @param type [Symbol] (optional) This can be anything supported by the API JSON schema types (integer, object, etc)
+    # @param schema [Hash] The schema for the response format. It can be a JSON schema or a simple hash.
+    # @return [Chat] (self)
+    def with_response_format(type = nil, **schema)
+      schema_hash = if type.is_a?(Symbol) || type.is_a?(String)
+                      { type: type == :json ? :object : type }
+                    elsif type.is_a?(Hash)
+                      type
+                    else
+                      {}
+                    end.merge(schema)
+
+      @response_schema = Schema.new(schema_hash)
+
+      self
+    end
+    alias with_structured_response with_response_format
+
     def ask(message = nil, with: {}, &)
       add_message role: :user, content: Content.new(message, with)
       complete(&)
@@ -86,17 +136,23 @@ module RubyLLM
 
     def complete(&) # rubocop:disable Metrics/MethodLength
       @on[:new_message]&.call
-      response = @provider.complete(
-        messages,
-        tools: @tools,
-        temperature: @temperature,
-        model: @model.id,
-        connection: @connection,
-        &
-      )
+      response = @provider.with_response_schema(@response_schema) do
+        @provider.complete(
+          messages,
+          tools: @tools,
+          temperature: @temperature,
+          model: @model.id,
+          connection: @connection,
+          &
+        )
+      end
+
       @on[:end_message]&.call(response)
 
       add_message response
+
+      @response_schema = nil # Reset the response schema after completion of this chat thread
+
       if response.tool_call?
         handle_tool_calls(response, &)
       else
--- a/lib/ruby_llm/provider.rb
+++ b/lib/ruby_llm/provider.rb
@@ -31,6 +31,29 @@ module RubyLLM
         parse_list_models_response response, slug, capabilities
       end
 
+      ##
+      # @return [::RubyLLM::Schema, NilClass]
+      def response_schema
+        Thread.current['RubyLLM::Provider::Methods.response_schema']
+      end
+
+      ##
+      # @param response_schema [::RubyLLM::Schema]
+      def with_response_schema(response_schema)
+        prev_response_schema = Thread.current['RubyLLM::Provider::Methods.response_schema']
+
+        result = nil
+        begin
+          Thread.current['RubyLLM::Provider::Methods.response_schema'] = response_schema
+
+          result = yield
+        ensure
+          Thread.current['RubyLLM::Provider::Methods.response_schema'] = prev_response_schema
+        end
+
+        result
+      end
+
       def embed(text, model:, connection:, dimensions:)
         payload = render_embedding_payload(text, model:, dimensions:)
         response = connection.post(embedding_url(model:), payload)
--- a/lib/ruby_llm/schema.rb
+++ b/lib/ruby_llm/schema.rb
@@ -1,0 +1,76 @@
+# frozen_string_literal: true
+
+module RubyLLM
+  ##
+  # Schema class for defining the structure of data objects.
+  # Wraps the #Hash class
+  # @see #Hash
+  class Schema
+    delegate_missing_to :@schema
+
+    def initialize(schema = {})
+      @schema = deep_transform_keys_in_object(schema.to_h.dup, &:to_sym)
+    end
+
+    def [](key)
+      @schema[key.to_sym]
+    end
+
+    def []=(key, new_value)
+      @schema[key.to_sym] = deep_transform_keys_in_object(new_value, &:to_sym)
+    end
+
+    # Adds the new_value into the new_key key for every sub-schema that is of type: :object
+    # @param new_key [Symbol] The key to add to each object type.
+    # @param new_value [Boolean, String] The value to assign to the new key.
+    def add_to_each_object_type!(new_key, new_value)
+      add_to_each_object_type(new_key, new_value, @schema)
+    end
+
+    # @return [Boolean]
+    def present?
+      @schema.present? && @schema[:type].present?
+    end
+
+    private
+
+    def add_to_each_object_type(new_key, new_value, schema)
+      return schema unless schema.is_a?(Hash)
+
+      if schema[:type].to_s == :object.to_s
+        add_to_object_type(new_key, new_value, schema)
+      elsif schema[:type].to_s == :array.to_s && schema[:items]
+        schema[:items] = add_to_each_object_type(new_key, new_value, schema[:items])
+      end
+
+      schema
+    end
+
+    def add_to_object_type(new_key, new_value, schema)
+      if schema[new_key.to_sym].nil?
+        schema[new_key.to_sym] = new_value.is_a?(Proc) ? new_value.call(schema) : new_value
+      end
+
+      schema[:properties]&.transform_values! { |value| add_to_each_object_type(new_key, new_value, value) }
+    end
+
+    ##
+    # Recursively transforms keys in a hash or array to symbols.
+    # Borrowed from ActiveSupport's Hash#deep_transform_keys
+    # @param object [Object] The object to transform.
+    # @param block [Proc] The block to apply to each key.
+    # @return [Object] The transformed object.
+    def deep_transform_keys_in_object(object, &block)
+      case object
+      when Hash
+        object.each_with_object({}) do |(key, value), result|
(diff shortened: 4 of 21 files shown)


lib/ruby_llm/schema.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    def [](key)
+      @schema[key.to_sym]
+    end
+
+    def []=(key, new_value)
+      @schema[key.to_sym] = deep_transform_keys_in_object(new_value, &:to_sym)
+    end
+
+    def add_to_each_object_type!(new_key, new_value)
```
jayelkaake, line 26: TODO @jayelkaake: Add doc for this method

danielfriis:
I like this approach! Very simple implementation.

danielfriis:
Take a look at #90. It was a schema builder, I thought about adding, similar to how tools are defined.

lib/ruby_llm/chat.rb
```diff
@@ -86,14 +136,16 @@ def each(&)
...
       @on[:new_message]&.call
-      response = @provider.complete(
-        messages,
-        tools: @tools,
-        temperature: @temperature,
-        model: @model.id,
-        connection: @connection,
-        &
-      )
+      response = @provider.with_response_schema(@response_schema) do
```
jayelkaake, line 139: @crmne should we reset the @response_schema after the completion so it doesn't apply to subsequent messages in the chat?

crmne: no, `with_*` applies to all consequent messages

lib/ruby_llm/active_record/acts_as.rb
```diff
@@ -158,14 +164,19 @@ def persist_message_completion(message) # rubocop:disable Metrics/AbcSize,Metric
...
+
+          # These are optional fields:
+          @message.try('model_id=', message.model_id)
+          @message.try('tool_call_id=', tool_call_id)
+          @message.try('input_tokens=', message.input_tokens)
+          @message.try('output_tokens=', message.output_tokens)
+          @message.try('content_schema=', message.schema)
+
+          @message.save!
+
```
crmne: what has this to do with structured outputs?

lib/ruby_llm/provider.rb
```diff
@@ -31,6 +31,29 @@ def list_models(connection:)
...
+          Thread.current['RubyLLM::Provider::Methods.response_schema'] = response_schema
+
+          result = yield
+        ensure
+          Thread.current['RubyLLM::Provider::Methods.response_schema'] = prev_response_schema
+        end
+
+        result
+      end
+
```
crmne, line 56: what's all this about threads? we have contexts now

jayelkaake:
FYI @crmne I've been battle-testing this code at my company Osello (which is also a sponsor now of this project) and making tweaks over the last week.

If you took a look early last week, please take a fresh look now (same with my PR #124)

kieranklaassen:
You rock @jayelkaake !

lib/ruby_llm/chat.rb
jayelkaake: @sirwolfgang good idea!

Although `as` might be a bit too ambiguous. Maybe `.with_next_response_in_format(...)` or something like that?

sirwolfgang: @jayelkaake I think linguistically/ergonomically it would be better to split and not use the `.with` prefix at all, making it a shorter scan/parse. 

Totally open to ideas other than `.as`, but also curious to what you think it might collide with? I don't think it's structurally more ambiguous than `.with`. The only other `as` convention I can think of is rails routes, so I also don't think it's mnemonically overloaded.

Could extend it to `.as_format`, `.as_type`, `.as_response_type` or something else, if we want to preserve the root namespace.

Otherwise, if we could delay execution like that of AR, I can see the argument for making it a post setting; something like:
```ruby
agent.ask("...?").in(type: :integer)
agent.ask("...?").as(type: :integer)
agent.ask("...?").structured_as(type: :integer)
```

jayelkaake: I like the idea, just doesn't read well with English "ask someone to do something as...".... and also it should be clear you're not modifying the _query_ like it is with AR, it's modifying the _response_.

I like the `agent.ask("...?").[and_]respond_with(:integer)` postfix format better but I think that might require some major updates to this library I think.

Most APIs don't let you mutate the response format in real-time, so LLMs are kind of introducing the need for a new pattern maybe? I've been scratching my head about these things a lot over the last couple weeks! 😅

sirwolfgang: Yeah, I think refactoring to support a more dynamic method chaining should be a different PR; But we could setup the expect syntax here, and build towards that; Since it should functionally work in either order.
`agent.<token>.ask("...?") => agent.ask("...?").<token>`

response/d feels a little weird to me. I could see this interface also make sense for loading personas; like `respond_as(:support_agent)`. _Which might be the process of chaining assistants for processing. Like_ 
```ruby
timekeeper = RubyLLM.chat.with_tool(TIME)
groot = RubyLLM.chat.with_instructions(GROOT)

timekeeper.ask("What time is it?").respond_as(groot) => "I am grooooooot"
```

sirwolfgang:
LMK if there's anything I can contribute to this to help get it over the link @crmne @jayelkaake. I'd like to get this in along with https://github.com/crmne/ruby_llm/pull/144 to move forward on the path of using structured responses with OpenRouter

lib/ruby_llm/active_record/acts_as.rb
```diff
@@ -158,14 +164,19 @@ def persist_message_completion(message) # rubocop:disable Metrics/AbcSize,Metric
...
+          # These are required fields:
+          @message.role = message.role
+          @message.content = message.content
+
+          # These are optional fields:
+          @message.try('model_id=', message.model_id)
+          @message.try('tool_call_id=', tool_call_id)
+          @message.try('input_tokens=', message.input_tokens)
+          @message.try('output_tokens=', message.output_tokens)
+          @message.try('content_schema=', message.schema)
```
tkoenig, line 176: @jayelkaake I think this should be ` @message.try('content_schema=', message.content_schema)`, right?

jayelkaake: Yes, thanks for pointing out @tkoenig 

It was actually a merge problem for me working with multiple branches since this one hasn't been reviewed yet 😛 (was fixed in my dev branch). I think I need to stop making changes for a little while until it gets a review so I'm not just writing vapourware.

Eric-Guo:
@jayelkaake I just do a rebase in main and [trick part like this](https://github.com/crmne/ruby_llm/commit/1bc2b85bfad700f89595a185d4ae5f697873f219#diff-befd09844859a119cf3b5d4cd47a5eb2020d61e089ad9e62f92938d0c84f56ceR27), FYI and hoping this PR merged soon.

lib/ruby_llm/chat.rb
```diff
@@ -31,6 +31,56 @@ def initialize(model: nil, provider: nil, assume_model_exists: false, context: n
...
+    #   chat.with_response_format(
+    #     :object,
+    #     properties: { hobbies: { type: :array, items: { type: :string, enum: %w[Soccer Golf Hockey] } } }
+    #   )
+    #   chat.ask('Provide at least 1 hobby.').content
+    #   # => { "hobbies" => ["Soccer"] }
+    #
+    # You can also provide the JSON schema you want directly to the method like this:
+    #   chat.with_response_format(type: :object, properties: { age: { type: :integer } })
+    #   # => { "age" => 31 }
```
mauri, line 56: @jayelkaake @crmne Thanks for all the work put in this! I've been trying the API and found it to be a bit surprising and adds complexity on top of OpenAI nuances.

Comparing the calls
- `chat.with_response_format(:object, properties: { age: { type: :integer } })`
- `chat.with_response_format(type: :object, properties: { age: { type: :integer } })` 
The inclusion or not of the `type:` changes a lot the way we invoke OpenAI. One relies on json_mode and the other on json_schema but it's not very clear from the API.

In addition, the support of structured output or json_mode depends also on the model used. Old models will not support json_schema so maybe that's something we want to factor in? I.e. use one or the other based on the model.

OpenAI doesn't recommend JSON mode except for older models; and I do believe its API  a product of its time. The fact that we have to append more instructions to the original prompt and ask for json is a sign of that (why do it in english btw?), so I don't think RubyLLM should default to it relegating the response format `json_schema`.

lib/ruby_llm/chat.rb
```diff
@@ -31,6 +31,56 @@ def initialize(model: nil, provider: nil, assume_model_exists: false, context: n
...
+    #   # => { "name" => "Tobias", "email" => "tobias@example.com" }
+    #   chat.ask('Provide a sample customer data object with name and email keys.').content
+    #   # => { "first_name" => "Michael", "email_address" => "michael@example.com" }
+    #
+    # @param type [Symbol] (optional) This can be anything supported by the API JSON schema types (integer, object, etc)
+    # @param schema [Hash] The schema for the response format. It can be a JSON schema or a simple hash.
+    # @return [Chat] (self)
+    def with_response_format(type = nil, **schema)
+      schema_hash = if type.is_a?(Symbol) || type.is_a?(String)
+                      { type: type == :json ? :object : type }
```
mauri, line 71: isn't this :json_object? [0]

[0] https://platform.openai.com/docs/guides/structured-outputs?api-mode=chat#json-mode

joshmlewis:
Thank you for the work on this PR. It's definitely a common use case I was sad to see wasn't in the gem already. Anything I can do to help get it over the finish line?

crmne (changes requested):
Thanks for the work on this @jayelkaake, and for the sponsorship support. I've taken a close look at this PR and while I appreciate the effort, I don't think this API design fits with RubyLLM's philosophy.

**The core issue:** This approach is inconsistent with how RubyLLM handles structured data elsewhere. Compare your proposed API:

```ruby
chat.with_response_format(:object, properties: { hobbies: { type: :array, items: { type: :string, enum: %w[Soccer Golf Hockey] } } })
```

To how we handle Tools:

```ruby
class Weather < RubyLLM::Tool
  description "Gets current weather for a location"
  param :latitude, desc: "Latitude (e.g., 52.5200)"
  param :longitude, desc: "Longitude (e.g., 13.4050)"
  
  def execute(latitude:, longitude:)
    # implementation
  end
end
```

The Tools API is self-documenting, reusable, and feels Ruby-native. Your approach is essentially raw JSON Schema dumped into method calls, which I have to backtrack on.

**Additional concerns:**

1. **Parameter overloading is confusing** - sometimes a symbol, sometimes symbol + hash, sometimes just hash
2. **Thread.current conflicts with Context** - we just shipped Context in 1.3.0 specifically to avoid thread-local state
3. **Magic behavior switching** between `json_schema` and `json_object` modes is too implicit

**What structured responses could look like in RubyLLM:**

(need to think this through)

```ruby
class CustomerData < RubyLLM::Schema
  string :name, required: true
  string :email
  integer :age, range: 18..100
end

class HobbyList < RubyLLM:Schema
  array :hobbies, items: { type: :string, enum: %w[Soccer Golf Hockey] }
end

# Usage
response = chat.with_schema(CustomerData).ask("Generate a customer")
# or
response = chat.ask("Generate a customer", schema: CustomerData)

puts response.name      # => "John Doe"
puts response.email     # => "john@example.com" 
puts response.age       # => 42
```

This approach would be:
- Consistent with our existing patterns
- Reusable across conversations
- Self-documenting
- Properly integrated with Context instead of Thread.current

I know this means starting over, but I'd rather ship something that feels like RubyLLM than rush a feature that fights the library's design. When structured responses land, they need to feel like they belong.

Given the strong community interest in this feature (as seen in #11), I'm prioritizing getting this right. Would you be interested in exploring this class-based approach, or would you prefer I handle the implementation?

jayelkaake:
Hey @crmne thanks for the feedback!

I'd be happy to update the PR with your suggestions.

Adding the context system is amazing. I was going to suggest that, so now that it exists it'll greatly simplify my implementation.

I'll work on it and submit my update for re-review.

lib/ruby_llm/provider.rb
jayelkaake: This was a thread-safe way to set response schema. 

I'll be able to refactor it to use context instead (assuming the context system thread-safe, haven't looked yet).

lib/ruby_llm/active_record/acts_as.rb
jayelkaake: There are 2 things here depending on what you're asking about @crmne :

### Why use `try`?
The [repo documentation here](https://rubyllm.com/guides/rails#setting-up-your-rails-application) says that there are required fields and the other fields are optional. 

Either docs need to be changed and a major version increased to v2 to indicate semantic versioning reverse-incompatibility, or using `try` lets it stay optional for now.

### Why store the schema in the DB?
Since the LLMs send the entire thread on each request they also require the schema for each message sent. Without it, you can't continue conversations that were stored in the DB a later date.

In fact, I also later realized that the tool call param schemas need to be stored as well for the same reason.

There are probably ways to not store the schema and instead link it to a class locally, but that would break all past conversations if the response schema in the code changed.

@crmne let me know if you're wondering about something else.

danielfriis:
FYI. Did a bit of work towards class-defined schemas here https://github.com/crmne/ruby_llm/pull/90

lib/ruby_llm/active_record/acts_as.rb
crmne: We don't need to use a `try` as the `message` referenced there is a `Message` instance which sets these parameters to `nil` or a suitable default in case they don't exist.

Totally understand the schema needing to be in the message and up for that.

jayelkaake: @crmne Unless fallbacks were added to to the `ActsAs` concern _after_ this commit, you will get an error on save if you follow the docs exactly for the Rails models and don't add the `output_tokens` column to your Rails model. 

That's why the `try` is needed. You can try it for yourself if you want.

`@message.class != message.class` in this method - maybe that's what's causing confusion?  I'm trying to add more docs to methods to reduce that confusion in my fork (which I'll add in to this PR)

Let me know if I missed something here.

crmne: Ah I see what you mean now! In the docs I have mistakenly kept the comment that says that these fields are optional. My bad. Fixing.

jayelkaake: Haha thanks - if you're updating the docs I'll just remove the `try`s because others would not have been able to proceed anyways with basic functionality if they ran into the same issue as me. Do you agree?

crmne: I did https://github.com/crmne/ruby_llm/commit/c48c8ebda05c41acd17614afde90af37e9cc106c

That's a good idea. Let's remove the `try`s.

Eric-Guo:
Glad new [`with_params`](https://github.com/thape-cn/web/commit/0c2fa58e850b262d4b52a513499afa4996756230) as workaround and looking forward to #11 full support!

crmne:
Thank you for this PR\! Your ideas have been incorporated into the structured output implementation that just landed in be85257.

I've implemented many of your suggestions:
- Automatic switching to JSON mode/structured output based on schema presence
- Support for both manual schemas and RubyLLM::Schema objects
- Provider-specific handling (OpenAI's json_schema, Gemini's responseSchema)
- Automatic JSON parsing of responses

Key differences from the original PR:
- Used `with_schema` method name for consistency with other `with_*` methods
- Added a `force` flag to bypass model capability checks when needed
- Integrated deeply with the Rails persistence layer

Full documentation: https://rubyllm.com/guides/chat#structured-output-with-json-schemas-with_schema

The feature will be available in v1.4.0. Thanks again for the initial implementation - it was very helpful in shaping the final design\!

PR #132, opened 2025-04-22
Use Parsera LLM Specs API instead of `capabilities.rb`

crmne:
- Removed the OpenAI capabilities module and related methods.
- Simplified model parsing for OpenAI and Gemini providers, focusing on essential attributes.
- Introduced a new utility method for deep key symbolization in hashes.
- Updated OpenRouter model parsing to enhance data extraction and structure.
- Deleted obsolete browser and CLI helper tasks to streamline the codebase.
- Consolidated model updater logic into the Rake task for better maintainability.
- Improved error handling and logging throughout the model update process.


lib/ruby_llm/utils.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module RubyLLM
+  # Utility methods for the RubyLLM library.
+  module Utils
+    module_function
+
+    def symbolize_keys_deep(hash)
```
jayelkaake, line 8: This is like ActiveSupport's `deep_symbolize_keys`, which I also needed in 2 PRs [here](https://github.com/crmne/ruby_llm/pull/124/files#diff-a7a6f03a46978b8c867bf641262ba489f05e88ef500c160af627077e75496442R59-R76
) and [here](https://github.com/crmne/ruby_llm/pull/131/files#diff-a7a6f03a46978b8c867bf641262ba489f05e88ef500c160af627077e75496442R53-R70)

You'll want to cover Array child items though, so I recommend just taking my/ActiveSupport's implementation and putting it here, then I'll refactor my PRs to use your Util method instead.

lib/ruby_llm/models.rb
```diff
@@ -26,37 +26,54 @@ def models_file
...
+
+      def fetch_provider_models
+        Provider.configured_providers(RubyLLM.config).flat_map do |provider|
+          provider.list_models(connection: provider.connection(RubyLLM.config))
+                  .map { |data| ModelInfo.new(data.merge(default_model_attributes)) }
+        end
+      end
+
+      def fetch_parsera_models
+        response = Faraday.new('https://api.parsera.org/v1/llm-specs', request: { timeout: 60 })
```
jayelkaake, line 58: I don't like the idea of a core dependency on a commercial 3rd party's API, though very much appreciate the value that Parsera is doing by standardizing LLM capabilities into a JSON.

I know the code isn't called during any main runtime functions and also appreciate the support from Parsera (seems like great company).

Maybe just move this code to the rake task? Don't feel strongly about this so if you don't want to that's fine.

jayelkaake:
Love this - the initiative by Parsera to standarize capabilities of models is fantastic!

Although, I'm not confident that the LLMs are going to follow the same configs within the same capabilities so I question whether the capabilities should just be implemented in the provider classes (and if they're not implemented it means they aren't supported).

For example, structured output is supported by Gemini but it is a vastly different implementation than the other LLMs.

PR #139, opened 2025-04-25
Allow setting temperature in constructor

jayelkaake:
# Why?
Temperature isn't really "part of the conversation" so it feels much better to define it as a configuration parameter in the constructor.

# What?
* Add ability to specify temperature in the constructor like `Chat.new(temperature: 0.2).ask("Say something more consistently...")
* You can still define it with the `with_temperature` method as well if you want.

# Also
* Added IntelliJ project file and ruby version files to `.gitignore` since they shouldn't be committed by contributors based on the way this project is setup.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,8 +47,9 @@ build-iPhoneSimulator/
 # for a library or gem, you might want to ignore these files since the code is
 # intended to run in multiple environments; otherwise, check them in:
 Gemfile.lock
-# .ruby-version
-# .ruby-gemset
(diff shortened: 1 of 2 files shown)


crmne:
No, sorry.

Temperature affects the *next* message, not the whole chat session. It's a per-request parameter.

We already have `with_temperature(value)` which clearly modifies the upcoming `ask`. Adding it to the constructor creates ambiguity and a second way to do the same thing. One way is better.

PR #156, opened 2025-05-07
Attachments support for acts_as_message

bborn:
Trying again now that the Content implementation has been redone 65fdfe4

This PR adds support for passing attachments when using the acts_as Rails helpers. The main change involves serializing/deserializing the message contents to/from JSON in the database.


bborn:
@crmne  does this approach support direct-uploaded files (`direct_upload: true`)? 
- Could you share an example of how you'd expect to handle attachments in a background job? (You can't pass files as job arguments so I guess you'd need to attach the files first? Now sure how you see that working).

PR #157, opened 2025-05-07
Use foreign_key instead of to_s for acts_as methods

bborn:


lib/ruby_llm/active_record/acts_as.rb
```diff
@@ -24,19 +24,19 @@ def acts_as_chat(message_class: 'Message', tool_call_class: 'ToolCall')
                    to: :to_llm
         end
 
-        def acts_as_message(chat_class: 'Chat', tool_call_class: 'ToolCall', touch_chat: false) # rubocop:disable Metrics/MethodLength
+        def acts_as_message(chat_class: 'Chat', tool_call_class: 'ToolCall', **options) # rubocop:disable Metrics/MethodLength
```
timaro, line 27: for whatever it's worth, I didn't do this just to avoid changing the call  signature, but +1.

PR #162, opened 2025-05-08
Handle OpenAI organization and project IDs

### LINKED ISSUE #160: Handle OpenAI organization and project IDs
OpenAI’s API authentication requires organization and project IDs in certain scenarios, as outlined in their [docs].

>If you belong to multiple organizations or access projects through a legacy user API key, pass a header to specify which organization and project to use for an API request:

The configuration should add organization and project IDs to the OpenAI provider.

@crmne, thank you for your great work!

[docs]: https://platform.openai.com/docs/api-reference/authentication

xymbol:
This commit adds support for OpenAI organization and project IDs.

Closes #160

### DIFF
--- a/lib/ruby_llm/configuration.rb
+++ b/lib/ruby_llm/configuration.rb
@@ -13,6 +13,8 @@ module RubyLLM
     # Provider-specific configuration
     attr_accessor :openai_api_key,
                   :openai_api_base,
+                  :openai_organization_id,
+                  :openai_project_id,
                   :anthropic_api_key,
                   :gemini_api_key,
                   :deepseek_api_key,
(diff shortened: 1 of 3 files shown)


lib/ruby_llm/configuration.rb
```diff
@@ -56,7 +58,7 @@ def initialize
 
     def inspect
       redacted = lambda do |name, value|
-        if name.match?(/_key|_secret|_token$/)
+        if name.match?(/_id|_key|_secret|_token$/)
```
xymbol, line 61: Not sure about this, but probably a good idea.

ayestoro:
The performance optimizations here are quite clever. Nice engineering work!

PR #164, opened 2025-05-12
Don't call blob on a blob

bborn:


crmne:
Hey @bborn have you tested this?

bborn:
Yes, working for me by direct-uploading in the front end and passing the signed ids to the background job, then finding the blobs and attaching them:

```
blob = ActiveStorage::Blob.find_signed(signed_id)
final_message = chat.ask(message, with: [blob]) do |chunk|
   ...
end
```

Didn't add tests (there's nothing in the `acts_as_spec` around testing ActiveStorage stuff yet).

I'm now debugging some weird `"Invalid content type. image_url is only supported by certain models."` errors that are popping up. Seems like images aren't getting encoded correctly on subsequent LLM calls, but haven't tracked it down yet.

crmne:
Ok, I'm merging this. Indeed we're lacking tests for ActiveStorage integration. I was having some issues setting that up.

PR #165, opened 2025-05-13
Update deepseek-chat and reasoner models info

Eric-Guo:
DeepSeek official running `deepseek-chat` in v3-0324, so if `deepseek/deepseek-chat-v3-0324` provided by openrouter support those capabilities, the official DeepSeek certainly will do.

I notice Ruby-LLM will get such model info from https://api.parsera.org/v1/llm-specs but any method to update in the API?

@raznem


crmne:
We have a section about not submitting PRs for models.json in the https://github.com/crmne/ruby_llm/blob/main/CONTRIBUTING.md file. Closing.

PR #168, opened 2025-05-14
Dify API support

Eric-Guo:


tpaulshippy:
Hey Eric,

Can you help me understand what this PR is aiming at? From what I can tell, Dify is a hostable API for building AI applications. Adding it as a "provider" in RubyLLM doesn't feel quite right to me but maybe I'm missing something. I guess the closest analogy I could see would maybe be OpenRouter but that one is not hostable I don't believe and is really just a proxy to the providers (similar to Bedrock). 

From your commits, this almost looks like a ruby client to the file upload feature of Dify. But wouldn't a ruby client for Dify make more sense as a separate gem?

tpaulshippy:
On second look I do see that chat interactions are supported. Is Dify attempting to compete with OpenRouter and Bedrock? Is that how you use it?

Eric-Guo:
There is many model providers indeed today, when we as Rubyist developing AI application, I found it's pretty time consuming when optimize the prompt, evaluate different model or try some agent tooling. Those things are *not* relative with ruby_llm, even not programming.

That's the place Dify shining, As Rails developers, I using ruby_llm to developing more user friendly UI for the client. Support Dify calling will make me focus programming and leave other work to the PM/QA people. 

After model prompt, model selection or agent toolings recipes accepted by the customer. I can **eject** from dify and make minimal code change and calling model directly to make ruby program not depend on dify. (I'm not yet reach this point 🤣)

Anyway, that's the way I currently using Dify and ruby_llm.

> Is Dify attempting to compete with OpenRouter and Bedrock?

It's good question. I'm nor Dify employee or paid customer, I just using Dify as open source tools connecting with QA/PM, so my guess: Dify is not competing with those Model provider?

tpaulshippy:
Cool. I was super curious and had not heard of Dify yet. Looks interesting. Thanks!

Eric-Guo:
> 4. Incomplete code

Now dify provider support full message.

> 3. No tests

Will add after below design decision made.

> 2. Attachment hacking with UUID

Hard to avoid as the Dify API requires the chat with file to using upload_document API to upload to Dify first, then using the return UUID as local_file representative.

> 1. API pollution: Adding upload_document to Chat only works for Dify. 

Just move `upload_document` to provider and will call `chat.provider.upload_document` now.

The key argue here is if we need to support Dify as provider and if it is yes, then whether Dify upload document API is application logic or should be part of RubyLLM API provider.

Current RubyLLM design is great and no provider-specific features, I'm fully appreciate and that's why I using RubyLLM and hoping continue using it after `eject` from Dify.

So I'm quite OK this PR cannot merge if RubyLLM designer no plan to provide any provider-specific features in future.

Hoping this PR is still usefully for other Dify & Ruby user.

crmne:
@Eric-Guo there are other providers that have upload functionality, but we need to get that right first instead of frankesteining uploading documents for Dify first and then breaking the API later. Are you willing to follow my requests for changes _and then_ helping us make file uploads work?

Eric-Guo:
@crmne You means handles attachments in ask method? this use case is different with Dify, Dify upload files first and get uuid, then using uuid to represent the file instead of when `ask` method, upload the image file as the part of the actually one API call.

```ruby
    it 'handles attachments in ask method' do # rubocop:disable RSpec/ExampleLength,RSpec/MultipleExpectations
      chat = Chat.create!(model_id: model)

      image_upload = uploaded_file(image_path, 'image/png')

      response = chat.ask('What do you see?', with: image_upload)

      user_message = chat.messages.find_by(role: 'user')
      expect(user_message.attachments.count).to eq(1)
      expect(response.content).to be_present
    end
```

So my understanding is you means combining the Dify upload function API call with RubyLLM#ask method?

It's works, but quite bad idea, at least for my case, I'm currently doing a contract review project using Ruby LLM:

image


There is 40 question which I build as 40 apps in Dify, with different prompt template, so current my calling API is 40 + 1 and no duplicate files upload, if combine, then it will become 40*2 Dify api call.

crmne:
@Eric-Guo, you're solving the wrong problem at the wrong level.

**The provider abstraction exists for a reason.** RubyLLM provides a clean, unified interface across LLM providers. The moment we start adding provider-specific methods like `upload_document`, we've broken that abstraction and turned the library into a kitchen sink.

Your use case - uploading 40 documents once and referencing them by UUID across multiple chat sessions - is **application logic**, not LLM provider logic. That's exactly the kind of workflow optimization that belongs in your application layer, not in a library that's trying to provide a consistent interface.

**Here's what you should do:**

1. Strip this PR down to **just the basic chat provider**. Remove all upload functionality.
2. Handle file uploads in your application code. Upload to Dify directly, store the UUIDs, then pass them in chat messages.
3. If you need file upload abstractions later, we'll design them properly across ALL providers that support uploads - not hack them in for one provider.

You're trying to optimize for your specific 40-document workflow, but that's premature optimization that breaks the library's design principles. Build the foundation first, optimize your specific use case in your app layer.

**This is basic library design:** Keep the core abstraction clean. Provider-specific workarounds belong in your application, not in the shared interface.

Ship the basic Dify chat provider. Handle your file uploads in your app. When we're ready to tackle file uploads properly across providers, we'll do it right.

Eric-Guo:
> This is basic library design: Keep the core abstraction clean. Provider-specific workarounds belong in your application, not in the shared interface.

tiny correct, it's acturally  not on the shared interface anymore.

```ruby
response = chat.provider.upload_document('/Users/guochunzhong/git/my_contract.docx')
```

It's on the RubyLLM specfic provider level. I known RubyLLM want to hide the provider concept to the user and it's great, but it's also hard as the [first installation step](https://rubyllm.com/#installation) already mention the provider.

> you're solving the wrong problem at the wrong level.

RubyLLM is first, the *AI API wrapper library*, then the common library interface for all AI interface, I'm not persuading you merge this PR, I would not merge this PR either if I were maintainer.

But as AI API wrapper library, RubyLLM should like the ActiveRecord, unify all DB interfaces, but also leave [some postgresql feature possible](https://guides.rubyonrails.org/active_record_postgresql.html), I'm not say `chat.provider` is good design here, but it's at least progamtic and far from wrong.

I suggest just leave this PR as open and draft status and I will continue add more provider which won't introduce upload file dilemma problems anymore, like Kimi K2 maybe?

crmne:
You're still missing the point, Eric.

**`chat.provider.upload_document()` is still breaking the abstraction.** You're exposing provider internals and encouraging users to write provider-specific code. That's exactly what we don't want.

Your ActiveRecord comparison is wrong. PostgreSQL features in Rails are still abstracted through ActiveRecord's unified interface - they don't expose raw `connection.postgresql_specific_method()`. We provide `add_index using: :gin` not `connection.create_gin_index()`.

**Here's what happens with your approach:**
```ruby
# User code becomes provider-dependent
if chat.provider.is_a?(RubyLLM::Providers::Dify)
  uuid = chat.provider.upload_document(file)
  response = chat.ask("Review this", with: uuid)
else
  # What now? Different code path for every provider?
end
```

That's exactly the vendor lock-in we're trying to avoid.

**I'm closing this PR.** Not because I don't appreciate the work, but because it sets a bad precedent. Every provider maintainer will want to add their special methods, and we'll end up with a fragmented mess.

Fork it, keep your branch, use it for your projects. When we design file uploads properly - with a unified interface that works across providers - we'll do it right.

The library's job is to hide provider differences, not expose them through escape hatches.

**Closed.**

Eric-Guo:
Sorry I have to release a gem with different name `dify_llm`, [reason 1](https://github.com/langgenius/ruby-sdk/pull/6#issuecomment-3270480721) and reason 2 is due to recent [introduction of lfs](b167114a0742a4b5b6419850220aac6503d46d92) using from source repo will lead bundler install failing.

```diff
-gem "ruby_llm", git: "https://git.thape.com.cn/ai/ruby_llm.git", branch: :dify_api
+gem "dify_llm", require: "ruby_llm"
```

I'm open to change dify_api branch if ruby_llm decide support upload new file in separate API, as far as I know [anthropic file api](https://docs.anthropic.com/en/api/files-create) is in Beta.

PR #171, opened 2025-05-14
docs: add guide for using custom model names with acts_as helpers

matheuscumpian:
While integrating RubyLLM with an existing Rails application that already had a `Chat` model, I discovered that the framework supports using custom model names with the `acts_as` helpers. This isn't explicitly documented, but is a valuable feature for real-world Rails applications where naming conflicts or existing conventions might exist.

Let me know if I'm missing something on this behaviour


matheuscumpian:
closes #183

docs/guides/rails.md
```diff
@@ -474,6 +474,46 @@ This setup allows for:
 
 Your `Chat`, `Message`, and `ToolCall` models are standard ActiveRecord models. You can add any other associations, validations, scopes, callbacks, or methods as needed for your application logic. The `acts_as` helpers provide the core persistence bridge to RubyLLM without interfering with other model behavior.
 
+{: .note }
+You can use custom model names by passing parameters to the `acts_as` helpers. For example, if you prefer `Conversation` over `Chat`, you could use `acts_as_chat` in your `Conversation` model and then specify `chat_class: 'Conversation'` in your `Message` model's `acts_as_message` call.
```
crmne, line 499: No need for a note when the content is right below.

docs/guides/rails.md
```diff
@@ -474,6 +474,46 @@ This setup allows for:
...
+
+#### Using Custom Model Names
```
crmne: Why `<h4>`? it's right after an `h2` so I'd say `h3`

crmne:
@matheuscumpian want to resolve the requested changes?

matheuscumpian:
@crmne thanks for the review and sorry for the lateness, I've just addressed the comments on a new commit

PR #187, opened 2025-05-18
a little refactor

max-power:


lib/ruby_llm/chat.rb
```diff
@@ -106,7 +106,7 @@ def complete(&)
     end
 
     def add_message(message_or_attributes)
-      message = message_or_attributes.is_a?(Message) ? message_or_attributes : Message.new(message_or_attributes)
+      message = Message.wrap(message_or_attributes)
```
crmne, line 109: Not sure what's the value of this

lib/ruby_llm/models.rb
crmne, line 1: This has already been changed this way.

crmne (changes requested):

PR #189, opened 2025-05-20
feat(Configuration): Add http_proxy configuration option 

stadia:
implement proxy setup in Connection


crmne (changes requested):
Hey, thank you for this! Could you simply add it to the configuration guide too?

crmne (approved):

PR #194, opened 2025-05-20
Added ENV variable to silence capabilities warning

### LINKED ISSUE #191: Disable the `Capabilities may not be accurately reflected` warning.
Can we disable the `Capabilities may not be accurately reflected` warning? It's incredibly annoying for anyone running models locally with LMStudio or ollama.

```
W, [2025-05-20T12:59:54.809886 #96685]  WARN -- RubyLLM: Assuming model 'qwen3-30b-a3b-mlx' exists for provider 'RubyLLM::Providers::OpenAI'. Capabilities may not be accurately reflected.
```

I would even settle for an ENV variable to turn it off.

nickpoorman:
Fixes #191

### DIFF
--- a/lib/ruby_llm/models.rb
+++ b/lib/ruby_llm/models.rb
@@ -61,8 +61,10 @@ module RubyLLM
             modalities: { input: %w[text image], output: %w[text] },
             metadata: { warning: 'Assuming model exists, capabilities may not be accurate' }
           )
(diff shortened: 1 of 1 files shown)


crmne (changes requested):
Thank you for your contribution.

Could you make this into an option in configuration instead of an env var?

Make sure to update the `configuration.md` as well.

PR #197, opened 2025-05-21
Message role shortcuts and role inquiry methods

max-power:
define class methods for each Role ex. Message.user(content) and role inquiry methods ex. message.assistent?


crmne (commented):
This is nice syntactic sugar, but what purpose does it serve beyond that?

Issue #203, opened 2025-05-22
Claude 3.7 Sonnet alias is gone


tpaulshippy:
After aliases started getting automatically generated, we lost this:

```
{
  "claude-3-7-sonnet": {
    "anthropic": "claude-3-7-sonnet-20250219",
    "bedrock": "us.anthropic.claude-3-7-sonnet-20250219-v1:0",
    "openrouter": "us.anthropic.claude-3-7-sonnet-20250219-v1:0"
  },
```

Commit: https://github.com/crmne/ruby_llm/commit/120526c38a53f29408be2f59453a9b9cc3298548

crmne:
@tpaulshippy is it solved by the PR?

crmne:
It looks like it. Closing.

PR #206, opened 2025-05-23
Issue 203: Fix model alias matching for provider prefixes (claude-3-7-sonnet support)

Sami-Tanquary:
## Resolves issue #203

Fix model alias matching to properly handle provider prefixes, particularly for Bedrock models that can have varying prefixes (e.g., `anthropic.` or `us.anthropic.`).

## Changes
- Updated alias matching logic to handle provider prefixes more flexibly
- Fixed support for `claude-3-7-sonnet` model which uses the `us.anthropic.` prefix
- Ran aliases.rake to add proper model aliases back for Claude-3-7-sonnet

## Problem
The alias matching system was too strict with provider prefixes, causing models like `claude-3-7-sonnet` to fail matching when the Bedrock version used a different prefix (`us.anthropic.`) compared to other Claude models (`anthropic.`).

## Solution
Modified the alias matching to be more flexible with provider prefixes while maintaining the correct model identification. This ensures compatibility with both existing and new model versions regardless of their provider prefix format.

## Testing
- Verified that `claude-3-7-sonnet` model works correctly with both Anthropic API and Bedrock
- Added spec for handling different provider prefixes correctly for bedrock models
- Ensured existing model aliases continue to work as expected

### DIFF
--- a/lib/ruby_llm/aliases.json
+++ b/lib/ruby_llm/aliases.json
@@ -29,6 +29,14 @@
     "anthropic": "claude-3-5-sonnet-20241022",
     "bedrock": "anthropic.claude-3-5-sonnet-20240620-v1:0:200k"
   },
+  "claude-3-7-sonnet": {
+    "anthropic": "claude-3-7-sonnet-latest",
+    "bedrock": "us.anthropic.claude-3-7-sonnet-20250219-v1:0"
+  },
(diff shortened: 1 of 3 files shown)


tpaulshippy:
We hit this bug this week and it prevented a new feature from getting demonstrated. We worked around it with our own alias, but happy to see it get fixed in the gem.

PR #210, opened 2025-05-24
Truncate tool function names to 64 characters, which is the OpenAI limit

mattparlane:
I have some very long tool names in my codebase, and I ran into the OpenAI 64-character limit.

My fix is pretty blunt, and it's system-wide rather than provider-specific.  Obviously happy if you want to re-work it, just opening the discussion really.

I ran into an error about an invalid API key when running this against `anthropic/claude-3-5-haiku-20241022` -- not sure what that's about, it passes on all other providers.

Thanks!

### DIFF
--- a/lib/ruby_llm/tool.rb
+++ b/lib/ruby_llm/tool.rb
@@ -57,6 +57,7 @@ module RubyLLM
           .gsub(/([a-z\d])([A-Z])/, '\1_\2')
           .downcase
           .delete_suffix('_tool')
+          .slice(0, 64)
     end
 
     def description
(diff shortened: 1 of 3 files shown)


nofxx:
About that, and as user can ask about the tools, ai may reply like

Available tools: Get strange things (via `may_look_not--look_so_frindly`)

What if an option to put the name you want?

```
  class StrangeAndLongName < RubyLLM::Tool
    description "Get a strange thing"
    name "MoreFriendly"
```

can it use spaces?

crmne:
Why can't you shorten the tool names? Genuine question.

nofxx:
Hey, yeah... moved my tools out of namespaces, but I actually aiming at I18n, sometimes the model spits out the english tool names in a portuguese app heh...

crmne:
It sounds like a problem you can solve in app code, which goes against the [contributing](https://github.com/crmne/ruby_llm/blob/main/CONTRIBUTING.md) guide. Closing

PR #218, opened 2025-06-02
Fix Encoding::CompatibilityError: Unicode Normalization not appropriate for ASCII-8BIT

Eric-Guo:
Encoding::CompatibilityError: Unicode Normalization not appropriate for ASCII-8BIT is introduced after #215, commit: d34592efe00697457ac020cd5ec80a13c2827fed

The fix is copy from:
https://github.com/knu/ruby-domain_name/pull/27


tpaulshippy:
Should we have a spec?

crmne:
A spec is a great idea!

crmne:
@Eric-Guo I looked at the spec. When is this problem going to happen?

Eric-Guo:
image

It's happen in my cursor ruby notebook, notice I running `rake install:local` to get main version.

lib/ruby_llm/tool.rb
```diff
@@ -49,14 +49,14 @@ def parameters
...
-          .unicode_normalize(:nfkd)
-          .encode('ASCII', replace: '')
-          .gsub(/[^a-zA-Z0-9_-]/, '-')
-          .gsub(/([A-Z]+)([A-Z][a-z])/, '\1_\2')
-          .gsub(/([a-z\d])([A-Z])/, '\1_\2')
-          .downcase
-          .delete_suffix('_tool')
+      klass_name = self.class.name
+      normalized = klass_name.ascii_only? ? klass_name : klass_name.unicode_normalize(:nfkd)
+      normalized.encode('ASCII', replace: '')
```
crmne: This assumes the encoding is going to be either ASCII or Unicode. Is there a way to convert any encoding to unicode instead?

crmne (changes requested):
Thank you for this contribution! I can see the value in it.

crmne:
Also make sure the tests pass before pushing. `overcommit --install`

PR #220, opened 2025-06-03
Feat: Add caching for tools

JulianPasquale:
Hey @crmne, love the work you're doing with the gem!

This is just an idea/proposal, and I completely understand if this is something you don't want to add as part of the gem, or you think it would be better as a separate plugin or gem. It's not completely ready to merge yet, but wanted to share it before spending more time on it.

I've been playing around with the gem, and found a couple scenarios where caching the output of a tool call would really help saving some costs and also making things faster, for instance if you have a tool like this to get the content of a webpage with Tavily:

```ruby
class TavilyExtract < RubyLLM::Tool
  description "Extract web page content from a specified URL"
  param :url, desc: "The URL of the page to extract the content from"

  def initialize
    @api_key = ENV.fetch("TAVILY_API_KEY", nil)
  end

  def execute(url:)
    uri = "https://api.tavily.com/extract"
    headers = {
      "Authorization" => "Bearer #{api_key}",
      "Content-Type" => "application/json"
    }
    body = {
      urls: url,
      extract_depth: 'advanced'
    }.to_json

    JSON.parse(HTTParty.post(uri, headers: headers, body: body).body)
  end
end
```

Of course the content of the page can change, but maybe that's not super critical for you and having a cached version of it is good enough, or you can simply put a short expiration time for it.

Not a huge fan of depending on `activestorage` for this, but just wanted to provide a generic mechanism to support multiple caching adapters. Any other idea is welcome.

### DIFF
--- a/gemfiles/rails_7.1.gemfile.lock
+++ b/gemfiles/rails_7.1.gemfile.lock
@@ -2,6 +2,7 @@ PATH
   remote: ..
   specs:
     ruby_llm (1.3.0rc1)
+      activesupport
       base64
       event_stream_parser (~> 1)
       faraday (>= 1.10.0)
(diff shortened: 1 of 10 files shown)


crmne:
Hi @JulianPasquale, thank you for the kind words, but caching tool outputs is out of scope for RubyLLM: it's a very specific use case that's easy to implement on top of it.

PR #225, opened 2025-06-04
Add support for custom logger via configuration

### LINKED ISSUE #223: Add support for a custom logger.
Currently RubyLLM only allows setting the log file and level, so you can't assign a custom or pre-existing logger. For example, in a Rails application it'd be nice to be able to assign `Rails.logger`. I was about to open a pull request to add the ability to do so, but I realized there's two options for how it's introduced and I'm curious which would be preferred.

### Configuration Accessor

A `Configuration#logger` accessor could be added to the configuration class.

```ruby
RubyLLM.configure do |config|
  config.logger = Rails.logger
end
```

I think this makes the most sense personally, but it's also a bit confusing given the existing ability to customize the file and level already. I was planning to support both by doing the following:

```ruby
def logger
  @logger ||= config.logger || Logger.new(
    config.log_file,
    progname: 'RubyLLM',
    level: config.log_level
  )
end
```

### Class Attribute Writer

A `RubyLLM.logger=` writer could be add to the top-level module. This is what I started with until I considered the configuration options.

```ruby
RubyLLM.logger = Rails.logger
```

Let me know what you think between the two options, or if you have another idea. Either way happy to implement and submit a PR.

tristandunn:
Fixes #223.

### DIFF
--- a/lib/ruby_llm.rb
+++ b/lib/ruby_llm.rb
@@ -67,7 +67,7 @@ module RubyLLM
     end
 
     def logger
-      @logger ||= Logger.new(
+      @logger ||= config.logger || Logger.new(
         config.log_file,
         progname: 'RubyLLM',
         level: config.log_level
--- a/lib/ruby_llm/configuration.rb
+++ b/lib/ruby_llm/configuration.rb
@@ -36,6 +36,7 @@ module RubyLLM
                   :retry_interval_randomness,
                   :http_proxy,
                   # Logging configuration
+                  :logger,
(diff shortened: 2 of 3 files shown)


crmne (changes requested):
Thanks @tristandunn for this contribution.

The code LGTM, I'd just love to have a documentation change too in which we also warn the user that `log_file` and `log_level` are not used when a custom `logger` is specified.

crmne (approved):
LGTM!

PR #234, opened 2025-06-09
Prompt caching

### LINKED ISSUE #13: Support prompt caching
As an effort to reduce rate limits and increase token efficiency.

Anthropic just released an update on this: https://www.anthropic.com/news/token-saving-updates

tpaulshippy:
## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

- [ ] I read the [Contributing Guide](https://github.com/crmne/ruby_llm/blob/main/CONTRIBUTING.md)
- [ ] This isn't application-specific logic that belongs in user code
- [ ] This benefits most users, not just my specific use case

## Quality check

## API changes

- [ ] Breaking change
- [ ] No API changes

## Related issues

Resolves #13

### DIFF
--- a/lib/ruby_llm/chat.rb
+++ b/lib/ruby_llm/chat.rb
@@ -25,6 +25,7 @@ module RubyLLM
       @temperature = 0.7
       @messages = []
       @tools = {}
+      @cache_prompts = { system: false, user: false, tools: false }
       @on = {
         new_message: nil,
         end_message: nil
@@ -92,6 +93,11 @@ module RubyLLM
       messages.each(&)
     end
 
+    def cache_prompts(system: false, user: false, tools: false)
+      @cache_prompts = { system: system, user: user, tools: tools }
+      self
+    end
+
     def complete(&)
       @on[:new_message]&.call
       response = @provider.complete(
@@ -99,6 +105,7 @@ module RubyLLM
         tools: @tools,
         temperature: @temperature,
         model: @model.id,
+        cache_prompts: @cache_prompts.dup,
         connection: @connection,
         &
       )
--- a/lib/ruby_llm/provider.rb
+++ b/lib/ruby_llm/provider.rb
@@ -10,7 +10,8 @@ module RubyLLM
     module Methods
       extend Streaming
 
-      def complete(messages, tools:, temperature:, model:, connection:, cache_prompts: { system: false, user: false, tools: false }, &)
+      def complete(messages, tools:, temperature:, model:, connection:,
+                   cache_prompts: { system: false, user: false, tools: false }, &)
         normalized_temperature = maybe_normalize_temperature(temperature, model)
 
         payload = render_payload(messages,
--- a/lib/ruby_llm/provider.rb
+++ b/lib/ruby_llm/provider.rb
@@ -10,13 +10,14 @@ module RubyLLM
     module Methods
       extend Streaming
 
-      def complete(messages, tools:, temperature:, model:, connection:, &)
+      def complete(messages, tools:, temperature:, model:, connection:, cache_prompts: { system: false, user: false, tools: false }, &)
         normalized_temperature = maybe_normalize_temperature(temperature, model)
 
         payload = render_payload(messages,
                                  tools: tools,
                                  temperature: normalized_temperature,
                                  model: model,
+                                 cache_prompts: cache_prompts,
                                  stream: block_given?)
 
         if block_given?
--- a/lib/ruby_llm/providers/anthropic/media.rb
+++ b/lib/ruby_llm/providers/anthropic/media.rb
@@ -7,11 +7,11 @@ module RubyLLM
       module Media
         module_function
 
-        def format_content(content)
-          return [format_text(content)] unless content.is_a?(Content)
+        def format_content(content, cache: false)
+          return [format_text(content, cache:)] unless content.is_a?(Content)
 
           parts = []
-          parts << format_text(content.text) if content.text
+          parts << format_text(content.text, cache:) if content.text
 
           content.attachments.each do |attachment|
             case attachment.type
@@ -29,60 +29,84 @@ module RubyLLM
           parts
         end
 
-        def format_text(text)
-          {
-            type: 'text',
-            text: text
-          }
+        def format_text(text, cache: false)
+          with_cache_control(
+            {
+              type: 'text',
+              text: text
+            },
+            cache:
+          )
         end
 
-        def format_image(image)
+        def format_image(image, cache: false)
           if image.url?
-            {
-              type: 'image',
-              source: {
-                type: 'url',
-                url: image.source
-              }
-            }
+            with_cache_control(
+              {
+                type: 'image',
+                source: {
+                  type: 'url',
+                  url: image.source
+                }
+              },
+              cache:
+            )
           else
-            {
-              type: 'image',
-              source: {
-                type: 'base64',
-                media_type: image.mime_type,
-                data: image.encoded
-              }
-            }
+            with_cache_control(
+              {
+                type: 'image',
+                source: {
+                  type: 'base64',
+                  media_type: image.mime_type,
+                  data: image.encoded
+                }
+              },
+              cache:
+            )
           end
         end
 
-        def format_pdf(pdf)
+        def format_pdf(pdf, cache: false)
           if pdf.url?
-            {
-              type: 'document',
-              source: {
-                type: 'url',
-                url: pdf.source
-              }
-            }
+            with_cache_control(
+              {
+                type: 'document',
+                source: {
+                  type: 'url',
+                  url: pdf.source
+                }
+              },
+              cache:
+            )
           else
-            {
-              type: 'document',
-              source: {
-                type: 'base64',
-                media_type: pdf.mime_type,
-                data: pdf.encoded
-              }
-            }
+            with_cache_control(
+              {
+                type: 'document',
+                source: {
+                  type: 'base64',
+                  media_type: pdf.mime_type,
+                  data: pdf.encoded
+                }
+              },
+              cache:
+            )
           end
         end
 
-        def format_text_file(text_file)
-          {
-            type: 'text',
-            text: Utils.format_text_file_for_llm(text_file)
-          }
+        def format_text_file(text_file, cache: false)
+          with_cache_control(
+            {
+              type: 'text',
+              text: Utils.format_text_file_for_llm(text_file)
+            },
+            cache:
+          )
+        end
+        module_function
+
+        def with_cache_control(hash, cache: false)
+          return hash unless cache
+          hash.merge(cache_control: { type: 'ephemeral' })
         end
       end
     end
--- a/lib/ruby_llm/providers/anthropic/media.rb
+++ b/lib/ruby_llm/providers/anthropic/media.rb
@@ -102,10 +102,10 @@ module RubyLLM
             cache:
           )
         end
-        module_function
 
         def with_cache_control(hash, cache: false)
           return hash unless cache
+
           hash.merge(cache_control: { type: 'ephemeral' })
         end
       end
--- a/lib/ruby_llm/providers/anthropic/chat.rb
+++ b/lib/ruby_llm/providers/anthropic/chat.rb
@@ -11,12 +11,14 @@ module RubyLLM
           '/v1/messages'
         end
 
-        def render_payload(messages, tools:, temperature:, model:, stream: false, cache_prompts: { system: false, user: false, tools: false })
+        def render_payload(messages, tools:, temperature:, model:, stream: false,
+                           cache_prompts: { system: false, user: false, tools: false })
           system_messages, chat_messages = separate_messages(messages)
           system_content = build_system_content(system_messages, cache: cache_prompts[:system])
 
           build_base_payload(chat_messages, temperature, model, stream, cache: cache_prompts[:user]).tap do |payload|
-            add_optional_fields(payload, system_content: system_content, tools: tools, cache_tools: cache_prompts[:tools])
+            add_optional_fields(payload, system_content: system_content, tools: tools,
+                                         cache_tools: cache_prompts[:tools])
           end
         end
 
@@ -26,14 +28,14 @@ module RubyLLM
 
         def build_system_content(system_messages, cache: false)
           system_messages.flat_map.with_index do |msg, idx|
-            cache = idx == system_messages.size - 1 ? cache : false
+            cache = false unless idx == system_messages.size - 1
             format_system_message(msg, cache:)
           end
         end
 
         def build_base_payload(chat_messages, temperature, model, stream, cache: false)
           messages = chat_messages.map.with_index do |msg, idx|
-            cache = idx == chat_messages.size - 1 ? cache : false
+            cache = false unless idx == chat_messages.size - 1
             format_message(msg, cache:)
           end
 
--- a/lib/ruby_llm/providers/anthropic/chat.rb
+++ b/lib/ruby_llm/providers/anthropic/chat.rb
@@ -11,12 +11,12 @@ module RubyLLM
           '/v1/messages'
         end
 
-        def render_payload(messages, tools:, temperature:, model:, stream: false)
+        def render_payload(messages, tools:, temperature:, model:, stream: false, cache_prompts: { system: false, user: false, tools: false })
           system_messages, chat_messages = separate_messages(messages)
-          system_content = build_system_content(system_messages)
+          system_content = build_system_content(system_messages, cache: cache_prompts[:system])
 
-          build_base_payload(chat_messages, temperature, model, stream).tap do |payload|
-            add_optional_fields(payload, system_content:, tools:)
+          build_base_payload(chat_messages, temperature, model, stream, cache: cache_prompts[:user]).tap do |payload|
+            add_optional_fields(payload, system_content: system_content, tools: tools, cache_tools: cache_prompts[:tools])
           end
         end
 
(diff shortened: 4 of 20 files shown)


tpaulshippy:
@crmne As I don't have an Anthropic key, I'll need you to generate the VCR cartridges for that provider. Hoping everything just works, but let me know if not.

crmne:
@tpaulshippy this would be great to have! Will you be willing to enable it on all providers?

I'll do a proper review when I can.

tpaulshippy:
My five minutes of research indicates that at least OpenAI and Gemini take the approach of automatically caching for you based on the size and structure of your request. So the only support I think we'd really need for those two is to populate the cached token counts on the response messages. Unless we want to try to support explicit caching on the Gemini API but that looks complex and not as commonly needed.

Do you know of other providers that require payload changes for prompt caching?

lib/ruby_llm/providers/anthropic/media.rb
```diff
@@ -29,60 +29,84 @@ def format_content(content)
...
+              text: Utils.format_text_file_for_llm(text_file)
+            },
+            cache:
+          )
+        end
+
+        def with_cache_control(hash, cache: false)
+          return hash unless cache
+
+          hash.merge(cache_control: { type: 'ephemeral' })
```
tpaulshippy, line 110: Realizing this might cause errors on older models that do not support caching. If it does, we could raise here, or just let the API validation handle it. I'm torn on whether the capabilities check complexity is worth it as these models are probably so rarely used.

tpaulshippy:
> @crmne As I don't have an Anthropic key, I'll need you to generate the VCR cartridges for that provider. Hoping everything just works, but let me know if not.

Scratch that. I decided to stop being a cheapskate and just pay Anthropic their $5.

tpaulshippy:
Looking to implement this in our project and now I'm wondering if it should be an opt out rather than an opt in. If you are using unique prompts every time I guess it adds some cost to cache them but my guess is in most applications prompts will get repeated, especially system prompts.

docs/guides/prompt-caching.md
crmne, line 1: I don't feel like it needs its own guide. It's a small feature that can be added in the chat guide.

lib/ruby_llm/chat.rb
```diff
@@ -92,16 +95,22 @@ def each(&)
...
+        messages: messages,
         tools: @tools,
         temperature: @temperature,
         model: @model.id,
-        connection: @connection,
-        &
+        cache_prompts: @cache_prompts.dup,
+        connection: @connection
       )
+      response = @provider.complete(params, &)
```
crmne: I don't feel like it's necessary to make a new `CompletionParams` class. Let's _remove_ complexity.

lib/ruby_llm/completion_params.rb
crmne, line 1: I don't feel like it's necessary to make a new `CompletionParams` class. Let's _remove_ complexity.

lib/ruby_llm/provider.rb
crmne, line 1: I don't feel like it's necessary to make a new `CompletionParams` class. Let's _remove_ complexity.

spec/fixtures/large_prompt.txt
crmne, line 1: why do you need this? you can prompt cache a small prompt too.

crmne (changes requested):
Thank you for this feature @tpaulshippy, however there are several improvements I'd like you to make before we merge this.

On top of the ones made in the comments, and the most important one, I'd like to have prompt caching implemented in all providers.

Plus I have not fully checked the logic in `providers/anthropic` but the patch seems a bit heavy-handed with the amount of changes needed at first glance. Where all changes necessary or could it be done in a simpler manner?

lib/ruby_llm/chat.rb
tpaulshippy: This was necessary only because rubocop complained about the number of parameters passed. Should I just ignore rubocop on that?

tpaulshippy:
> I'd like to have prompt caching implemented in all providers.

Did you see [this](https://github.com/crmne/ruby_llm/pull/234#issuecomment-2962928284)? Is the request to populate the cached token counts on the response messages for OpenAI and Gemini?

spec/fixtures/large_prompt.txt
tpaulshippy: Really?

[This](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#cache-limitations) would indicate otherwise.

"Shorter prompts cannot be cached"

crmne: My bad, I should have checked the docs before writing! In that case, you can still generate a (nonsensical) prompt like I do for the big message test. [massive_text = 'a' * 1_000_000](https://github.com/crmne/ruby_llm/blob/78482e5763a553956132769e23279b6dd99da8f8/spec/ruby_llm/chat_error_spec.rb#L73). Just make it the right size according to the docs.

lib/ruby_llm/chat.rb
crmne: @tpaulshippy yes I'd say so, just add a disable comment.

crmne:
> Did you see [this](https://github.com/crmne/ruby_llm/pull/234#issuecomment-2962928284)? Is the request to populate the cached token counts on the response messages for OpenAI and Gemini?

Thank you for pointing that out, I had missed it. I think it would certainly be a nice addition to RubyLLM to have all providers have almost the same level of support of caching.

tpaulshippy:
> > Did you see [this](https://github.com/crmne/ruby_llm/pull/234#issuecomment-2962928284)? Is the request to populate the cached token counts on the response messages for OpenAI and Gemini?
> 
> Thank you for pointing that out, I had missed it. I think it would certainly be a nice addition to RubyLLM to have all providers have almost the same level of support of caching.

Ok we have a bit of a naming issue. Here's the property names we get from each provider:

Anthropic
`cache_creation_input_tokens`
`cache_read_input_tokens`

OpenAI
`cached_tokens`

Gemini
`cached_content_token_count`

My reading of the docs indicates that the OpenAI and Gemini values correspond pretty closely with the `cache_read_input_tokens` in Anthropic. 

What should we call these properties in the Message?

crmne:
For the naming, let's go with:
- `cached_tokens` - maps to the cache read values from all providers (the main property developers will use)
- `cache_creation_tokens` - Anthropic-specific cache creation cost (nil for other providers)

This keeps it consistent with our existing `input_tokens`/`output_tokens` pattern while handling the provider differences cleanly.

Can you update the Message properties to use these names? Thanks Paul!

tpaulshippy:
@crmne Ok I addressed all the feedback. Only issue I'm running into is that I'm unable to get openai and gemini to actually report any cached tokens. I may need help from someone who knows those APIs better than me. The code to report their cached token counts is in place but I commented out the tests since I could not get them to work.

[Gemini docs on this](https://ai.google.dev/gemini-api/docs/caching?lang=python#implicit-caching)
[OpenAI docs on this](https://platform.openai.com/docs/guides/prompt-caching)

smerickson:
@tpaulshippy thanks so much for working on this!  I'd love to be able to switch to RubyLLM for my AI library needs and this was one of the features I was still waiting on.  Really appreciate you leading the charge.

tpaulshippy:
> @crmne Ok I addressed all the feedback. Only issue I'm running into is that I'm unable to get openai and gemini to actually report any cached tokens. I may need help from someone who knows those APIs better than me. The code to report their cached token counts is in place but I commented out the tests since I could not get them to work.

I realized that I wasn't actually calling the API twice, which would be necessary to hit the cache. But even after trying that I still couldn't get gemini/openai to report cache hits.

tpaulshippy:
Had to play with the length/complexity of the prompts. Finally got them to cache. Should be ready to go.

jordanyeo:
This will be huge for me- possible to get a review and merge? Thanks!

tpaulshippy:
I rolled this out 2 weeks ago. So far the numbers show that I am paying 55% less than I would have without this.

crmne (changes requested):
Paul, this PR is 90% VCR cassettes, but it's still a monster and I feel like the actual code is overengineered. Passing `cache_control` through 7 layers? No thanks.

However, how about _embracing_ the fact that there may be moments we want to augment each message?

```ruby
# Just add message-level params
chat.add_message(
  role: :user,
  content: "huge doc",
  params: { cache_control: { type: 'ephemeral' } }
)
```

That's it. Add `params` to Message.

We can then even add:

```ruby
chat.with_message_params(cache_control: { type: 'ephemeral' })
```

In order to apply it to every message. That's it.

Then adding a provider agnostic implementation of cache control becomes trivial.

tpaulshippy:
That could work for system and user messages. How about tools? 

I still think there will be some complexity in the anthropic provider that may be unavoidable.

tpaulshippy:
One complication is that these params have to be nested under the content array for anthropic.

So it would actually need to be something like this:

```
chat.add_message(
  role: :user,
  content: "huge doc",
  params: { content: [{ cache_control: { type: 'ephemeral' } }] }
)
```

Right?

tpaulshippy:
I don't think adding the params to all messages with chat.with_message_params will generally work as you suggest.

Couple of reasons - 
1. Anthropic limits you to 4 cache breakpoints per request.
2. You generally want your cache breakpoints on the last system or the last user message to get the maximum use of the cache.

tpaulshippy:
I do like the suggestion of adding params to messages because that could give more fine grained control over exactly where you want the breakpoints. Would also enable passing optional keys on other providers like this one from OpenAI --
image
EDIT: Not seeing this option on the OpenAI responses API...

docs/_core_features/chat.md
```diff
@@ -500,6 +500,30 @@ puts "Total Conversation Tokens: #{total_conversation_tokens}"
...
+### Enabling
+For Anthropic models, you can opt-in to prompt caching which is documented more fully in the [Anthropic API docs](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching).
+
+Enable prompt caching using the `cache_prompts` method on your chat instance:
+
+```ruby
+chat = RubyLLM.chat(model: 'claude-3-5-haiku-20241022')
+
+# Enable caching for different types of content
+chat.cache_prompts(
```
tpaulshippy: The more I think about it, the more I'm thinking it is odd to have a provider specific feature enabled like this. I wish anthropic was just more like the others in this area.

tpaulshippy (commented):
Would like to explore making this a per message option and maybe something like a with_tools_params option to make this more provider agnostic.

tpaulshippy:
> One complication is that these params have to be nested under the content array for anthropic.
> 
> So it would actually need to be something like this:
> 
> ```
> chat.add_message(
>   role: :user,
>   content: "huge doc",
>   params: { content: [{ cache_control: { type: 'ephemeral' } }] }
> )
> ```
> 

Looks like to make this work, deep_merge will have to be enhanced to support arrays. Currently it just overrides the whole array rather than merging the hashes within. Will need something like this:
image
What do you think @crmne ?

tpaulshippy:
I'm honestly a bit torn on the params approach. My concern is that it will require consumers to deal with a lot of messy hashes and avoiding that is one of the main reasons I switched to this library. On the other hand this particular feature (opting in to caching with per message breakpoints) is very Anthropic specific so following the pattern set with other provider-specific features it should be enabled via a params hash.

tpaulshippy:
I decided to try a different approach. Since the opting into caching is a provider specific feature, I decided to introduce `with_provider_options` where you can pass either named parameters or an instance of an options class.

I don't love the `params: { cache_control: { type: 'ephemeral' } }` approach for a couple of reasons:
1. The hash looks messy and feels like I'm going back to langchainrb
2. It requires people to understand the intricacies of the provider payload. Avoiding this is one of the reasons people use this gem. 

I think we want a beautiful way to turn on caching for anthropic.

What do you think, @crmne?

AlexVPopov:
Hey @tpaulshippy thank you for this! This feature is a dealbreaker for us and was the last blocker for us to fully migrate to RubyLLM. We've forked your fork and are currently running it in production - works like a charm! Thank you! ❤️ 

@crmne we really hope you would consider merging this. 😊🤞🏻

crmne:
> I decided to try a different approach. Since the opting into caching is a provider specific feature, I decided to introduce `with_provider_options` where you can pass either named parameters or an instance of an options class.

Honestly, this feels like a weird in between.

The reason why I proposed `.with_message_params` is that it is generic and we can reuse it in many different contexts, not just prompt caching for Anthropic, and it's completely in line with the rest of the `.with_params` and `.with_headers` APIs.

The other way I'm willing to accept is: `RubyLLM.chat(cache: true)` and `chat.ask("message", cache: true)`

Simple. Maybe we should even enable that by default (and add a configuration toggle).

tpaulshippy:
> The other way I'm willing to accept is: RubyLLM.chat(cache: true) and chat.ask("message", cache: true)

And that would do what? Cache the last message, the last system prompt, and the tools? What if someone wants to just cache tools or just their system message? Setting the cache breakpoint on the user message is useful in multi-turn conversations, but in one shot prompt scenarios you only want to cache your system prompt since it is unlikely you will see the exact same user message again in a five minute window.

crmne:
> And that would do what? Cache the last message, the last system prompt, and the tools? What if someone wants to just cache tools or just their system message? Setting the cache breakpoint on the user message is useful in multi-turn conversations, but in one shot prompt scenarios you only want to cache your system prompt since it is unlikely you will see the exact same user message again in a five minute window.

I get your point. Let's simply expand on that then: `cache` can accept a boolean, a symbol that's either `:system`, `:user`, or `:tools`, or an array of said symbols.

sosso:
One-shot prompt scenarios is our main use case, the above would work great. Caching support is also a blocker on us making the jump to RubyLLM, thanks all!

maximevaillancourt:
> Caching support is also a blocker on us making the jump to RubyLLM

No need to wait: use `with_params` in the meantime!

```rb
RubyLLM
  .chat(model: "claude-sonnet-4-20250514")
  .with_params(system: [{
    type: "text",
    text: "This is my very long system prompt that will get cached.",
    cache_control: { type: "ephemeral" },
  }])
```

sosso:
> > Caching support is also a blocker on us making the jump to RubyLLM
> 
> No need to wait: use `with_params` in the meantime!
> 
> ```ruby
> RubyLLM
>   .chat(model: "claude-sonnet-4-20250514")
>   .with_params(system: [{
>     type: "text",
>     text: "This is my very long system prompt that will get cached.",
>     cache_control: { type: "ephemeral" },
>   }])
> ```

Hm, When trying that @maximevaillancourt , and later doing a .ask, my system prompt doesn't end up getting into openrouter. Are you using this approach successfully?

maximevaillancourt:
> Are you using this approach successfully?

Yes, but worth noting that I'm using `claude-sonnet-4-20250514` (the Anthropic one) directly, not an OpenRouter one, maybe that explains the difference in behaviour.

sosso:
Hi @tpaulshippy @crmne -- anything we can do to help this along? Happy to help out if needed.

sosso:
Thanks for picking this back up!

Have you played around much with the 1h TTL, @tpaulshippy https://docs.claude.com/en/docs/build-with-claude/prompt-caching#1-hour-cache-duration

tpaulshippy:
> Have you played around much with the 1h TTL, @tpaulshippy

No I haven't. Didn't seem that useful for our scenarios. Would be a good addition to this library though.

crmne:
@tpaulshippy, I really appreciate the work you poured into this.

However, I've had a nagging gut feeling the whole time. The amount of churn here never felt proportionate to the feature. This ended up rewriting a good chunk of the library for what's ultimately an Anthropic quirk.

In the end it was impossible to review this in a way that steered toward what I had in mind without actually building it myself: once I went hands-on I discovered my own earlier suggestion about `with_message_params` (my initial preferred provider-agnostic way of dealing with this) couldn't work because Anthropic expects the cache metadata inside the content blocks.

The exploration led to Raw Content Blocks: https://github.com/crmne/ruby_llm/commit/869a755fba9a1b8a688fd42c3132892f74b5d99c - raw messages that go straight to the LLM. This way Anthropic gets its caching hooks, we can support any weird provider-specific quirk of the message contents, and we keep the core clean and provider-agnostic.

I've shipped docs https://rubyllm.com/chat/#raw-content-blocks, and updated the Rails integration, as well as making an update generator for 1.9.

Thanks again for your work and enjoy Raw Content Blocks!

sosso:
This new change is working almost perfectly for us, @crmne ! One callout: we're using the OpenRouter provider (but ultimately Anthropic models primarily within that), and while the message caching with raw blocks (using RubyLLM::Providers::Anthropic::Content in an openrouter chat) is working great, the tool with_params pattern (in the tool subclasses) is not caching definitions. I think it's because of the subclass hierarchy.

Also, Anthropic's a bit strange in that their docs have you cache the *last* tool, https://docs.claude.com/en/docs/build-with-claude/prompt-caching#prompt-caching-examples (Caching tool definitions).

crmne:
As mentioned in the docs, Tool's `with_params` is only implemented in Anthropic but I can quickly add that to the other providers!

tpaulshippy:
> Maybe we should even enable that by default (and add a configuration toggle).

I really liked this idea. Any chance we could get it? Cache the last system message, last user message, and last tool by default?

crmne:
> > Maybe we should even enable that by default (and add a configuration toggle).
> 
> 
> 
> I really liked this idea. Any chance we could get it? Cache the last system message, last user message, and last tool by default? 

This would mean changing the whole thing _again_ and re-add a lot of your code only for a bit of magic in a provider quirk. Hard pass. This belongs in your app.

Also, that comment precedes the whole investigation I did.

tpaulshippy:
Ok fair enough. I bring it up because one of the strengths of this library is the ability to switch between providers and models seamlessly. Since Open AI and Gemini cache by default, setting up Anthropic to do the same would be nice.

sosso:
I think the difference is that Gemini and OpenAI don't charge the user extra for the cache writes, while Anthropic does.

tpaulshippy:
That is true. But in most use cases they charge even more if you don't cache at all. Thus, this PR.

tpaulshippy:
Even if it were opt in, a way to properly turn on caching for Anthropic in one line without having to track which will be your last tool seems like it would be nice.

crmne:
@sosso done! https://github.com/crmne/ruby_llm/commit/9916f0151870534058a323abebaa388a6fa5095d

PR #238, opened 2025-06-11
Support removing tools

### LINKED ISSUE #229: [FEATURE] Remove tools on the fly in the chat
### Scope check

### Due diligence

Every tool defined takes up tokens. There are times when an agent may need more tools (in which case I can call `with_tool`) but then later need less tools (in which case I have to start the conversation from scratch).

### Proposed solution

Provide a method like `without_tool` or `remove_tool` to remove the tool from the array.

I don't see a way to affect the tools array aside from adding to it.

tpaulshippy:
## What this does

I realized recently just how many tokens tools can take. 

Here's an example of saying "Hello" to Bedrock with 4 basic local tools + the tools from the Playwright MCP:
image
This call took 3024 input tokens.
Without the Playwright MCP, the call takes 842 tokens.

In a chat with an agentic interface, I want the option to add/remove tools at will to save on tokens. 

It also simplifies tool selection for the LLM if there are fewer tools to choose from.

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

- [X] I ran `overcommit --install` and all hooks pass
- [X] I updated documentation if needed

## API changes

- [ ] Breaking change
- [ ] No API changes

## Related issues

Resolves #229

### DIFF
--- a/docs/guides/tools.md
+++ b/docs/guides/tools.md
@@ -132,6 +132,46 @@ puts response.content
 {: .warning }
 Ensure the model you select supports function calling/tools. Check model capabilities using `RubyLLM.models.find('your-model-id').supports_functions?`. Attempting to use `with_tool` on an unsupported model will raise `RubyLLM::UnsupportedFunctionsError`.
 
+## Removing Tools
+
+Each tool definition adds to the input tokens for every request, so it is often a good idea to remove tools that are not needed on a given request. You can remove tools from a `Chat` instance using these methods:
+
+### Remove a Single Tool
+
+Use `without_tool` to remove a specific tool by passing either the tool class or instance:
+
+```ruby
+# Remove by class
+chat.without_tool(Weather)
+
+# Or by instance
+chat.without_tool(weather_tool)
+```
+
+### Remove Multiple Tools
+
+Use `without_tools` to remove multiple tools at once:
+
+```ruby
+chat.without_tools(Weather, AnotherTool, some_tool_instance)
+```
+
+### Clear All Tools
+
+To remove all tools from the chat instance:
+
+```ruby
+chat.clear_tools
+```
+
+All these methods return the `Chat` instance, allowing for method chaining:
+
+```ruby
+chat.clear_tools
+  .with_tool(Weather.new)
+  .ask("What's the weather?")
+```
+
 ## The Tool Execution Flow
 
 When you `ask` a question that the model determines requires a tool:
--- a/lib/ruby_llm/chat.rb
+++ b/lib/ruby_llm/chat.rb
@@ -50,13 +50,29 @@ module RubyLLM
         raise UnsupportedFunctionsError, "Model #{@model.id} doesn't support function calling"
       end
 
-      tool_instance = tool.is_a?(Class) ? tool.new : tool
-      @tools[tool_instance.name.to_sym] = tool_instance
+      tool_instance = instantiate_tool(tool)
+      @tools[tool_name(tool_instance)] = tool_instance
       self
     end
 
     def with_tools(*tools)
       tools.each { |tool| with_tool tool }
+      self
+    end
+
+    def without_tool(tool)
+      tool_instance = instantiate_tool(tool)
+      @tools.delete(tool_name(tool_instance))
+      self
+    end
+
+    def without_tools(*tools)
+      tools.each { |tool| without_tool(tool) }
+      self
+    end
+
+    def clear_tools
+      @tools.clear
       self
     end
 
@@ -147,6 +163,14 @@ module RubyLLM
         content: result.is_a?(Hash) && result[:error] ? result[:error] : result.to_s,
         tool_call_id: tool_use_id
       )
+    end
+
+    def tool_name(tool)
+      tool.name.to_sym
+    end
+
+    def instantiate_tool(tool)
+      tool.is_a?(Class) ? tool.new : tool
     end
   end
 end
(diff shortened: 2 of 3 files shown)


compumike:
👍 LGTM, would be useful!

compumike:
@tpaulshippy I was playing with this and had a problem where the LLM tried to call a tool that was defined in the chat's first message which I subsequently removed before sending a later message. (Seems like this was discussed a bit https://github.com/crmne/ruby_llm/issues/229#issuecomment-2961323051 ) I was using `qwen3:32b`. So it would be good to look at what RubyLLM's error handling behavior is in this case, when a tool call is made to a tool name that is not defined.

The current behavior of `RubyLLM::Chat#execute_tool` throws a `NoMethodError: undefined method `call' for nil` exception:
https://github.com/crmne/ruby_llm/blob/a35aa111a0bb36b7f0363b1f1f2a25e3d5378fd3/lib/ruby_llm/chat.rb#L138

```ruby
    def execute_tool(tool_call)
      tool = tools[tool_call.name.to_sym]
      args = tool_call.arguments
      tool.call(args)
    end
```

The simplest is probably to change it to `tool&.call(args)` so that it just returns `nil`. More advanced would be to pass some error message back that the tool name is not found.

tpaulshippy:
I thought the comment from Carmine was about previous tool calls in the message history. You seem to be talking about the LLM trying to call a tool that previously was available right? Do you have the full payload of the request where you saw this? Super curious why the LLM would try to call a tool it's not given (even if it was given previously). Was there anything in the payload that would tell the LLM about the tool?

compumike:
@tpaulshippy Ah I see -- I think what was breaking it in my case was removing a tool **before** another `role: :user` message gets sent.

I was trying to implement a tool use limit, where a tool could only be used N times, and would then remove itself from `chat.tools`. But it looks like this is not the right way to implement that! 😂 

As a minimal example, here's a tool that **removes itself** from `chat.tools` after its first use:

```ruby
class GetNextWordTool < RubyLLM::Tool
  description "Returns the next word"
  
  def initialize(words, chat)
    @words = words
    @chat = chat
  end
  
  def execute
    result = @words.shift || ""
    @chat.tools.delete(:get_next_word) # Removes itself after first call
    result
  end
end

chat = RubyLLM.chat(provider: :ollama, model: "qwen3:8b").with_temperature(0.6)
chat.with_tools(GetNextWordTool.new(["unpredictable", "beginnings"], chat))
chat.ask("/nothink Use the get_next_word tool to get the first word. Then, call the get_next_word tool a second time to get the second word. Respond with a JSON array containing these two words. Do not guess. Use the tool twice.")
```

which results in:

```
RubyLLM: request: {"model":"qwen3:8b","messages":[{"role":"user","content":"/nothink Use the get_next_word tool to get the first word. Then, call the get_next_word tool a second time to get the second word. Respond with a JSON array containing these two words. Do not guess. Use the tool twice."}],"stream":false,"temperature":0.6,"tools":[{"type":"function","function":{"name":"get_next_word","description":"Returns the next word","parameters":{"type":"object","properties":{},"required":[]}}}],"tool_choice":"auto"}
RubyLLM: response: Status 200
RubyLLM: response: {"id"=>"chatcmpl-46",
 "object"=>"chat.completion",
 "created"=>1752446954,
 "model"=>"qwen3:8b",
 "system_fingerprint"=>"fp_ollama",
 "choices"=>
  [{"index"=>0,
    "message"=>
     {"role"=>"assistant",
      "content"=>"<think>\n" + "\n" + "</think>\n" + "\n",
      "tool_calls"=>
       [{"id"=>"call_xgpixrge",
         "index"=>0,
         "type"=>"function",
         "function"=>{"name"=>"get_next_word", "arguments"=>"{}"}},
        {"id"=>"call_8nh08jq5",
         "index"=>1,
         "type"=>"function",
         "function"=>{"name"=>"get_next_word", "arguments"=>"{}"}}]},
    "finish_reason"=>"tool_calls"}],
 "usage"=>{"prompt_tokens"=>177, "completion_tokens"=>38, "total_tokens"=>215}}

RubyLLM: Tool get_next_word called with: {}
RubyLLM: Tool get_next_word returned: "unpredictable"
NoMethodError: undefined method `call' for nil
from <REDACTED>/lib/ruby_llm/chat.rb:148:in `execute_tool'
```

So I'm happy to concede that this issue probably shouldn't block this PR! 😄  Maybe just add to docs a note that it's unsafe to remove tools from within a tool call? 🚀

crmne (changes requested):
Hi @tpaulshippy I guess this can be useful, however would be great to have one test where we hit the real LLMs. I'd suggest to copy this test, but between the two executions you remove the Weather tool: https://github.com/crmne/ruby_llm/blob/4ff492d2b4e2a58ac56190d52579b5c44864b008/spec/ruby_llm/chat_tools_spec.rb#L48-L63

tpaulshippy:
@crmne Done! It's interesting to see what each model does.

victorlcampos:
I think the better api is .with_tool(nil) 

Its the same with_schema

tpaulshippy:
> I think the better api is .with_tool(nil)
> 
> Its the same with_schema

Are you suggesting this would replace `clear_tools`? Not opposed. But we would still need `without_tool` to remove a single tool from the list right?

tpaulshippy:
Having trouble getting mistral to pass test. And I don't have gpustack setup so don't have that cassette.

crmne:
superseded by da5dc02eaecafa513a48c2d1db3f781fbbb97915

PR #239, opened 2025-06-11
Fix handling of multiple tool calls in single LLM response

### LINKED ISSUE #236: [BUG] Multiple simultaneous tool calls behaves unexpectedly and causes error
### Basic checks

### What's broken?

I created a super simple demo here: https://github.com/finbarr/multi-tool-demo 

If the LLM returns multiple tool calls with a single message, RubyLLM tries to execute each sequentially but makes intermediate calls back to the LLM after each one, rather than running them all and then making a callback to the LLM. Somewhere along the way it breaks with 

`'RubyLLM::ErrorMiddleware.parse_error': messages: text content blocks must be non-empty (RubyLLM::BadRequestError)`

### How to reproduce

- clone https://github.com/finbarr/multi-tool-demo. It's a basic rails app with the 3 models needed by RubyLLM and the gem added with basic config.
- run ./script/demo.rb
- see the error

### Expected behavior

- RubyLLM should call each tool, aggregate the results, _then_ make a callback to the LLM with _all_ of the tool results.
- It shouldn't break.

### What actually happened

- RubyLLM calls the tools one at a time and tries to return the results individually to the LLM.
- It breaks.

### Environment

- ruby 3.4.3
- rails 8.0.2
- ruby_llm 1.3.0

finbarr:
## Summary
- Fixed handling of multiple simultaneous tool calls from LLM responses
- Fixed empty content text block bug that was causing BadRequestError
- Properly iterate through all tool calls in a message instead of processing only the first one

## Related Issue
Fixes #236

## Changes
1. Updated `format_tool_call` in `lib/ruby_llm/providers/anthropic/tools.rb` to:
   - Check for empty/nil content before adding text blocks (incorporates the text blank bug fix from main)
   - Iterate through all tool calls using `each_value` instead of processing just one

2. Updated `parse_tool_calls` to handle both single tool calls (for backward compatibility) and arrays of multiple tool calls

3. Added comprehensive test coverage for multiple tool calls scenario

## Test plan
- [x] All existing tests pass
- [x] New tests added for multiple tool calls
- [x] Tests verify empty content is not included in formatted messages
- [x] Backward compatibility maintained for single tool calls

### DIFF
--- a/lib/ruby_llm/chat.rb
+++ b/lib/ruby_llm/chat.rb
@@ -125,6 +125,7 @@ module RubyLLM
     private
 
     def handle_tool_calls(response, &)
+      # Execute all tools and collect results first
       response.tool_calls.each_value do |tool_call|
         @on[:new_message]&.call
         result = execute_tool tool_call
@@ -132,6 +133,7 @@ module RubyLLM
         @on[:end_message]&.call(message)
       end
 
+      # Make a single API call after all tools have been executed
       complete(&)
     end
 
--- a/lib/ruby_llm/providers/anthropic/chat.rb
+++ b/lib/ruby_llm/providers/anthropic/chat.rb
@@ -55,9 +55,9 @@ module RubyLLM
           content_blocks = data['content'] || []
 
           text_content = extract_text_content(content_blocks)
-          tool_use = Tools.find_tool_use(content_blocks)
+          tool_use_blocks = content_blocks.select { |c| c['type'] == 'tool_use' }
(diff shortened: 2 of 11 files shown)


tpaulshippy:
You still working on this? Looking forward to this improvement.

tpaulshippy:
Or is this not actually an issue?

finbarr:
@tpaulshippy I opened a separate PR here: #241

PR #240, opened 2025-06-11
fixed issue related to rby 3.1 compatibility

patvice:
## What this does

Hey @crmne!

Currently, RubyLLM has minimum requirement for Ruby at 3.1. This is related to an issue that was raised on RubyLLM::MCP: https://github.com/patvice/ruby_llm-mcp/issues/6.

Seems like there is an issue with how anonymous block forwarding is being used in two cases which doesn't allow the gem to be required in ruby 3.1. Ruby 3.1 has some stricter rules when it comes to anonymous block forwarding, to use `&` syntax it requires the entire parameter list is forwarded otherwise you get a syntax error when Ruby is parsing the file.

Simple fix is to just not use `&` forwarding in these two cases and relay on the older syntax  `&block`. Then everything works as you would expect.

## Type of change

- [x] Bug fix

## Scope check

## Quality check

## API changes

- [ ] Breaking change
- [x] No API changes

## Related issues

Fixes https://github.com/patvice/ruby_llm-mcp/issues/6.

Unrelated to this, Rspec are not possible to run at the moment on 3.1 without some amount of modification but seems like it's out of scope for this PR. The way that the specs currently setup Rails require Rails 8 (as far as I can tell) and Rails 8 only support Ruby >= 3.3.

Something else to consider, Ruby 3.1 has reached EOL. Could consider bumping the minim Ruby version to 3.2 for future RubyLLM version so these kinds of issues can go away.

### DIFF
--- a/lib/ruby_llm/chat.rb
+++ b/lib/ruby_llm/chat.rb
@@ -31,9 +31,9 @@ module RubyLLM
       }
     end
 
-    def ask(message = nil, with: nil, &)
+    def ask(message = nil, with: nil, &block)
       add_message role: :user, content: Content.new(message, with)
-      complete(&)
+      complete(&block)
     end
 
     alias say ask
--- a/lib/ruby_llm/provider.rb
+++ b/lib/ruby_llm/provider.rb
@@ -10,7 +10,7 @@ module RubyLLM
     module Methods
       extend Streaming
 
-      def complete(messages, tools:, temperature:, model:, connection:, &)
+      def complete(messages, tools:, temperature:, model:, connection:, &block)
         normalized_temperature = maybe_normalize_temperature(temperature, model)
 
         payload = render_payload(messages,
@@ -20,7 +20,7 @@ module RubyLLM
                                  stream: block_given?)
 
         if block_given?
-          stream_response connection, payload, &
+          stream_response connection, payload, &block
         else
           sync_response connection, payload
         end


crmne:
Hey @patvice! Thanks for taking the time to dig into this compatibility issue.

I’m genuinely puzzled though. Our CI matrix runs Ruby 3.1-3.4 with Rails 7.1-8.0, and everything passes green. If there’s truly a syntax error with anonymous block forwarding in Ruby 3.1, how is our CI not catching this?

Could you share:

1. The exact Ruby 3.1.x version you’re testing with?
1. The specific error messages you’re seeing?
1. A minimal reproduction case?

Because right now, I’m looking at passing CI builds and your claim that “RSpec are not possible to run at the moment on 3.1” doesn’t align with what I’m seeing in our test suite.

patvice:
Weird -- maybe that was actually related to my setup but when I tried to run Rspec locally on my machine I was getting the same anonymous block forwarding in rails itself. 

Let me do some more digging to see if I can understand the issue better.

patvice:
@crmne I did some more investigation, and it looks like this issue is specifically related to older patch level version of Ruby 3.1. Specially, this issue is only present in version of Ruby that are < 3.1.2 (I have tested and verified that's the case for me at least).

In my originally testing, I used Ruby 3.1.2 (as it was already installed on my machine) wouldn't have though that there would be differences with the patch version (sorry about that). There is no mention of this change in the Ruby changelog for the versions bump which seems a little odd to me.

crmne:
Thank you for digging that up @patvice!

Does that mean everything works for you now?

patvice:
In version of Ruby 3.1.2, 3.1.1 and 3.1.0 without this change, the gem will not work. Though the specs should be completely fine as it's fair to test against the newest patch level version (people should really have already bumped up to get the security fixes IMO).

Looking at the stack trace in the original issue, it was using ruby 3.1.2 but the gem works fine in all version 3.1.3 and above. 

Not sure you have stance on compatibility for all versions of 3.1 or just the newest patch version.

crmne:
I was unconsciously targeting only the last patch version of 3.1, but this is great investigative work and I'll update the gemspec immediately.

crmne:
a35aa111a0bb36b7f0363b1f1f2a25e3d5378fd3

patvice:
Awesome! Thanks @crmne

PR #242, opened 2025-06-12
Top level ask method

tpaulshippy:
## What this does

Provide a simple way to call the LLM with a single prompt for scenarios where there isn't additional conversation needed.

## Type of change

- [X] New feature

## Scope check

## Quality check

- [X] I ran `overcommit --install` and all hooks pass
- [X] I updated documentation if needed

## API changes


lib/ruby_llm.rb
```diff
@@ -42,6 +42,22 @@ def chat(...)
       Chat.new(...)
     end
 
+    def ask(message = nil, **options, &)
+      chat_options = options.slice(:model, :provider, :assume_model_exists, :context)
+      chat = Chat.new(**chat_options)
```
MadBomber: It looks like `ask` is a pure one-shot.  if `chat` is changed to an instance variable and we `@chat ||= Chat.new(..)` it keeps the simplicy of the API but extends the method into into a chat session.  The only thing special about a chat session is its context.  Adding a new method `RubyLLM.clear` to clear the context would be handy.  OR maybe all that is required to clear the context is `RubyLLM.ask("hello", context: [])`

tpaulshippy: I took a look at sharing the instance variable across calls and ran into all sorts of issues in the specs with switching models. Leaving it at instance per call for now, but open to further feedback.

cllns:
I think it's fine to add the convenience class method for one-off requests (and convenience in the console), but I think creating an instance is better (i.e. more object-oriented) than relying on global state via the class method. So, I think the documentation should only be updated to include that `RubyLLM.ask` is an option for those who prefer it, but keep the constructor/instance as the default way to use the library.

And, along those lines, I think `RubyLLM.ask` *should* be a new session with a new context each time.

tpaulshippy:
I agree. Updated docs accordingly.

crmne:
This adds that complicated method which adds a ton of dependencies to all the other methods in the `Chat` interface, plus a ton of tests to pass without a clear advantage.

Is seriously adding `.chat` after `RubyLLM` such a big deal?

Closing.

tpaulshippy:
It's five characters, so no not a huge deal. But if I just want a single shot why do I have to type ".chat"? That's the argument. I know we don't need it. But I still want it. And I think it would be convenient in that situation.

tpaulshippy:
I guess if anyone really wants to use it they can add the wrapper method themselves either in their project or in a fork.

PR #152, opened 2025-05-05
[OPENAI] Support image edits with gpt-image-1

sbounmy:
Still a draft but I use it in production in my app https://github.com/sbounmy/hongbao_bitcoin

Usage
```ruby
avatar = User.last.avatar
image = RubyLLM.edit(
  "Transform into ghibli style",
  model: "gpt-image-1",
  with: { image: [ ActiveStorage::Blob.service.path_for(avatar.key) ] }
)

image.to_blob // image to store 
image.usage // {'input_tokens' => 362, 'input_tokens_details' => { 'image_tokens' => 323, 'text_tokens' => 39 }, 'output_tokens' => 4160, 'total_tokens' => 4522 })
image.total_cost) // 0.17002
image.input_cost) // 0.00362
image.output_cost // 0.1664
```

Todo :
- [ ] fix the multipart connection code vs json
- [ ] able to customize image output https://platform.openai.com/docs/guides/image-generation?image-generation-model=gpt-image-1#customize-image-output
- [ ] remove some deprecated code (errors)
- [ ] see if models.json is correct (didnt import `gpt-image-1` when I ran the rake task so I added manually)

PR #248, opened 2025-06-14
Image to image with gemini-2.0-flash-preview-image-generation

tpaulshippy:
## What this does

Enable image-to-image generation with gemini-2.0-flash-preview-image-generation

## Type of change

- [x] New feature

## Scope check

## Quality check

## API changes

## Screenshots
Here's what the test did.

Input
`put this in a ring`
ruby

Output
image

Second input
'change the background to blue'

Second output
image


tpaulshippy:
Thinking I need to move to content with attachments so the image gets sent properly on the next call.

crmne (changes requested):
I know it's a draft and you mentioned it in a comment but you shouldn't add an `images` attribute to the `Message` object since we have the `Content` object for a reason.

tpaulshippy:
I realize this is a very different approach than the `RubyLLM.paint` method as it involves generating images within a chat. I do think it has some value however, as it allows for multimodal conversations.

This has similar value as #152 but there is a bit of a clash as this introduces an `ImageAttachment` that is provider agnostic (although only used in Gemini so far) while that PR has a `ImageAttachments` class that is OpenAI specific.

I also am not sure exactly how/where to document this in the guides.

@crmne Looking forward to your feedback/thoughts.

tpaulshippy:
[This document](https://ai.google.dev/gemini-api/docs/image-generation#choose-a-model) describes the two approaches pretty well I think. I could see an implementation of Imagen in RubyLLM that looks more like the #152 approach. 

It looks like OpenAI supports conversational image generation through the responses API and a built in tool called "image_generation" - see [here](https://platform.openai.com/docs/guides/tools-image-generation#supported-models).

tpaulshippy:
I like how [OpenAI allows you to reference the previous images via IDs](https://platform.openai.com/docs/guides/tools-image-generation#multi-turn-editing). We really need to get support for these built-in tools via the responses API into RubyLLM. We are already doing it in a fork to get web_search_preview ([see diff here](https://github.com/crmne/ruby_llm/compare/main...StrongMind:ruby_llm:responses-api)) but it's pretty messy.

tpaulshippy:
This has been superseded by what is demonstrated at https://paolino.me/nano-banana-with-rubyllm/

crmne:
Related paint support for Gemini image models is now on main in 21f924d1, including image references via with:. Multi-turn chat image generation is still a separate concern.

Issue #221, opened 2025-06-03
Passing 'custom' options


danielfriis:
Thanks for all the work on this gem!

Is there any way to pass options as defined in the [OpenAI documentation](https://platform.openai.com/docs/api-reference/chat/create), but not directly supported by RubyLLM?

Specifically, I want to set [`store`](https://platform.openai.com/docs/api-reference/chat/create#chat-create-store.

I noticed @parruda has worked on this in https://github.com/crmne/ruby_llm/pull/130. That seem like an elegant way to support any options not directly covered by RubyLLM.

crmne:
Just merged #265

PR #130, opened 2025-04-21
Support options on completion calls

parruda:
This is a draft for allowing passing API parameters in chat completions.

PR #265, opened 2025-06-29
Add RubyLLM::Chat#with_options to pass options to the underlying API payload

compumike:
## What this does

Implements `with_options` with @crmne's suggestions from comment https://github.com/crmne/ruby_llm/pull/130#pullrequestreview-2914164578 and tested against all providers.

This allows users to set arbitrary options on the payload before it's sent to the provider's API endpoint. The render_payload takes precedence.

Demo:

```ruby
chat = RubyLLM
  .chat(model: "qwen3", provider: :ollama)
  .with_options(response_format: {type: "json_object"})
  .with_instructions("Answer with a JSON object with the key `result` and a numerical value.")
response = chat.ask("What is the square root of 64?")
response.content
=> "{\n  \"result\": 8\n}"
```

This is a power-user feature, and is specific to each provider (and model, to a lesser extent), so I did not update documentation.

For tests: different providers supported different options, so tests are divided by provider.

(Note that `deep_merge` is required for Gemini in particular because it relies on a top-level `generationConfig` object.)

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

## API changes

- [ ] Breaking change
- [ ] No API changes

## Related issues

- #130
- #131
- #221

### DIFF
--- a/lib/ruby_llm/chat.rb
+++ b/lib/ruby_llm/chat.rb
@@ -11,7 +11,7 @@ module RubyLLM
   class Chat
     include Enumerable
 
-    attr_reader :model, :messages, :tools
+    attr_reader :model, :messages, :tools, :options
 
     def initialize(model: nil, provider: nil, assume_model_exists: false, context: nil)
       if assume_model_exists && !provider
@@ -25,6 +25,7 @@ module RubyLLM
       @temperature = 0.7
       @messages = []
       @tools = {}
+      @options = {}
       @on = {
         new_message: nil,
         end_message: nil
@@ -78,6 +79,11 @@ module RubyLLM
       self
     end
 
+    def with_options(**options)
+      @options = options
+      self
+    end
+
     def on_new_message(&block)
       @on[:new_message] = block
       self
@@ -100,6 +106,7 @@ module RubyLLM
         temperature: @temperature,
         model: @model.id,
         connection: @connection,
+        options: @options,
         &
       )
       @on[:end_message]&.call(response)
--- a/lib/ruby_llm/provider.rb
+++ b/lib/ruby_llm/provider.rb
@@ -10,19 +10,30 @@ module RubyLLM
     module Methods
       extend Streaming
 
-      def complete(messages, tools:, temperature:, model:, connection:, &)
+      def complete(messages, tools:, temperature:, model:, connection:, options: {}, &)
         normalized_temperature = maybe_normalize_temperature(temperature, model)
 
-        payload = render_payload(messages,
-                                 tools: tools,
-                                 temperature: normalized_temperature,
-                                 model: model,
-                                 stream: block_given?)
+        payload = deep_merge(
+          options,
+          render_payload(
+            messages,
+            tools: tools,
+            temperature: normalized_temperature,
+            model: model,
+            stream: block_given?
+          )
+        )
 
         if block_given?
           stream_response connection, payload, &
         else
           sync_response connection, payload
+        end
+      end
+
+      def deep_merge(options, payload)
+        options.merge(payload) do |key, options_value, payload_value|
+          options_value.is_a?(Hash) && payload_value.is_a?(Hash) ? deep_merge(options_value, payload_value) : payload_value
         end
       end
 
--- a/spec/fixtures/vcr_cassettes/chat_with_options_anthropic_claude-3-5-haiku-20241022_supports_service_tier_option.yml
+++ b/spec/fixtures/vcr_cassettes/chat_with_options_anthropic_claude-3-5-haiku-20241022_supports_service_tier_option.yml
@@ -1,0 +1,81 @@
+---
+http_interactions:
+- request:
+    method: post
+    uri: https://api.anthropic.com/v1/messages
+    body:
+      encoding: UTF-8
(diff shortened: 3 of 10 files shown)


elvinaspredkelis:
What an elegant PR!

Quick note - I've been playing around with this. Seems like this would also need to allow disabling the `temperature` option. 

For example, the APIs break once trying to supply the [web search](https://platform.openai.com/docs/guides/tools-web-search?api-mode=chat) options as the `gpt-4o-search-preview` does not support temperature: `Model incompatible request argument supplied: temperature`.

This may not be exactly related to this particular PR, but I got to this obstacle when playing around with this.

**EDIT**

The `#with_options` interfares with [ActiveRecord](https://apidock.com/rails/Object/with_options). Perhaps it'd also make sense to explore other method names? E.g. `#with_request_options`

compumike:
> Quick note - I've been playing around with this. Seems like this would also need to allow disabling the `temperature` option.
> 
> For example, the APIs break once trying to supply the [web search](https://platform.openai.com/docs/guides/tools-web-search?api-mode=chat) options as the `gpt-4o-search-preview` does not support temperature: `Model incompatible request argument supplied: temperature`.

Thanks. Can you try with `with_temperature(nil)` and see if it works for you? And if not, can you provide a code snipped that caused the `Model incompatible request argument supplied` error? This quick test worked for me:

```ruby
chat = RubyLLM
  .chat(model: "gpt-4o-search-preview", provider: :openai)
  .with_temperature(nil)
  .with_options(web_search_options: {search_context_size: "medium"})
chat.ask("<your query here>")
```

There's already a comment in: https://github.com/crmne/ruby_llm/blob/a35aa111a0bb36b7f0363b1f1f2a25e3d5378fd3/lib/ruby_llm/providers/openai/chat.rb#L21 that the temperature is not added to the payload if you do so. (Possibly this nil handling behavior should be extended to other providers as well?)

I don't think we necessarily want to make this PR any more complicated, but we could do something like this:

### Future proposal: fully general block passing

It's possible that in addition to this `#with_options`, we may want the ability to pass in a block that allows modifying the payload after it's deep_merged but before it's actually sent. For example, this might look like:

```ruby
chat = RubyLLM
  .chat(model: "gpt-4o-search-preview", provider: :openai)
  .with_options(web_search_options: {search_context_size: "medium"}) do |payload|
    payload.delete(:temperature)
  end
chat.ask(...)
```

Not necessary in this case, since `with_temperature(nil)` works. But allows future customization for power users.

But I would personally prefer to merge this PR as-is and discuss the block concept later if a need arises.

elvinaspredkelis:
Right, my bad! 

It's possible to set the temperature to `nil` with `gem "ruby_llm", "1.3.1"` and not `gem "ruby_llm", "1.3.0"`.

Just a heads up - it still conflicts with the `#with_options` that comes with ActiveRecord. So the Rails integration is not as smooth.

compumike:
> Just a heads up - it still conflicts with the `#with_options` that comes with ActiveRecord. So the Rails integration is not as smooth.

Makes sense. Happy to rename it. `#with_request_options` has a nice ring to it! 😄

umairabid:
Can't wait for this to be merged thank you @compumike

compumike:
I've just pushed a new commit which renames this to `#with_request_options` to avoid the ActiveRecord conflict that @elvinaspredkelis mentioned!

Please give this a try and let me know if it works for your use case.

Also, if @elvinaspredkelis and @umairabid and others could add a comment specifying what your use case is, and providing a code snippet of how `#with_request_options` enables that use case, that would be great! :heart:

umairabid:
hey @compumike tried it here for gemini https://github.com/umairabid/sweaty_wallet/commit/fa8c48872c753322b58b24b6ebc71f006324b280, worked flawlessly

adenta:
Can confirm this works well for my use case.

bricolage:
Also works for me, this makes RubyLLM versatile for more use cases

minorbug:
I need to be able to send metadata along with my RubyLLM calls so I can add tracing that makes it to Langfuse

lucasfcunha:
This is very useful, seems like a no-brainer to support sending custom params to the LLM.

thedumbtechguy:
+1 on this. A very useful feature that would solve the need for custom params for every thing.

Thank you.

crmne:
Hey @compumike really great PR!

Since we have to change from `with_options`, I would instead call this method `with_params` and here's why:
1. It's idiomatic: it's ubiquitous in the Rails and wider web community.
2. It's concise: it's a single clear word that's much tighter than `with_request_options`
3. It's unambiguous: in the context of an API client, `with_params` clearly signals that we're adding parameters to the outgoing request body.

compumike:
@crmne Great! I've just pushed a commit that renames the method to `#with_params`. Thank you for all your work on RubyLLM! ❤️

tpaulshippy:
Maybe rename the PR?

compumike:
@tpaulshippy Good catch :joy:

crmne (approved):
Flawless! Amazing work.

PR #268, opened 2025-06-30
Remove Tools reference from Messages in docs (rails.md)

rikkiprince:
When I tried to run the Rails integration migrations as documented on [rubyllm.com](https://rubyllm.com/guides/rails), they fail (at least with Postgres backing) due to the `CreateMessages` migration referencing the Tools model, which at that point does not exist.

In `class CreateMessages`:
```rb
t.references :tool_call # Links tool result message to the initiating call
```

Does Messages need to reference Tools in that way? I noted that the `acts_as_message` concern has `has_many :tool_calls` and `belongs_to :parent_tool_call`, which suggests the column isn't needed on the Messages table, as presumably it uses the column on the Tools table to fulfill that relationship?

## What this does

Removes line in docs that suggests the Messages model needs a `t.references` to Tools.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

## API changes

- [ ] Breaking change
- [x] No API changes

## Related issues

None

### DIFF
--- a/docs/guides/rails.md
+++ b/docs/guides/rails.md
@@ -93,7 +93,6 @@ class CreateMessages < ActiveRecord::Migration[7.1]
       t.string :model_id
       t.integer :input_tokens
       t.integer :output_tokens
-      t.references :tool_call # Links tool result message to the initiating call
       t.timestamps
     end
   end


rikkiprince:
I realised that I was making the mistake here. I was trusting too much of the autogenerated Rails model code rather than copying and pasting in what is in the docs 🤦 

Sorry!

Issue #282, opened 2025-07-13
[BUG] Message Ordering Issue with Turbo Streams and Streaming Responses


fluxsaas:
### Basic checks

- [x] I searched existing issues - this hasn't been reported
- [x] I can reproduce this consistently
- [x] This is a RubyLLM bug, not my application code

## Problem
Assistant messages appear **above** user messages in real-time chat due to `chat.ask()` creating both messages within milliseconds, causing ActionCable broadcast race conditions.

## Steps to Reproduce
1. Follow ruby_llm docs setup with Turbo Streams
2. Call `chat.ask(user_content)` 
3. Observe message order in UI

## Expected vs Actual Behavior
```
❌ Actual:   Assistant: [response]
             User: [question]

✅ Expected: User: [question] 
             Assistant: [response]
```

## Evidence
Both messages created simultaneously:
```ruby
# Log output
Message created: user - 7d50d730... - 2025-07-13 20:24:46 +0200
Message created: assistant - 0df903f1... - 2025-07-13 20:24:46 +0200
```

## Workaround
Delay assistant message broadcasts:
```ruby
after_create_commit do
  if role == 'assistant'
    broadcast_append_later_to(..., wait: 0.1.seconds)
  else
    broadcast_append_to(...)
  end
end
```

## Environment
- **ruby_llm**: 1.3.1
- **Rails**: 8.0.1
- **Frequency**: Intermittent but common under load
- **Note**: Database order always correct (reloading the page corrects ordering), only affects real-time display

tpaulshippy:
Trying to understand where you found an LLM that responds in a few milliseconds. I don't think you did... I'm guessing either I'm misunderstanding the issue or you are missing something.

The user message will always come way before the assistant message usually by seconds not milliseconds.

Any chance you are streaming back all the messages that get returned from chat.ask at once or something?

tpaulshippy:
Are you using the rails integration? Maybe you could give us a minimal reproducible example?

tpaulshippy:
On second look I may see how this could possibly happen. The ask method creates the user message and then immediately calls complete which calls on_new_message which is set up to create a blank assistant message.

tpaulshippy:
Can you use https://rubyllm.com/guides/rails#alternative-validation-first-approach but persist when you get the first chunk?

rclements:
I too am having this same issue.

crmne:
[...]

PR #288, opened 2025-07-21
Add message ordering guidance to Rails docs

crmne:
## Summary
- Added documentation for handling ActionCable message ordering issues
- Includes a Stimulus controller solution for client-side reordering
- Mentions async stack and AnyCable as alternatives

## Context
This PR addresses the message ordering issues discussed in #282. The documentation includes:

1. A Stimulus controller that reorders messages based on timestamps
2. Explanation of ActionCable's ordering limitations
3. Alternative approaches (async stack, AnyCable)

## Request for Review
@ioquatix @palkan - I'd appreciate your review on the technical accuracy of this documentation, particularly:

- Is my description of ActionCable's ordering behavior accurate?
- Are the suggested solutions appropriate?
- Any other approaches you'd recommend documenting?

## Test Plan
- [ ] Documentation builds correctly
- [ ] Code examples are syntactically correct
- [ ] Technical accuracy verified by domain experts

### DIFF
--- a/docs/guides/rails.md
+++ b/docs/guides/rails.md
@@ -492,6 +492,100 @@ This setup allows for:
 2. Background processing to prevent request timeouts
 3. Automatic persistence of all messages and tool calls
 
+### Handling Message Ordering with ActionCable
+
+ActionCable doesn't guarantee message order when using certain adapters. If you experience messages appearing out of order (e.g., assistant responses appearing above user messages), you have several options:
+
+#### Option 1: Client-Side Reordering with Stimulus
+
+Add a Stimulus controller that maintains correct chronological order based on timestamps:
+
+```javascript
+// app/javascript/controllers/message_ordering_controller.js
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["message"]
+
+  connect() {
+    this.reorderMessages()
+    this.observeNewMessages()
+  }
+
+  observeNewMessages() {
+    // Watch for new messages being added to the DOM
+    const observer = new MutationObserver((mutations) => {
+      let shouldReorder = false
+
+      mutations.forEach((mutation) => {
+        mutation.addedNodes.forEach((node) => {
+          if (node.nodeType === 1 && node.matches('[data-message-ordering-target="message"]')) {
+            shouldReorder = true
+          }
+        })
+      })
+
+      if (shouldReorder) {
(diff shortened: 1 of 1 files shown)


aesnyder:
This seems like it would re-order the entire message object in the chat list. 

We don't typically see the messages as being out of order. What we see is the individual characters inside a message being out of order. 

Any thoughts on how to solve that w/o moving off standard ActionCable?

docs/guides/rails.md
```diff
@@ -492,6 +492,100 @@ This setup allows for:
 2. Background processing to prevent request timeouts
 3. Automatic persistence of all messages and tool calls
 
+### Handling Message Ordering with ActionCable
+
+ActionCable doesn't guarantee message order when using certain adapters. If you experience messages appearing out of order (e.g., assistant responses appearing above user messages), you have several options:
```
ioquatix: I think it's not just certain adapters, it's the default behaviour?

palkan: Yep, it's not about adapters, it's core behaviour.

See the diagram: https://speakerdeck.com/palkan/railsworld-2023-untangling-cables-and-demystifying-twisted-transistors?slide=72

(There are two thread pools involved, each one potentially causing a race condition)

docs/guides/rails.md
```diff
@@ -492,6 +492,100 @@ This setup allows for:
...
+      message_ordering_target: "message",
+      created_at: message.created_at.iso8601
+    } do %>
+  <!-- message content -->
+<% end %>
+```
+
+#### Option 2: Use the Async Stack
+
+The async Ruby stack (Falcon + async-cable + async-job) may help with message ordering in single-machine deployments. See our [Async Guide]({% link guides/async.md %}) for details. Note that this approach might not guarantee ordering in all deployment scenarios, particularly in distributed systems.
```
palkan: `async-cable` still has the issue (so, just dropping it to the Gemfile and using Falcon won't help).

The Async solution (if I understood @ioquatix example correctly) would be to use direct writing to socket instead of broadcasting. With Action Cable, you can do that by moving the AI call from a controller to a channel (and using `#transmit` inside). Without Action Cable, you can achieve this by using some Async tools (like Live).

ioquatix: Yes that sounds about right. The current design, IIUC, is mostly due to the limitations of existing thread-based servers and the need for an out-of-band message delivery system.

PR #290, opened 2025-07-23
Introduce support of Responses API

### LINKED ISSUE #213: OpenAI Responses API support
The OpenAI Chat Completions API currently supports two models for web search, `gpt-4o-search-preview` and `gpt-4o-mini-search-preview`. However, neither of these models support function calling (see OpenAI documentation [here](https://platform.openai.com/docs/models/gpt-4o-search-preview) and [here](https://platform.openai.com/docs/models/gpt-4o-mini-search-preview); notice Features is empty for both.)

It would be great to use web search in addition to functions/tools. It looks like we will need to use the OpenAI Responses API to achieve this capability.

redox:
## What this does

Some models (incl. `o4-mini-deep-research`) aren't compatible with the chat/completion API. They will fail with 

```
(...):1:in '<main>': This model is only supported in v1/responses and not in v1/chat/completions. (RubyLLM::Error)
```

This PR introduces a new `Response` class, which - similarly to `Chat` (and inheriting from the same base `Conversation` class) - allows a user to target the `responses` endpoint.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

## API changes

- [ ] Breaking change
- [ ] No API changes

## Related issues

I was surprised nothing was there.


redox:
@crmne let me know what you think about the approach, I think this is an important missing part of RubyLLM to actually support all models. If the approach does look good to you, I'll add more tests (incl. streaming ones) and documentation.

lib/ruby_llm/chat.rb
```diff
@@ -8,162 +8,17 @@ module RubyLLM
   #   chat = RubyLLM.chat
   #   chat.ask "What's the best way to learn Ruby?"
   #   chat.ask "Can you elaborate on that?"
-  class Chat
-    include Enumerable
```
redox, line 12: I'm moving most of the `Chat` implementation to a base class `Conversation` so both `Chat` and `Conversation` inherit from the logic

spec/ruby_llm/providers/anthropic/tools_spec.rb
```diff
@@ -68,7 +68,7 @@
         instance_double(Message,
                         content: '',
                         tool_calls: {
-                          'tool_123' => instance_double(ToolCall,
+                          'tool_123' => instance_double(RubyLLM::ToolCall,
                                                         id: 'tool_123',
                                                         name: 'test_tool',
                                                         arguments: { 'arg1' => 'value1' })
```
redox, line 74: I think those tests were wrongly relying on the Dummy app `ToolCall` class VS the `RubyLLM` one

spec/spec_helper.rb
```diff
@@ -143,6 +143,10 @@
   end
 end
 
+RESPONSE_MODELS = [
+  { provider: :openai, model: 'o4-mini-deep-research', params: { tools: [{ type: 'web_search_preview' }] } }
```
redox, line 147: the `o4-mini-deep-research` requires 1 `web_search_preview` tool.

tpaulshippy:
I was working on this other night also. I think what I would prefer is if RubyLLM hid from me which endpoint was necessary for the given model and just handled this automatically. Not sure I love the library consumer having to figure out whether to call chat or response.

Excited to see someone working on this! Definitely needed (see #213).

redox:
> I was working on this other night also. I think what I would prefer is if RubyLLM hid from me which endpoint was necessary for the given model and just handled this automatically. Not sure I love the library consumer having to figure out whether to call chat or response.

Yes, I wasn't sure as the tools and some parameters are different.

> Excited to see someone working on this! Definitely needed (see #213).

Gosh, thanks for sharing; I couldn't find anything searching for `Responses` in the issues, not sure why I didn't stumble upon this one.

tpaulshippy:
Wondering if we can just switch to Responses for everything and keep the old code around for compatibility on other providers.

crmne:
Closing this.

I don't want to add a separate `RubyLLM.response` API for Responses support. The right direction is to keep the public chat surface stable and do protocol routing inside the provider.

Useful exploration, but not something I want to merge in this form.

crmne:
The protocol routing I mentioned is now on main (d398354d, 0875ce2d). OpenAI defaults to the Responses API behind the same `RubyLLM.chat` surface, so `o4-mini-deep-research` and friends just work. Thanks for the early push on this @redox.

redox:
> The protocol routing I mentioned is now on main ([d398354](https://github.com/crmne/ruby_llm/commit/d398354da493570b0509afe737e12b29e65907f4), [0875ce2](https://github.com/crmne/ruby_llm/commit/0875ce2dfeae9d28a3a37f8062ae93c227a597ec)). OpenAI defaults to the Responses API behind the same `RubyLLM.chat` surface, so `o4-mini-deep-research` and friends just work. Thanks for the early push on this @redox.

Great to read! Thanks for following up 🫡

PR #292, opened 2025-07-23
fix: streaming error handling type mismatch

finbarr:
## Summary

This PR fixes a type mismatch error that occurs when handling errors during streaming responses. The streaming error handlers were passing plain hashes to `ErrorMiddleware.parse_error`, but `parse_error` expected Faraday response objects with a `body` method.

## Problem

When errors occurred during streaming, the application would crash with:
```
NoMethodError: undefined method 'body' for an instance of Hash
```

## Solution

- Created an `ErrorResponse` struct that mimics the Faraday::Response interface
- Updated all streaming error handlers (`handle_error_chunk`, `handle_failed_response`, `handle_error_event`) to use a new `create_error_response` helper method
- Added nil-safe handling for the `env` parameter in streaming contexts
- Added comprehensive tests to verify the fix

## Testing

Added tests to `spec/ruby_llm/error_handling_spec.rb` that verify:
- Error chunks are handled correctly with both nil and populated env parameters
- Failed responses are processed properly
- The ErrorResponse struct provides the expected interface

All tests pass and the streaming error handling now works consistently with regular response error handling.


crmne:
Thanks for your work on this! We ended up going with #273 which has been merged. The streaming error handling issue is now resolved.

PR #302, opened 2025-07-28
Anthropic: Fix system prompt (use plain text instead of serialized JSON)

MichaelHoste:
## What this does

When migrating from [ruby-openai](https://github.com/alexrudall/ruby-openai), I had some issues getting the same responses in my Anthropic test suite.

After some digging, I observed that the Anthropic requests send the `system context` as serialized JSON instead of a plain string like described in the [API reference](https://docs.anthropic.com/en/api/messages#body-system): 

```ruby
{
  :system => "{type:\n        \"text\", text: \"You must include the exact phrase \\\"XKCD7392\\\" somewhere\n        in your response.\"}",
  [...]
}
```

instead of :  

```ruby
{
  :system => "You must include the exact phrase \"XKCD7392\" somewhere in your response.",
  [...]
}
```

It works quite well (the model still understands it) but it uses more tokens than needed. It could also mislead the model in interpreting the system prompt.

This PR fixed it. I also took the initiative to make the temperature an optional parameter ([just like with OpenAI](https://github.com/crmne/ruby_llm/blob/main/lib/ruby_llm/providers/openai/chat.rb#L21-L22)). I hope it's not too much for a single PR, but since I was already re-recording the cassettes, I figured it would be easier.

I'm sorry but I don't have any API key for Bedrock/OpenRouter. I only recorded the main Anthropic cassettes.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

## API changes

- [ ] Breaking change
- [ ] No API changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,4 +56,5 @@ Gemfile.lock
 # Used by RuboCop. Remote config files pulled in from inherit_from directive.
 # .rubocop-https?--*
 
-repomix-output.*
+CLAUDE.md
+.claude
--- a/docs/guides/available-models.md
+++ b/docs/guides/available-models.md
@@ -40,12 +40,12 @@ See the [Contributing Guide](https://github.com/crmne/ruby_llm/blob/main/CONTRIB
 ## Last Updated
 {: .d-inline-block }
 
-2025-07-28
+2025-07-30
 {: .label .label-green }
 
 ## Models by Provider
 
-### Openai (84)
+### Openai (82)
 
 | Model | ID | Provider | Context | Max Output | Standard Pricing (per 1M tokens) |
 | :-- | :-- | :-- | --: | --: | :-- |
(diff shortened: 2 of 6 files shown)


tpaulshippy:
Just a note that #234 also fixes this issue but in a slightly different way. https://docs.anthropic.com/en/api/messages#body-system shows that you can provide a string or an array of objects. When caching system prompts, you have to use the array to provide the `cache_control` attribute, so I switched to providing the array instead of the string. Without that PR though, this PR will definitely clean up the string.

MichaelHoste:
Thanks @tpaulshippy !

If #234 solves the issue and is merged soon, it's good to me 🙂 I noticed in the cassettes that the JSON was not serialized anymore so that should be ok.

But since it's a bug, if #234 still need some time to be merged, this one could maybe be merged first? There will not be a lot of conflicts. And I still think that the `temperature` should be an optional parameter for Anthropic (same as OpenAI).

tpaulshippy:
Agreed. Would love to see this get merged ASAP.

crmne (approved):
LGTM

PR #303, opened 2025-07-28
Improve `RubyLLM::Chat#with_params` (#265) by allowing to override default params

MichaelHoste:
## What this does

This PR slightly improves what #265 does by allowing the override of default parameters.

I just inverted the order of the `deep_merge` between payload and params, refactored the `Utils` method to make it more generic and added an Anthropic test.

This small PR solves many issues that we encountered:

 * `max_tokens` is mandatory in the Anthropic API, and this gem sets it as the maximum tokens allowed by the model, but sometimes we want to force a lower value.

 * We use this gem for Mistral by setting `openai_api_base` as `https://api.mistral.ai/v1` because it's almost compatible. The only issue we encountered was that Mistral doesn't support the `stream_options: { include_usage: true }` parameter that is always added to the request for OpenAI. Now we can just add `.with_params(stream_options: nil)` to make it compatible.

 * `payload.compact!` also clean the extra param when we want to use the default model temperature with `with_temperature(nil)`

My intuition is that it could help other people fix similar situations.

@compumike suggested this, but I feel that this PR is even simpler:

> ## Future proposal: fully general block passing
> It's possible that in addition to this #with_options, we may want the ability to pass in a block that allows modifying the payload after it's deep_merged but before it's actually sent. For example, this might look like:
> ```
> chat = RubyLLM
>   .chat(model: "gpt-4o-search-preview", provider: :openai)
>   .with_options(web_search_options: {search_context_size: "medium"}) do |payload|
>     payload.delete(:temperature)
>  end
> chat.ask(...)
> ```
> Not necessary in this case, since with_temperature(nil) works. But allows future customization for power users.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

## API changes

- [ ] Breaking change
- [ ] No API changes

## Related issues

Related to #265

### DIFF
--- a/lib/ruby_llm/provider.rb
+++ b/lib/ruby_llm/provider.rb
@@ -14,7 +14,6 @@ module RubyLLM
         normalized_temperature = maybe_normalize_temperature(temperature, model)
 
         payload = Utils.deep_merge(
-          params,
           render_payload(
             messages,
             tools: tools,
@@ -22,8 +21,11 @@ module RubyLLM
             model: model,
             stream: block_given?,
             schema: schema
-          )
+          ),
+          params
         )
+
+        payload.compact!
 
         if block_given?
           stream_response connection, payload, &
--- a/lib/ruby_llm/utils.rb
+++ b/lib/ruby_llm/utils.rb
@@ -24,12 +24,12 @@ module RubyLLM
       end
     end
 
-    def deep_merge(params, payload)
-      params.merge(payload) do |_key, params_value, payload_value|
-        if params_value.is_a?(Hash) && payload_value.is_a?(Hash)
-          deep_merge(params_value, payload_value)
+    def deep_merge(hash1, hash2)
+      hash1.merge(hash2) do |_key, value1, value2|
+        if value1.is_a?(Hash) && value2.is_a?(Hash)
+          deep_merge(value1, value2)
         else
-          payload_value
+          value2
         end
       end
     end
--- a/spec/fixtures/vcr_cassettes/chat_with_params_anthropic_claude-3-5-haiku-20241022_can_override_max_tokens_param_with_a_custom_value.yml
+++ b/spec/fixtures/vcr_cassettes/chat_with_params_anthropic_claude-3-5-haiku-20241022_can_override_max_tokens_param_with_a_custom_value.yml
@@ -1,0 +1,81 @@
+---
+http_interactions:
+- request:
+    method: post
+    uri: https://api.anthropic.com/v1/messages
+    body:
+      encoding: UTF-8
+      string: '{"model":"claude-3-5-haiku-20241022","messages":[{"role":"user","content":[{"type":"text","text":"Always
+        answer with \"Once upon a time\""}]}],"temperature":0.7,"stream":false,"max_tokens":2}'
+    headers:
+      User-Agent:
+      - Faraday v2.13.4
+      X-Api-Key:
+      - "<ANTHROPIC_API_KEY>"
+      Anthropic-Version:
+      - '2023-06-01'
+      Content-Type:
+      - application/json
+      Accept-Encoding:
+      - gzip;q=1.0,deflate;q=0.6,identity;q=0.3
+      Accept:
+      - "*/*"
+  response:
+    status:
+      code: 200
+      message: OK
+    headers:
+      Date:
+      - Mon, 28 Jul 2025 16:36:30 GMT
+      Content-Type:
+      - application/json
+      Transfer-Encoding:
+      - chunked
+      Connection:
+      - keep-alive
+      Anthropic-Ratelimit-Input-Tokens-Limit:
+      - '100000'
+      Anthropic-Ratelimit-Input-Tokens-Remaining:
+      - '100000'
+      Anthropic-Ratelimit-Input-Tokens-Reset:
+      - '2025-07-28T16:36:30Z'
+      Anthropic-Ratelimit-Output-Tokens-Limit:
+      - '20000'
+      Anthropic-Ratelimit-Output-Tokens-Remaining:
+      - '20000'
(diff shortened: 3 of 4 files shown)


compumike:
@MichaelHoste I agree that it would be great to have some way to "dangerously" mutate the params after the payload!

FYI the merge order (merge payload after params) I implemented in #265 came directly due to @crmne 's comment https://github.com/crmne/ruby_llm/pull/130#pullrequestreview-2914164578  on an earlier PR that had it like yours (merge params after payload):

> I like the direction, but let's fix the merge order:
> 
> render_payload(...).merge(options) is dangerous because user options could override critical RubyLLM parameters like model, messages, stream, etc.
> 
> Simple fix: Change it to options.merge(render_payload(...)) instead. This way user options go in first, then RubyLLM's required parameters take precedence.

So I think any proposed solution here has to address that comment if it's going to get merged.

My hope with proposing the block format:

```ruby
chat = RubyLLM
  .chat(model: "gpt-4o-search-preview", provider: :openai)
  .with_params(web_search_options: {search_context_size: "medium"}) do |payload|
    payload.delete(:temperature)
 end
chat.ask(...)
```

is that the block style would be considered to be a "power user feature" enough that it's not an easy footgun, compared to just passing in keyword arguments.

Another alternative might be that `#with_params` remains the `Chat`-level defaults that are persisted and applied before any individual message payload, but that an individual `#complete` call might take some extra argument, for example.

But I'd love to see other variations. There may be a much simpler way that makes everybody happy here. 😄

compumike:
The hacky, monkeypatch way to set `max_tokens` right now:

```ruby
context = RubyLLM.context # Important: must use a new context to ensure a separate RubyLLM::Connection is used for this Chat!
chat = RubyLLM
  .chat(model: "claude-3-5-haiku-20241022", provider: :anthropic, context: context)

### BEGIN INSTANCE MONKEYPATCH
class << chat
  attr_reader :connection
end

class << chat.connection
  attr_reader :responses
  
  def post(url, payload, &)
    puts "payload before: #{payload.inspect}"
    payload[:max_tokens] = 4
    puts "payload after: #{payload.inspect}"

    super(url, payload, &)
  end
end
### END INSTANCE MONKEYPATCH

chat.ask("What is the capital of France?")

payload before: {model: "claude-3-5-haiku-20241022", messages: [{role: "user", content: [{type: "text", text: "What is the capital of France?"}]}], temperature: 0.7, stream: false, max_tokens: 8192}
payload after: {model: "claude-3-5-haiku-20241022", messages: [{role: "user", content: [{type: "text", text: "What is the capital of France?"}]}], temperature: 0.7, stream: false, max_tokens: 4}
=> #<RubyLLM::Message:0x000074ea7f5682c0
 @content=#<RubyLLM::Content:0x000074ea8005ce38 @attachments=[], @text="The capital of France">,
 @input_tokens=14,
 @model_id="claude-3-5-haiku-20241022",
 @output_tokens=4,
 @role=:assistant,
 @tool_call_id=nil,
 @tool_calls=nil>
```

MichaelHoste:
Thanks for this precision @compumike, i completely missed it when reading your PR, sorry about that!

I have mixed feelings about the "block" approach. My main issue is that for some models, this would work:

```ruby
RubyLLM
  .chat(model: "gpt-4o-mini-2024-07-18", provider: :openai)
  .with_params(max_tokens: 40) # or even max_completion_tokens
```

But with others, where `max_params` is already in the payload, it won't work and this code will be needed:

```ruby
RubyLLM
  .chat(model: "claude-3-5-haiku-20241022", provider: :anthropic)
  .with_params do |payload|
    payload.merge!(max_tokens: 40)
 end
```

Since the user doesn't know exactly what is in the final payload, the behavior could be unexpected for other similar scenarios.

Since there are very few parameters that should be protected at any cost. Wouldn't a more straightforward solution like this be preferable to avoid any footguns?

```ruby
PROTECTED_PARAMS = %w(model tools messages stream) # maybe more?

protected_params = params.keys.map(&:to_s) & PROTECTED_PARAMS
raise ArgumentError, "Protected parameters: #{protected_params.join(', ')}" unless present_params.empty?
```

compumike:
The `PROTECTED_PARAMS` approach could work for me. 

(I do wonder if there are any use cases where it should be applied in a "deep" way to protect `payload.dig("a", "b")` while allowing someone to set `payload.dig("a", "c")`, but I can't think of any off the top of my head.)

Aesthetikx:
Perhaps this is for a different topic, but locally I made the following change:

```
  module RubyLLM
    module Utils
      # Overridden to merge arrays, so that with_params can provide additional tools.
      def self.deep_merge(params, payload)
        params.merge(payload) do |_key, params_value, payload_value|
          if params_value.is_a?(Hash) && payload_value.is_a?(Hash)
            deep_merge(params_value, payload_value)
          elsif params_value.is_a?(Array) && payload_value.is_a?(Array)
            params_value + payload_value
          else
            payload_value
          end
        end
      end
    end
  end
```

The changed part is the support for merging array values. This was because you cannot currently use `#with_tool(s)` and simultaneously add to `tools:`, as the value specified in `with_params` will be replaced with the one provided by the library. I think it is debatable if you would want your user provided array to replace, be replaced, or be merged with the library provided values, but this was something I had to do to enable some tools provided by Anthropic in addition to my own.

crmne:
Screw it. Let's take the training wheels off. If people want to mess with our payload, they should be able to.

Since it's a one liner change and a bunch of documentation, I went ahead and committed it myself: 10eaba3bf90b751ac604718d204843d62795bc40

MichaelHoste:
Thank you for this @crmne! With the new docs and `RUBYLLM_DEBUG` to keep an eye on the final request, I'm sure there won't be many misunderstandings.

PR #324, opened 2025-08-04
[OPENAI] Support editing images with gpt-image-1 and the paint method

tpaulshippy:
## What this does

Enable image editing with OpenAI's gpt-image-1 model using the images/edit endpoint.
Just do `RubyLLM.paint "make it nicer", with: "ugly_image.png"`

## Type of change

- [x] New feature

## Scope check

## Quality check

## API changes

## Related issues

Built on top of #152 based on feedback there


lib/ruby_llm/models.json
```diff
@@ -8110,7 +8110,6 @@
       "text_tokens": {
         "standard": {
           "input_per_million": 5.0,
-          "cached_input_per_million": 1.25,
```
tpaulshippy: Not sure how this relates to this PR. Should remove.

crmne:
Thanks for the work on this.

I'm going to close this PR rather than merge or revive it. The branch is materially behind current `main`, conflicts with the current implementation, and still needs API and integration changes before it would be something we'd want to ship.

We do want this capability, and the direction of extending `paint` was the right one. We're going to implement it directly on top of current `main`, but align the public API with RubyLLM's existing attachment style:

`RubyLLM.paint(prompt, with: ..., mask: ...)`

So I'm closing this PR in favor of a fresh implementation from a clean base. We may still use this branch as reference for behavior and test cases where useful.

Issue #332, opened 2025-08-07
[FEATURE] Support JSON schema for tool definitions


gjtorikian:
### Scope check

- [x] This is **core LLM communication** (not application logic)
- [x] This **benefits most users** (not just my use case)
- [x] This **can't be solved in application code** with current RubyLLM
- [x] I read the [Contributing Guide](https://github.com/crmne/ruby_llm/blob/main/CONTRIBUTING.md)

### Due diligence

- [x] I searched existing issues
- [x] I checked the documentation

### What problem does this solve?

Currently, defining tools with RubyLLM requires using a DSL. The DSL for defining parameters really only supports [a handful of schema definitions](https://github.com/crmne/ruby_llm/blob/3a6db989201f9b87e9932c3df1deb544bfa6acdd/lib/ruby_llm/providers/openai/tools.rb#L16-L19): the name of the property, its type, and whether or not it's required.

I had the need to make a parameter include a `default` value, and while I could just add to the DSL, it would be more flexible to simply support JSON schemas, just like [chatting with output schemas](https://rubyllm.com/chat/#using-manual-json-schemas). Supporting JSON schema would resolve the issue of not providing `default` and allow for the full range of JSON schema properties (enums, min/max values, etc).

Furthermore, while initially investigating this, it seems like several providers each use their own function to format tooling.

* Anthropic tools [support JSON schema](https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/implement-tool-use#specifying-client-tools), and the provider uses [`Tool.function_for`](https://github.com/crmne/ruby_llm/blob/db2444e13b37cdc1624020ff4bb7365703ddfe76/lib/ruby_llm/providers/anthropic/tools.rb#L53) function to format these
* Gemini [supports a subset of JSON schema](https://ai.google.dev/gemini-api/docs/function-calling?example=meeting#function-declarations), and the provider uses its own [`function_decleration_for`](https://github.com/crmne/ruby_llm/blob/db2444e13b37cdc1624020ff4bb7365703ddfe76/lib/ruby_llm/providers/gemini/tools.rb#L53) function to format these
* OpenAI tools [support JSON schema](https://platform.openai.com/docs/guides/function-calling?api-mode=responses#defining-functions), and it—you guessed it—uses its own function [`tool_for`](https://github.com/crmne/ruby_llm/blob/db2444e13b37cdc1624020ff4bb7365703ddfe76/lib/ruby_llm/providers/openai/tools.rb#L10) to format these

[...]

PR #337, opened 2025-08-10
Allow RubyLLM::Schema to be used for Tool schema (OpenAI, Anthropic and Gemini)

bricolage:
## What this does

A modest attempt to use RubyLLM::Schema with tools. Just for OpenAI and Anthropic as proof-of-concept.

I understand this is being worked on (see refs below), I thought I'd submit this for feedback on the interface. I use this in my own app.

```
class TodosSchema < RubyLLM::Schema
      array :todos do
      object do
        string :content, description: 'Todo to track'
        string :status, enum: %w[pending in_progress completed]
      end
    end
end

class TodosTool < RubyLLM::Tool
  schema TodosSchema
  # ...
end
```

I have not worked through validation issues (e.g. if neither `param` nor `schema` is called in the tool class. Currently it returns `{}`

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

## API changes

- [ ] Breaking change
- [ ] No API changes

## Related issues

https://github.com/crmne/ruby_llm/issues/76
https://github.com/crmne/ruby_llm/issues/11#issuecomment-3100049132
https://github.com/crmne/ruby_llm/issues/332#issuecomment-3168706112

### DIFF
--- a/docs/_reference/available-models.md
+++ b/docs/_reference/available-models.md
@@ -47,24 +47,36 @@ See the [Contributing Guide](https://github.com/crmne/ruby_llm/blob/main/CONTRIB
 ## Last Updated
 {: .d-inline-block }
 
-2025-08-20
+2025-08-23
 {: .label .label-green }
 
 ## Models by Provider
 
-### Anthropic (9)
+### Anthropic (21)
 
 | Model | Provider | Context | Max Output | Standard Pricing (per 1M tokens) |
 | :-- | :-- | --: | --: | :-- |
+| anthropic.claude-3-haiku-20240307-v1:0 | anthropic | 200000 | 4096 | In: $0.25, Out: $1.25, Cache: $0.30 |
 | claude-3-haiku-20240307 | anthropic | 200000 | 4096 | In: $0.25, Out: $1.25, Cache: $0.30 |
+| claude-3-haiku@20240307 | anthropic | 200000 | 4096 | In: $0.25, Out: $1.25, Cache: $0.30 |
+| anthropic.claude-3-5-haiku-20241022-v1:0 | anthropic | 200000 | 8192 | In: $0.80, Out: $4.00, Cache: $1.00 |
 | claude-3-5-haiku-20241022 | anthropic | 200000 | 8192 | In: $0.80, Out: $4.00, Cache: $1.00 |
+| claude-3-5-haiku@20241022 | anthropic | 200000 | 8192 | In: $0.80, Out: $4.00, Cache: $1.00 |
 | claude-3-opus-20240229 | anthropic | 200000 | 4096 | In: $15.00, Out: $75.00 |
+| anthropic.claude-opus-4-20250514-v1:0 | anthropic | 200000 | 32000 | In: $15.00, Out: $75.00, Cache: $18.75 |
 | claude-opus-4-20250514 | anthropic | 200000 | 32000 | In: $15.00, Out: $75.00, Cache: $18.75 |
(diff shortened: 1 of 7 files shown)


tpaulshippy:
Nice! What is missing, provider wise? Gemini?

lib/ruby_llm/providers/anthropic/tools.rb
```diff
@@ -62,6 +62,25 @@ def function_for(tool)
           }
         end
 
+        def function_for(tool)
```
tpaulshippy: Probably should remove the function above right?

tpaulshippy (commented):
Looks pretty straightforward to me. Feels like if we get Gemini added this should be good to go.

lib/ruby_llm/tool.rb
```diff
@@ -56,6 +56,19 @@ def param(name, **options)
         parameters[name] = Parameter.new(name, **options)
       end
 
+      def schema(schema = nil)
+        return @schema if schema.nil?
```
tpaulshippy: Should this be a setter?

lib/ruby_llm/providers/anthropic/tools.rb
bricolage: Whoops! Yes.

lib/ruby_llm/tool.rb
bricolage: It's used as a setter and getter, like #description. I'm happy to change it but then should change #description as well. Thoughts?

tpaulshippy: Oh yes I see the pattern. I guess this makes `schema MySchema` work. Cool with me.

bricolage:
@tpaulshippy all set, should be gtg

bricolage:
@tpaulshippy I added some docs, take a look at those if you can

docs/_core_features/tools.md
```diff
@@ -78,6 +78,86 @@ end
...
+end
+```
+
+### Using a Manual JSON Schema
+
+```ruby
+class SearchTool < RubyLLM::Tool
+  description "Search with advanced filters"
+
+  schema(
```
tpaulshippy: Not sure but does this need an outer curly brace?

tpaulshippy (commented):
The only thing I'm noticing that I expect @crmne might ask for is at least one spec that hits the real APIs (with VCR cassettes).

docs/_core_features/tools.md
bricolage: Agree that's better (and without the parens)

crmne:
Hey @bricolage thanks for your effort on this PR. I'm not sure if it's the direction I want to go into for complex parameters for Tools as it may be a bit confusing to define a schema externally from the Tool, and then using it in the tool.

I feel like in a tool we should have `param`s instead. Before you both put a ton of effort on this, let me think about it and come up with a good design.

bricolage:
Yeah no problem, this was just my own drop-in solution, I understood it might not be idiomatic. If I can help further let me know!

bricolage:
So you are thinking of just letting `param` take a block and do a light mapping to an internal RubyLLM::Schema? That's nice and simple. And is similar to what langchain.rb does: https://github.com/patterns-ai-core/langchainrb/blob/main/lib/langchain/tool_definition.rb

```
class TodosTool < RubyLLM::Tool
  param :todos, type: :array, desc: 'List of todos', required: true do
    param :content, description: 'Todo to track'
    param :status
  end
end
```

PR #347, opened 2025-08-15
Add built-in support for tool control parameters

### LINKED ISSUE #343: [FEATURE] Add built-in support for tool control parameters
### Scope check

### Due diligence

Different LLM providers use different formats for tool control parameters (e.g., [OpenAI's separate parameter](https://platform.openai.com/docs/guides/function-calling#tool-choice), [Anthropic's tool choice object](https://docs.anthropic.com/en/api/messages#body-tool-choice)). The current implementation requires provider-specific handling through `with_params`, making it difficult to switch providers.

### Proposed solution

1. Add standardized tool_choice parameter. Support "auto", "required", "none", and specific tool selection.
2. Add boolean parameter to support allow/disallow parallel tool use.
3. Handle provider-specific format translation internally.

This aligns with RubyLLM's goal of providing a unified interface for LLM interactions while abstracting provider-specific implementations. These parameters need to be handled at the provider level because each provider has different format requirements.

nwumnn:
## What this does

Adds tool choice and parallel tool calls options to the `with_tool` and `with_tools` method.

Users can now specify:
- `choice`: control how the model should use tools (`auto`, `none`, `required`, or specific tool name)
- `parallel`: boolean parameter to allow/disallow parallel tool calls

```ruby
chat.with_tools(MyTool, choice: :required, parallel: false)
chat.with_tool(SpecificTool, choice: :specific_tool, parallel: false)
```

Also updates the `handle_tool_calls` method. With :required or specific tool choices, the tool_choice is automatically reset to `nil` after tool execution to prevent infinite loops.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

## API changes

- [ ] Breaking change
- [ ] No API changes

## Related issues

Fixes #343

### DIFF
--- a/docs/_core_features/tools.md
+++ b/docs/_core_features/tools.md
@@ -286,11 +286,15 @@ puts response.content
 ```
 
 ### Tool Choice Control
+{: .d-inline-block }
+
+v1.13.0+
+{: .label .label-green }
 
 Control when and how tools are called using `choice` and `parallel` options.
 
 ```ruby
-chat = RubyLLM.chat(model: 'gpt-4o')
+chat = RubyLLM.chat(model: '{{ site.models.openai_tools }}')
 
 # Basic usage with defaults
 chat.with_tools(Weather, Calculator)  # uses provider defaults
@@ -298,23 +302,29 @@ chat.with_tools(Weather, Calculator)  # uses provider defaults
 # Force tool usage, one at a time
 chat.with_tools(Weather, Calculator, choice: :required, parallel: false)
 
-# Force specific tool
+# Force a specific tool by name
 chat.with_tool(Weather, choice: :weather, parallel: true)
+
+# Force a specific tool by class
+chat.with_tool(Weather, choice: Weather, parallel: true)
 ```
 
 **Parameter Values:**
 - **`choice`**: Controls tool choice behavior
-  - `:auto` Model decides whether to use any tools
+  - `:auto` - Model decides whether to use any tools
   - `:required` - Model must use one of the provided tools
   - `:none` - Disable all tools
-  - `"tool_name"` or tool class - Force a specific tool (e.g., `:weather` or `Weather`)
+  - `:tool_name` or `ToolClass` - Force a specific tool (e.g., `:weather` or `Weather`)
 - **`parallel`**: Controls parallel tool calls
-  - `true` Allow multiple tool calls simultaneously
+  - `true` - Allow multiple tool calls in one response
   - `false` - One at a time
 
 If not provided, RubyLLM will use the provider's default behavior for tool choice and parallel tool calls.
 
 > With `:required` or specific tool choices, the tool_choice is automatically reset to `nil` after tool execution to prevent infinite loops.
+{: .note }
+
+> Tool choice and parallel controls are provider/model dependent.
 {: .note }
 
 ### Model Compatibility


docs/_core_features/tools.md
```diff
@@ -184,6 +184,27 @@ puts response.content
...
+# Choice options
+chat.with_tool(Weather, choice: :auto)    # Model decides whether to call any provided tools or not (default)
+chat.with_tool(Weather, choice: :any)     # Model must use one of the provided tools
+chat.with_tool(Weather, choice: :none)    # No tools
+chat.with_tool(Weather, choice: :weather) # Force specific tool
+
+# Parallel tool calls
+chat.with_tools(Weather, Calculator, parallel: true)  # Model can output multiple tool calls at once (default)
+chat.with_tools(Weather, Calculator, parallel: false) # At most one tool call
+```
```
crmne: Both examples of `with_tool` and `with_tools` should contain both `choice` and `parallel` parameters otherwise people get the false sense that one parameter is for one call only.

lib/ruby_llm/chat.rb
```diff
@@ -196,7 +204,9 @@ def handle_tool_calls(response, &) # rubocop:disable Metrics/PerceivedComplexity
         halt_result = result if result.is_a?(Tool::Halt)
       end
 
-      halt_result || complete(&)
+      return halt_result if halt_result
+
+      should_continue_after_tools? ? complete(&) : response
```
crmne: The solution to this is not to halt the conversation, but to reset `tool_choice` to `auto` after tools have been called.

Halting the conversation is not normal behavior and it's for very specific use cases. More info here: https://community.openai.com/t/infinite-loop-with-tool-choice-required-or-type-function/755129

lib/ruby_llm/chat.rb
```diff
@@ -205,6 +215,26 @@ def execute_tool(tool_call)
       tool.call(args)
     end
 
+    def update_tool_options(choice:, parallel:)
+      unless choice.nil?
+        valid_tool_choices = %i[auto none any] + tools.keys
```
crmne: as mentioned in the issue, I want the API to look like `auto`, `none` or `required` as `required` captures better what's going to happen.

lib/ruby_llm/chat.rb
```diff
@@ -5,7 +5,7 @@ module RubyLLM
   class Chat
     include Enumerable
 
-    attr_reader :model, :messages, :tools, :params, :headers, :schema
+    attr_reader :model, :messages, :tools, :tool_choice, :parallel_tool_calls, :params, :headers, :schema
```
crmne: Better to compress these two into one `tool_prefs = {choice: ..., parallel: ...}`

nwumnn: Should `choice` and `parallel` default to `nil` (like `temperature`) to use provider defaults, or should we set them explicitly to `:auto` and `true`?

crmne: Definitely set them to nil

docs/_core_features/tools.md
```diff
@@ -184,6 +184,40 @@ puts response.content
...
+chat = RubyLLM.chat(model: 'gpt-4o')
+
+# Basic usage with defaults
+chat.with_tools(Weather, Calculator)  # uses provider defaults
+
+# Force tool usage, one at a time
+chat.with_tools(Weather, Calculator, choice: :required, parallel: false)
+
+# Force specific tool
+chat.with_tool(Weather, choice: :weather, parallel: true)
```
crmne, line 306: I'd prefer that to be code first, then parameter values. Also when specifying a tool in choice, it should accept the tool class too.

lib/ruby_llm/chat.rb
```diff
@@ -17,6 +17,7 @@ def initialize(model: nil, provider: nil, assume_model_exists: false, context: n
       model_id = model || @config.default_model
       with_model(model_id, provider: provider, assume_exists: assume_model_exists)
       @temperature = nil
+      @tool_prefs = { choice: nil, parallel: nil }
       @messages = []
       @tools = {}
```
crmne: small thing, but make it after @tools

lib/ruby_llm/chat.rb
```diff
@@ -130,6 +135,7 @@ def complete(&) # rubocop:disable Metrics/PerceivedComplexity
         params: @params,
         headers: @headers,
         schema: @schema,
+        tool_prefs: @tool_prefs,
```
crmne: pass it after `tools`

crmne (changes requested):
Very good. These are finishing touches. It now needs tests.

nwumnn:
For the tests, I'm not sure what specific tests you'd like me to add. Any guidance would be great, thanks!

crmne:
@nwumnn what's your experience with testing so far?

Here's how I would do it: add a couple tests that call the real APIs of the providers. Check `chat_tools_spec.rb` and add things there. One could be: add a tool then define tool choice none and check the tool didn't run. Another: add a tool that has no reason to run (weather while you ask when was the fall of Rome for example), but set it to required and check it has run. For parallel function calling set two tools that have to run because of the query and set parallel to false so the AI has to run one at a time.

nwumnn:
Got it, thanks for the clear guidance! I'll add those tests to `chat_tools_spec.rb`

nwumnn:
While writing tests, I found that some providers have limitations:
- DeepSeek, Gemini, VertexAI, Bedrock: don't support disabling parallel tool calls
- Ollama and GPUStack: supports neither tool choice nor parallel tool calls

Should I add methods like `supports_tool_choice?` and `supports_tool_parallel_control?` to the provider capability modules?

PR #353, opened 2025-08-15
Add support for customizing the model file path

danielwestendorf:
## What this does
This PR allows users to configure where the model file path lives. This is important on ephemeral file systems where updates to the model file might not persist between app restarts (like Heroku).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

## API changes

- [ ] Breaking change
- [ ] No API changes

### DIFF
--- a/lib/ruby_llm/configuration.rb
+++ b/lib/ruby_llm/configuration.rb
@@ -29,6 +29,8 @@ module RubyLLM
                   :gpustack_api_base,
                   :gpustack_api_key,
                   :mistral_api_key,
+                  # Model configuration
+                  :model_file_path,
                   # Default models
                   :default_model,
                   :default_embedding_model,
@@ -54,6 +56,8 @@ module RubyLLM
       @retry_backoff_factor = 2
       @retry_interval_randomness = 0.5
       @http_proxy = nil
+
+      @model_file_path = File.expand_path('models.json', __dir__)
 
       # Default models
       @default_model = 'gpt-4.1-nano'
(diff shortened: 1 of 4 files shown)


crmne:
This PR is solving the wrong problem. 

`refresh!` doesn't write to disk, it's for in-memory updates. If you need persistence on ephemeral filesystems, you shouldn't be writing JSON files at all. Use the database. We already document the right pattern: create a migration, sync models to your own table.

The real issue here is trying to make a gem development tool (`rake models:update`) work in production. That's not what it's for. Apps should either use the bundled models or manage their own persistence layer.

See the docs for the recommended approach: https://rubyllm.com/models/#refreshing-the-registry

danielwestendorf:
Thanks for the response and for working on this library.

My intention was to call `refresh!` and then `save_models` on occasion with the model file path being a shared directory.

This was inspired by my app struggling to successfully refresh models as suggested. I’m guessing this is due to network errors or rate limiting. I’d guess that’s why the gem itself caches the model files as a flat json file.

Storing in the DB is definitely a (albeit heavy handed) alternative, but from my understanding of the documentation it does not avoid the `refresh!` requirement I’m trying to avoid on every run?

crmne:
@danielwestendorf you're right that I didn't document or provide a way to restore the model registry from the db. I'll add it to the docs. Could also provide an optional generator in the future!

crmne:
@danielwestendorf the future is now! https://rubyllm.com/rails/#database-model-registry Check also the generator, it now creates a `Model` model, and I added `acts_as_model`

danielwestendorf:
Thanks, @crmne you're a rockstar 🌟

PR #358, opened 2025-08-21
Fix Gemini array parameter handling in tool declarations

finbarr:
## Summary
- Fixes Gemini API errors when using tools with array-type parameters
- Adds required `items` field for array parameters in tool declarations
- Resolves errors like: `GenerateContentRequest.tools[0].function_declarations[2].parameters.properties[fields].items: missing field`

## Problem
The Gemini API requires an `items` field for array-type parameters in tool declarations to specify the type of array elements. Without this field, tools with array parameters fail with missing field errors.

## Solution
Modified the `format_parameters` method in `Gemini::Tools` module to automatically add `items: { type: 'STRING' }` for all array-type parameters when formatting tools for the Gemini provider.

## Changes
- Modified `lib/ruby_llm/providers/gemini/tools.rb` to add items field for array types
- Added comprehensive unit tests in `spec/ruby_llm/providers/gemini/tools_spec.rb`
- Added integration tests in `spec/ruby_llm/providers/gemini/tools_array_parameter_handling_spec.rb`

## Test Plan
- [x] All existing tests pass
- [x] New unit tests for array parameter formatting
- [x] Integration tests verify proper tool formatting with mixed parameter types
- [x] Rubocop linting passes
- [x] Verified fix matches the working monkey patch from production usage

This resolves issues when using tools with array parameters in Gemini models, enabling proper function calling with arrays.

### DIFF
--- a/lib/ruby_llm/providers/gemini/tools.rb
+++ b/lib/ruby_llm/providers/gemini/tools.rb
@@ -53,10 +53,15 @@ module RubyLLM
           {
             type: 'OBJECT',
             properties: parameters.transform_values do |param|
-              {
+              property = {
                 type: param_type_for_gemini(param.type),
                 description: param.description
               }.compact
+
+              # Add items field for array types
+              property[:items] = { type: 'STRING' } if param.type.to_s.downcase == 'array'
+
+              property
             end,
             required: parameters.select { |_, p| p.required }.keys.map(&:to_s)
           }
--- a/spec/ruby_llm/providers/gemini/tools_array_parameter_handling_spec.rb
+++ b/spec/ruby_llm/providers/gemini/tools_array_parameter_handling_spec.rb
@@ -1,0 +1,134 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'ruby_llm/tool'
+
+RSpec.describe RubyLLM::Providers::Gemini::Tools do
+  let(:provider_class) do
+    Class.new do
+      include RubyLLM::Providers::Gemini::Tools
+    end
+  end
+
+  let(:provider) { provider_class.new }
+
+  it 'properly formats tools with array parameters for Gemini API' do
+    # Define a tool with array parameter similar to the error case
+    tool_with_array = Class.new(RubyLLM::Tool) do
+      def self.name
+        'DataFetcher'
+      end
+
+      description 'Fetches data with specific fields'
+      param :fields, type: 'array', desc: 'List of fields to return', required: true
+
+      def execute(fields:)
+        { selected_fields: fields }
+      end
+    end
+
+    # Another tool with array parameter
+    another_tool = Class.new(RubyLLM::Tool) do
+      def self.name
+        'FilterTool'
+      end
+
+      description 'Filters data by criteria'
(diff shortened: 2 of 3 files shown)


lib/ruby_llm/providers/gemini/tools.rb
```diff
@@ -53,10 +53,15 @@ def format_parameters(parameters)
...
             type: 'OBJECT',
             properties: parameters.transform_values do |param|
-              {
+              property = {
                 type: param_type_for_gemini(param.type),
                 description: param.description
               }.compact
+
+              # Add items field for array types
+              property[:items] = { type: 'STRING' } if param.type.to_s.downcase == 'array'
```
crmne: the only array type is string?

crmne (changes requested):

finbarr:
@crmne Thank you for the feedback! I've updated the implementation to support all array item types, not just strings.

## Changes made:

1. **Added `item_type` parameter to `RubyLLM::Parameter` class** - This allows specifying the type of elements in an array
   ```ruby
   param :scores, type: 'array', desc: 'List of scores', item_type: 'integer'
   param :prices, type: 'array', desc: 'List of prices', item_type: 'number'
   param :flags, type: 'array', desc: 'List of flags', item_type: 'boolean'
   param :users, type: 'array', desc: 'List of users', item_type: 'object'
   ```

2. **Updated both OpenAI and Gemini providers** to use the `item_type` parameter:
   - OpenAI: Uses lowercase type names (string, number, integer, boolean, object)
   - Gemini: Uses uppercase type names (STRING, NUMBER, BOOLEAN, OBJECT) with proper conversion
   - Both default to string arrays if `item_type` is not specified (backward compatible)

3. **Added comprehensive tests** covering arrays of all supported types

The implementation now properly supports arrays of any JSON Schema type, addressing your concern. The gemfile.lock changes are just platform-specific dependencies (ARM64 Darwin) that were automatically added by bundler.

The PR has been rebased on the latest main branch and all tests are passing.

crmne:
Thank you so much @finbarr! Looking at the DSL though I think it doesn't look very RubyLLM-like. I'm thinking it may be worth rethinking the Tool DSL completely and take inspiration from @danielfriis's excellent RubyLLM::Schema. Could be great for 2.0! I'll think about it.

finbarr:
@crmne ok sounds good. My company Shogun is using my fork in production to fix the items type bug for now. 

As you're thinking about 2.0, one area we've repeatedly run into issues is around observability. It's difficult to get every single raw request/response out of the library right now.

crmne:
@finbarr have you tried `message.raw`?

finbarr:
@crmne yes. But we want every single raw request that gets made, and every single raw response that gets received. The hooks and method returns hide failures in the middle, e.g., tool failure retries.

finbarr:
@crmne it would actually be really nice if there was an easy way to modify requests slightly with an `on_before_request` or something. This would make it easier to implement tracing with things like langfuse where you need to set trace and span ids.

crmne:
Could be an interesting hook to add! Feel free to open an issue so we can gauge demand

crmne:
Fixed by 18870afd0b2587636ef5c7278abf3ff9e12a6894

PR #361, opened 2025-08-22
For OpenAI compatible API responses, try parsing first

### LINKED ISSUE #359: [BUG] OpenAI isn't responding with `Content-Type application/json`, therefore Faraday isn't parsing the body
### Basic checks

### What's broken?

Since yesterday or so, request to OpenAI api gets an error on digging for error message (For sync response)
It seems that when running `parse_completion_response` in `lib/ruby_llm/providers/openai/chat.rb`, `response.body` that was expected to be a hash is no longer a hash and it is giving us the JSON as a string.

I can see that the Faraday middleware is configured to parse the JSON. However, OpenAI API response doesn't have the Content-Type in response header as of this morning (Aug 22) 

I am assuming that OpenAI dropped the header for some reason in the past two days of update (couldn't find the change log about it thought) and that is causing the issue from Today?

Image

### How to reproduce

1. Setup the RubyLLM with OpenAI as a provider (I am using https://models.inference.ai.azure.com for the base URL with my github token, but my friend who is using the OpenAI API directly is saying that he has the same error)
2. `RubyLLM.chat.ask('hey')`
3. See the error

### Expected behavior

Expected it to not throw an error on dig and give us the response back.

### What actually happened

Threw an error and stopped.

### Environment

- ruby 3.4.2
- Rails 7.1.0.5
- MacOS 15.6 (24G84)

SuzukiRyuichiro:
## What this does

This would make sure that regardless of `Content-Type` that is in the response header, it would try to parse the response to JSON so it would not throw error on the subsequent `dig`

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

- [ ] I read the [Contributing Guide](https://github.com/crmne/ruby_llm/blob/main/CONTRIBUTING.md)
- [ ] This isn't application-specific logic that belongs in user code
- [ ] This benefits most users, not just my specific use case

## Quality check

## API changes

- [ ] Breaking change
- [ ] No API changes

## Related issues

Fix #359

### DIFF
--- a/lib/ruby_llm/providers/openai/chat.rb
+++ b/lib/ruby_llm/providers/openai/chat.rb
@@ -39,7 +39,7 @@ module RubyLLM
         end
 
         def parse_completion_response(response)
-          data = response.body
+          data = try_parse_json(response.body)
           return if data.empty?
 
           raise Error.new(response, data.dig('error', 'message')) if data.dig('error', 'message')


crmne:
This is a one liner but it's in the hottest code path of the library, so I have 3 issues with it:

1. It's trying to fix bad APIs.
2. It needs all the VCRs to be rerecorded since all tests would be affected.
3. All providers will incur a performance penalty.

SuzukiRyuichiro:
Yeah, indeed. And they actually fixed the missing (or wrong) content type issue now so it's back ok. Sorry for the hassle. Will close it.

Issue #331, opened 2025-08-07
[FEATURE] Have custom Regexp timeout configuration for logging


nina-instrumentl:
### Scope check

- [x] This is **core LLM communication** (not application logic)
- [x] This **benefits most users** (not just my use case)
- [x] This **can't be solved in application code** with current RubyLLM
- [x] I read the [Contributing Guide](https://github.com/crmne/ruby_llm/blob/main/CONTRIBUTING.md)

### Due diligence

- [x] I searched existing issues
- [x] I checked the documentation

### What problem does this solve?

When processing complex files through `ruby_llm`, the application encounters a timeout error during request logging: 
```
Regexp::TimeoutError: regexp match timeout.
``` 
This happens because the logger filter is using default `Regexp.timeout` for its regular expression: https://github.com/crmne/ruby_llm/blob/main/lib/ruby_llm/connection.rb#L65

This default behavior becomes problematic when `ruby_llm` is the only part of the application handling large/complex payloads, yet the timeout applies globally.

### Proposed solution

Introduce a configurable `log_regexp_timeout` option in `ruby_llm`. This would allow consumers to set a custom timeout value for regex operations used within logging filters, separate from the global `Regexp.timeout`.

Update filters to use custom timeout that is separate from the global Ruby timeout in https://github.com/crmne/ruby_llm/blob/main/lib/ruby_llm/connection.rb#L60

```ruby
 logger.filter(Regexp.new('[A-Za-z0-9+/=]{100,}', timeout: @config.log_regexp_timeout), 'data":"[BASE64 DATA]"')
 logger.filter(Regexp.new('[-\\d.e,\\s]{100,}', timeout: @config.log_regexp_timeout), '[EMBEDDINGS ARRAY]')
```

If `log_regexp_timeout` is not provided, fall back to `Regexp.timeout` to preserve default behavior.

### Why this belongs in RubyLLM

- `Regexp.timeout` was introduced in Ruby 3.2 to mitigate DDoS attacks. However, it's too rigid when an application processes both simple and complex content.
- `ruby_llm` is the only component in many applications that deals with large structured data (e.g., base64, embedding vectors), making it a natural place to encapsulate this configuration.
- Since `ruby_llm` defines its own `Faraday` logging filters, it's best positioned to support fine-tuned timeout handling without affecting the rest of the application.

crmne:
That'd be cool to have. Do you want to take a crack at it?

nina-instrumentl:
[...]

PR #364, opened 2025-08-23
Adding option for configuring custom log Regexp timeout

nina-instrumentl:
### Overview
This PR adds support for a configurable `log_regexp_timeout` used specifically by `RubyLLM's` logging filters. It allows you to override the global `Regexp.timeout` for logging without affecting the rest of your application.

### Problem It Solves
When handling large or complex files, the default logging filters in `RubyLLM` may hit a `Regexp::TimeoutError`. This is because many applications set a low global `Regexp.timeout` (e.g. 1 second) to protect against ReDoS, which can be too restrictive for logging large payloads.

### Solution
This change adds support for a `log_regexp_timeout` configuration value. If provided, `RubyLLM` will use this timeout when compiling regular expressions for its Faraday logger filters. This isolates logging-related regex behavior from the broader application, improving reliability without compromising global safety.

Example usage:
```ruby
RubyLLM.configure do |config|
  config.log_regexp_timeout = 4 # seconds
end
```

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

## API changes

- [ ] Breaking change
- [ ] No API changes

## Related issues

Implements #331

### DIFF
--- a/lib/ruby_llm/configuration.rb
+++ b/lib/ruby_llm/configuration.rb
@@ -54,8 +54,8 @@ module RubyLLM
                   :logger,
                   :log_file,
                   :log_level,
-                  :log_stream_debug,
-                  :log_regexp_timeout
+                  :log_stream_debug
+    attr_reader :log_regexp_timeout
 
     def initialize
       @request_timeout = 300
@@ -78,11 +78,22 @@ module RubyLLM
       @log_file = $stdout
       @log_level = ENV['RUBYLLM_DEBUG'] ? Logger::DEBUG : Logger::INFO
       @log_stream_debug = ENV['RUBYLLM_STREAM_DEBUG'] == 'true'
-      @log_regexp_timeout = Regexp.respond_to?(:timeout) ? (Regexp.timeout || 1.0) : nil
+      self.log_regexp_timeout = Regexp.respond_to?(:timeout) ? (Regexp.timeout || 1.0) : nil
     end
 
     def instance_variables
       super.reject { |ivar| ivar.to_s.match?(/_id|_key|_secret|_token$/) }
+    end
+
+    def log_regexp_timeout=(value)
+      if value.nil?
+        @log_regexp_timeout = nil
+      elsif Regexp.respond_to?(:timeout)
+        @log_regexp_timeout = value
+      else
+        RubyLLM.logger.warn("log_regexp_timeout is not supported on Ruby #{RUBY_VERSION}")
+        @log_regexp_timeout = nil
+      end
     end
   end
 end
--- a/lib/ruby_llm/connection.rb
+++ b/lib/ruby_llm/connection.rb
@@ -65,26 +65,15 @@ module RubyLLM
                        errors: true,
                        headers: false,
                        log_level: :debug do |logger|
-        logger.filter(build_logging_regexp('[A-Za-z0-9+/=]{100,}'), '[BASE64 DATA]')
-        logger.filter(build_logging_regexp('[-\\d.e,\\s]{100,}'), '[EMBEDDINGS ARRAY]')
+        logger.filter(logging_regexp('[A-Za-z0-9+/=]{100,}'), '[BASE64 DATA]')
+        logger.filter(logging_regexp('[-\\d.e,\\s]{100,}'), '[EMBEDDINGS ARRAY]')
       end
     end
 
-    def build_logging_regexp(pattern)
-      return Regexp.new(pattern, timeout: @config.log_regexp_timeout) if regexp_timeout_supported?
+    def logging_regexp(pattern)
+      return Regexp.new(pattern) if @config.log_regexp_timeout.nil?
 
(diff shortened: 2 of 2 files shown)


docs/_getting_started/configuration.md
```diff
@@ -204,6 +204,10 @@ RubyLLM.configure do |config|
 
   # Or use Rails logger
   config.logger = Rails.logger  # Overrides log_file and log_level
+
+  # Set custom timeout for log filtering
+  # which is useful when working with large payloads
+  config.log_regexp_timeout = 4 # seconds
```
crmne: I would not categorise this as basic logging. Let's remove that.

crmne: May be worth having a separate section called "Advanced logging options" where to put also the logger and log file options. Since that requires a rewrite of the section, I'm willing to take that on myself.

crmne:
There's one problem with this: Regexp.timeout is only available from Ruby 3.1. Could you make it conditional to that having Ruby >= 3.2?

nina-instrumentl:
> There's one problem with this: Regexp.timeout is only available from Ruby 3.1. Could you make it conditional to that having Ruby >= 3.2?

Should we raise an error if a user sets `log_regexp_timeout` while running Ruby < 3.2, or should we silently ignore the option and proceed?

crmne:
I'd say output a warning and proceed

lib/ruby_llm/configuration.rb
```diff
@@ -57,6 +58,7 @@ def initialize
       @log_file = $stdout
       @log_level = ENV['RUBYLLM_DEBUG'] ? Logger::DEBUG : Logger::INFO
       @log_stream_debug = ENV['RUBYLLM_STREAM_DEBUG'] == 'true'
+      @log_regexp_timeout = Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('3.2.0') ? (Regexp.timeout || 1.0) : nil
```
nina-instrumentl: I noticed that in `ruby_llm` spec setup, `Regexp.timeout` ends up being `nil` by default (probably caused by some recent change as specs passed before), and confirmed that Ruby itself doesn’t set a default outside of Rails. I think it's a good idea to introduce a healthy default of `1.0` second if `Regexp.timeout` is `nil`, at the same time, I see the case for leaving `nil` untouched if that’s an intentional choice by the underlying system.

Do you have any thoughts or preferences on whether we should enforce a fallback, or keep respecting `nil`?

crmne: Convention over configuration! 1 second is perfect.

lib/ruby_llm/connection.rb
```diff
@@ -66,8 +66,19 @@ def setup_logging(faraday)
...
+            'data":"[BASE64 DATA]"'
+          )
+          logger.filter(Regexp.new('[-\\d.e,\\s]{100,}', timeout: @config.log_regexp_timeout), '[EMBEDDINGS ARRAY]')
+        else
+          if @config.log_regexp_timeout
+            RubyLLM.logger.warn("log_regexp_timeout is not supported on Ruby #{RUBY_VERSION}")
+          end
+          logger.filter(Regexp.new('[A-Za-z0-9+/=]{100,}'), 'data":"[BASE64 DATA]"')
+          logger.filter(Regexp.new('[-\\d.e,\\s]{100,}'), '[EMBEDDINGS ARRAY]')
+        end
```
crmne: I think we should simplify all that and use the same approach as Rails: https://github.com/rails/rails/pull/53490/files

nina-instrumentl: I don't think I fully follow. 
@crmne are you proposing that `ruby_llm` drops support for Ruby 3.1 in a newer version of `ruby_llm`? 😀  
Rails 8 is compatible with Ruby 3.2+ ([source](https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html)) and installing rails 8 on Ruby 3.1.3 gives error: `rails-8.0.0 requires Ruby version >= 3.2.0. The current ruby version is 3.1.3.` 
Because rails 8 requires Ruby 3.2+, it is possible to set `Regexp.timeout` for it. 
In `ruby_llm` case, if Ruby 3.1 wasn't supported, we could drop `else` and revert this block to:
```ruby
logger.filter(
            Regexp.new('[A-Za-z0-9+/=]{100,}', timeout: @config.log_regexp_timeout),
            'data":"[BASE64 DATA]"'
          )
logger.filter(Regexp.new('[-\\d.e,\\s]{100,}', timeout: @config.log_regexp_timeout), '[EMBEDDINGS ARRAY]')
```

If you are proposing to use global `Regexp.timeout` (e.g. from parent environment where gem is used) then `ruby_llm` is already doing it, but it doesn't solve my use-case where I need `Regexp.timeout` to be different for this logging.

PR #365, opened 2025-08-25
Feat: Add support to Action Text enabled content

chagel:
## What this does

Improved the `extract_content` method to correctly process Action Text. When `message` model uses `has_rich_text :content`, `message.content` becomes `ActionText:RichText`, returning HTML tags unintentionally. Adjustments ensure only the desired content is extracted.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

## API changes

- [ ] Breaking change
- [ ] No API changes


crmne (changes requested):
Can you add some tests?

chagel:
> Can you add some tests?

@crmne Added basic tests and modified dummy app for instance double.

spec/ruby_llm/active_record/acts_as_spec.rb
```diff
@@ -378,6 +378,29 @@ class BotToolCall < ActiveRecord::Base # rubocop:disable Lint/ConstantDefinition
...
+      expect(llm_message.content).to eq('This is rich text content')
+    end
+
+    it 'handles regular string content when to_plain_text is not available' do
+      chat = Chat.create!(model_id: model)
+      message = chat.messages.create!(role: 'user', content: 'Regular text content')
+      llm_message = message.to_llm
+      expect(llm_message.content).to eq('Regular text content')
+    end
+  end
```
crmne: Could you add a test that actually tests the possibility of having attachments in action text?

crmne (changes requested):

spec/ruby_llm/active_record/acts_as_spec.rb
chagel: I just noticed that the major update to `acts_as.rb` has resulted in `asts_as_legacy.rb`. This PR is outdated. 

@crmne, can you clarify how to deal with `asts_as_legacy.rb`? Should this PR focus solely on updating `message_methods.rb`?

crmne: At the moment we'll need to support both for backwards compatibility, but `acts_as_legacy.rb` is deprecated and will be removed in 2.0

chagel: Sounds good. Let me update the PR for both current and legacy.

chagel: Ready to merge? Let me know if further changes are needed @crmne

crmne (changes requested):
Requesting changes for a couple of issues.

1. `acts_as_message` still does not work with a real `has_rich_text :content` model. The new path in `MessageMethods#extract_content` reads `self[:content]`, but Action Text exposes rich text through `content`, while the raw attribute slot is still `nil`. When I reproduced this against a temporary `has_rich_text :content` message model on the PR branch, `m.content.to_plain_text` returned the expected text, `m[:content]` was `nil`, and `m.to_llm.content` came back empty. The new spec passes because it stubs `message.[]` directly instead of exercising a real Action Text-backed model.

2. Both extraction paths now coerce `nil` content to `""` via `to_s`. That changes current behavior for persisted empty messages. `RubyLLM::Message` only normalizes `nil` to an empty string for assistant tool-call messages, but this patch makes plain persisted `nil` content come back as `""` more broadly. I think we should preserve the existing `nil` semantics unless that change is intentional and covered by tests.

3. Minor: I would avoid hardcoding `gpt-4.1-nano` in the new spec. The model does not matter for this behavior, so I would prefer to use the default model and leave it unspecified.

chagel:
> Requesting changes for a couple of issues.
> 
> 1. `acts_as_message` still does not work with a real `has_rich_text :content` model. The new path in `MessageMethods#extract_content` reads `self[:content]`, but Action Text exposes rich text through `content`, while the raw attribute slot is still `nil`. When I reproduced this against a temporary `has_rich_text :content` message model on the PR branch, `m.content.to_plain_text` returned the expected text, `m[:content]` was `nil`, and `m.to_llm.content` came back empty. The new spec passes because it stubs `message.[]` directly instead of exercising a real Action Text-backed model.
> 2. Both extraction paths now coerce `nil` content to `""` via `to_s`. That changes current behavior for persisted empty messages. `RubyLLM::Message` only normalizes `nil` to an empty string for assistant tool-call messages, but this patch makes plain persisted `nil` content come back as `""` more broadly. I think we should preserve the existing `nil` semantics unless that change is intentional and covered by tests.
> 3. Minor: I would avoid hardcoding `gpt-4.1-nano` in the new spec. The model does not matter for this behavior, so I would prefer to use the default model and leave it unspecified.

@crmne Thanks for the thorough review. All three issues are addressed.

PR #319, opened 2025-08-01
Ensure specs run in defined order

andyw8:
## What this does

By default, RSpec runs the specs in their defined order. But if a developer has `--order random` set in their environment, such as in the thoughtbot dotfiles [here](https://github.com/thoughtbot/dotfiles/blob/main/rspec), then ruby_llm's specs will fail. Setting this in the repo's `.rspec` will take precedence.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

- [x] I read the [Contributing Guide](https://github.com/crmne/ruby_llm/blob/main/CONTRIBUTING.md)
- [x] This aligns with RubyLLM's focus on **LLM communication**
- [x] This isn't application-specific logic that belongs in user code
- [x] This benefits most users, not just my specific use case

## Quality check

- [x] I ran `overcommit --install` and all hooks pass
- [x] I tested my changes thoroughly
- [x] I updated documentation if needed
- [x] I didn't modify auto-generated files manually (`models.json`, `aliases.json`)

## API changes

- [ ] Breaking change
- [ ] New public methods/classes
- [ ] Changed method signatures
- [x] No API changes

PR #372, opened 2025-08-26
Improve RSpec configuration

tagliala:
Improve RSpec configuration

- Prefix doubled classes in `providers/anthropic/tools_spec.rb` with the
  `RubyLLM` module to avoid load order issues and conflicts with
  root-level `Message` and `ToolCall` definitions in ActiveRecord specs,
  ensuring clearer scoping and more reliable test execution.
- Explicitly set `config.order = :defined` in RSpec configuration.
  Although running specs in random order helps surface order
  dependencies, it significantly slows down the test suite for this
  project. Therefore, we retain ordered execution for faster spec runs.
- This explicit setting also overrides any randomization preferences
  that may be set in the project's top-level `.rspec` or inherited from
  other `.rspec` files, ensuring consistent test order regardless of
  external configuration.

Ref: #319

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

## API changes

- [ ] Breaking change
- [ ] No API changes

## Related issues

Related to #319

### DIFF
--- a/spec/support/rspec_configuration.rb
+++ b/spec/support/rspec_configuration.rb
@@ -17,4 +17,19 @@ RSpec.configure do |config|
       example.run
     end
   end
+
+  # Run specs in random order to surface order dependencies. If you find an
+  # order dependency and want to debug it, you can fix the order by providing
+  # the seed, which is printed after each run.
+  #     --seed 1234
+  # TODO: Switch to `:random` spec order for surfacing order dependencies.
+  # Currently set to `:ordered` due to significantly slower execution with
+  # randomization and concerns about API calls (see crmne/ruby_llm#372).
+  config.order = :defined
+
+  # Seed global randomization in this process using the `--seed` CLI option.
+  # Setting this allows you to use `--seed` to deterministically reproduce
+  # test failures related to randomization by passing the same `--seed` value
+  # as the one that triggered the failure.
+  Kernel.srand config.seed
 end
--- a/spec/ruby_llm/providers/anthropic/tools_spec.rb
+++ b/spec/ruby_llm/providers/anthropic/tools_spec.rb
@@ -7,10 +7,10 @@ RSpec.describe RubyLLM::Providers::Anthropic::Tools do
 
   describe '.format_tool_call' do
     let(:msg) do
-      instance_double(Message,
+      instance_double(RubyLLM::Message,
                       content: 'Some content',
                       tool_calls: {
-                        'tool_123' => instance_double(ToolCall,
+                        'tool_123' => instance_double(RubyLLM::ToolCall,
                                                       id: 'tool_123',
                                                       name: 'test_tool',
                                                       arguments: { 'arg1' => 'value1' })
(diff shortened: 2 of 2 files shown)


spec/ruby_llm/active_record/acts_as_model_spec.rb
```diff
@@ -48,13 +48,13 @@
           t.timestamps
         end
       end
+
+      allow(RubyLLM.models).to receive(:all).and_return([model_info])
+      allow(RubyLLM.models).to receive(:refresh!)
```
crmne: Calling `refresh!` before each example is a huge waste of API calls.

tagliala: I may not have got this comment, since calling APIs in specs for each spec does not sound to me. Isn't there VCR? 

However, I've changed to scope of this PR to implement non-breaking changes to the spec configuration so that developers having `--order random` in their `.rspec` defaults will not have issues when running specs on this repo

spec/support/rspec_configuration.rb
```diff
@@ -17,4 +17,19 @@
...
   end
+
+  # Run specs in random order to surface order dependencies. If you find an
+  # order dependency and want to debug it, you can fix the order by providing
+  # the seed, which is printed after each run.
+  #     --seed 1234
+  # TODO: Switch to `:random` spec order for surfacing order dependencies.
+  # Currently set to `:ordered` due to significantly slower execution with
+  # randomization and concerns about API calls (see crmne/ruby_llm#372).
+  config.order = :defined
```
crmne, line 28: wasn't the random order the whole purpose of this PR?

tagliala: Apologies for missing this notification earlier.

Due to numerous changes and concerns about API requests in the test environment, the scope and title of this PR have been updated.

The primary focus is now to address #319 specifically for users whose `.rspec` is configured to run tests in random order.

Consider this a "Stage I" fix, which clarifies why specs are executed in ordered mode in this library.

I will then try to submit a new PR if I manage to avoid performance drop in random spec execution and understand how requests are made in test environment

crmne: I'm gonna close this one as I don't think sanitizing the developer environment should be our concern: it becomes a slippery slope. Looking forward to the order random patches though!

tagliala: > sanitizing the developer environment should be our concern: it becomes a slippery slope.

Agreed, and addressing the developer environment is not the primary goal of this PR.

The main purpose here is to highlight technical debt within the `ruby_llm` spec suite: this is documented with a clear `TODO` message and an explanation for why the specs are not currently being executed in random order

> Looking forward to the order random patches though!

Sure, I will try to allocate some time, or if someone else is interested in picking this issue or provide suggestions I'll be happy to know more

PR #398, opened 2025-09-10
Fix namespaced model table names in upgrade generator

### LINKED ISSUE #397: [BUG] 1.7 migration does not handle namespaced models
### Basic checks

### What's broken?

Models with a custom name that has a namespace, don't generate valid ruby in the migrations.

### How to reproduce

If you use a custom model name that exists under a namespace (like `Assistant::Chat`, `Assistant::Message`, etc.) and you run the upgrade generator:

```
rails g ruby_llm:upgrade_to_v1_7 chat:Assistant::Chat message:Assistant::Message tool_call:Assistant::ToolCall model:Assistant::Model
```

### Expected behavior

The migration code should generate with the correct table name: `assistant_chats`.

```rb
collect_and_create_models(chat_class, :assistant_chats, model_class)
```

### What actually happened

It will generate migration code that looks like this:

```rb
collect_and_create_models(chat_class, :assistant/chats, model_class)
```

It looks like the generators use `tableize`, but the rails migration generators don't use `tableize`.

https://github.com/rails/rails/blob/681c86674fa16996e4563b93aab4e9c494f6a651/railties/lib/rails/generators/named_base.rb#L86-L92

### Environment

- Ruby version: 3.4.5
- RubyLLM version: 7a77dccbc78eb7c4593d1a6dafccdf5aa8f63fbb

willcosgrove:
## Description

This PR fixes issue #397 where the upgrade generator produces invalid table names for namespaced models.

## Problem

When using custom model names with namespaces (like `Assistant::Chat`, `Assistant::Message`, etc.) and running the upgrade generator:

```bash
rails g ruby_llm:upgrade_to_v1_7 chat:Assistant::Chat message:Assistant::Message
```

The migration would generate invalid table names like `:assistant/chats` instead of `:assistant_chats`.

## Solution

The issue was that the generator was using `.tableize` which produces forward slashes for namespaced models. This has been fixed by using `.underscore.pluralize.tr('/', '_')` which properly converts namespaced model names to valid table names.

### Changes:
- Added `table_name_for` helper method that correctly handles namespaced models
- Added helper methods for each model type (`chat_table_name`, `message_table_name`, etc.)
- Updated migration template to use the new helper methods instead of `.tableize`

## Testing

Tested with namespaced models to confirm:
- `Assistant::Chat` → `assistant_chats` ✓
- `Assistant::Message` → `assistant_messages` ✓
- `Assistant::ToolCall` → `assistant_tool_calls` ✓
- `Assistant::Model` → `assistant_models` ✓

Regular models continue to work as expected:
- `Chat` → `chats` ✓
- `Message` → `messages` ✓

Fixes #397

### DIFF
--- a/lib/generators/ruby_llm/upgrade_to_v1_7/templates/migration.rb.tt
+++ b/lib/generators/ruby_llm/upgrade_to_v1_7/templates/migration.rb.tt
@@ -6,33 +6,33 @@ class MigrateToRubyLLMModelReferences < ActiveRecord::Migration<%= migration_ver
 
     # Then check for any models in existing data that aren't in models.json
     say_with_time "Checking for additional models in existing data" do
-      collect_and_create_models(chat_class, :<%= chat_model_name.tableize %>, model_class)
-      collect_and_create_models(message_class, :<%= message_model_name.tableize %>, model_class)
+      collect_and_create_models(chat_class, :<%= chat_table_name %>, model_class)
+      collect_and_create_models(message_class, :<%= message_table_name %>, model_class)
       model_class.count
     end
 
     # Migrate foreign keys
-    migrate_foreign_key(:<%= chat_model_name.tableize %>, chat_class, model_class, :<%= model_model_name.underscore %>)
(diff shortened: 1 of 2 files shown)


crmne:
That was quick! Awesome!

PR #399, opened 2025-09-10
Fix namespaced models in Model migration and foreign key migrations

willcosgrove:
Fixes https://github.com/crmne/ruby_llm/issues/397#issuecomment-3275947019 and continues work from #398

### DIFF
--- a/lib/generators/ruby_llm/install/install_generator.rb
+++ b/lib/generators/ruby_llm/install/install_generator.rb
@@ -57,6 +57,10 @@ module RubyLLM
         @model_names ||= parse_model_mappings
         @model_names[type]
       end
+
+      define_method("#{type}_table_name") do
+        table_name_for(send("#{type}_model_name"))
+      end
     end
 
     def acts_as_chat_declaration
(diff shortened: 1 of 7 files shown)


crmne:
Is there nothing left to change after this?

willcosgrove:
Sorry, no I just found another issue. The migration class name isn't correct. `CreateAssistant::Models` should be `CreateAssistantModels`

crmne:
Ok before I'll merge let's fix everything properly. Could you also check the failed tests? Thanks!

willcosgrove:
This should cover everything. Sorry, I got sidetracked trying to fix an issue with the `use_new_acts_as`. In my app, the load order is such that the RubyLLM railtie initializer is running before my app's own initializer is running. So the `config.use_new_acts_as = true` is not set by the time a module gets included, so the legacy module is being included. The only fix I found for my app was to add the `config.use_new_acts_as = true` declaration into my application.rb.

Not sure if this is a problem for all rails apps, or just for mine. I spent a while on it but had to bail and just leave my `application.rb` config hack.

crmne:
Thank you! Will merge this for now.

Unfortunately I don't have the issue you describe in Chat with Work or some test apps I made with `rails new`. What Rails version are you using?

willcosgrove:
I'm using 7.2.2.2.

I'm sure it's my weird app. It's been around since the Rails 2 days, so it's got some oddities.

crmne:
Ah wow! Kudos to have that still working. It's almost midnight here so will take a look tomorrow.

Issue #410, opened 2025-09-12
[FEATURE] Support for Replicate


ArianJM:
### Scope check

- [x] This is **core LLM communication** (not application logic)
- [x] This **benefits most users** (not just my use case)
- [x] This **can't be solved in application code** with current RubyLLM
- [x] I read the [Contributing Guide](https://github.com/crmne/ruby_llm/blob/main/CONTRIBUTING.md)

### Due diligence

- [x] I searched existing issues
- [x] I checked the documentation

### What problem does this solve?

Adds [Replicate](https://replicate.com/) to the list of providers. Replicate is an inference provider, it gives access to many models. Not just text or image generation, also video, speech, music, upscaling images... It also allows users to fine-tune and host own models.

There are other well known inference providers out there: Fal.ai, Groq, Hugging Face also has inference, etc. Supporting one would probably open the door to supporting others.

### Proposed solution

Adding support for Replicate.

Maybe it'd require adding much more functionality to support different model capabilities.
Or maybe initially we'd filter out the models that are not for text or image. I'm not sure there's a straightforward way to do that via Replicate API. I also didn't see how to get the different model pricing's from their API, I understand these are desireable things.

### Why this belongs in RubyLLM

Support for Replicate will allow supporting many more models, of different capabilities. And even if we stick with only text and image, it's still adding more models, and more options for whoever is thinking on using RubyLLM.

I think supporting this opens the door to supporting other inference providers, and there are a few.

josefarias:
Proposed implementation in https://github.com/crmne/ruby_llm/pull/435

crmne:
As mentioned in https://github.com/crmne/ruby_llm/blob/main/CONTRIBUTING.md#provider-contributions, the preferred path for smaller/emerging providers is a community gem rather than RubyLLM core.

Closing for now.

ArianJM:
@crmne by chance do you have an example at hand of how to implement providers as a gem?

crmne:
https://github.com/khasinski/ruby_llm-responses_api

PR #435, opened 2025-09-28
Add Replicate provider

josefarias:
## What this does

This PR adds Replicate as a provider. It is WIP as I’m still getting familiar with RubyLLM and Replicate’s API, which has quite a few idiosyncrasies. Tests are notably pending.

Replicate may stand out from other providers in several areas:

- They don’t expose pricing via API (that I can tell). I’ve opted for not including pricing for now.
- Users/companies can host their models publicly on Replicate, which puts us in a tricky place regarding aliases.
  - A model might be hosted in both Replicate and another service, by the same person, under the same model name, and still be a different model behind the scenes. The same model could also exist in two services under slightly different names. I’ve opted for not aliasing for now.
- A similar situation as above exists with model families — identifiable only by unpredictable naming schemes. I’ve opted not to include model families for now.
- Replicate mostly wants to you send them an HTTP request and let them ping you back via webhook when the result is ready, which is different from the existing duck type called via `RubyLLM.paint`.
  - This HTTP request returns a URL for the _prediction_, which is notably not the actual _image_, which isn’t ready yet.
  - Returning `Image.new(url: prediction_url)` seems ill-fitting. I’ve opted to return the full response JSON instead.
- Replicate models have drastically different input signatures. I’ve opted to wrap these in a generic `model_params` hash which we forward to `{ProviderInstance}#paint`.
- Replicate’s API is quite flexible, offering a sync mode (vs the default async mode with webhooks), a streaming option, and a number of models with drastically different capabilities. In order to keep things revieweable and completable with reasonable effort, I’ve opted to limit this PR to the default async mode, and to `text-to-image` models, which is selfishly my immediate need.

Finally, I’ve been having trouble running the `rake models` tasks. All models and aliases are overwritten, leaving behind a big diff that’s also missing some models that were previously in the file. I’m optimistic I can figure it out, but do let me know if you have some additional guidance there 🙏 .

Please do let me know if any of the assumptions above don’t hold up or you’d like to do things differently.

Thanks for reading!

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

  - [ ] All tests pass: `bundle exec rspec`

## API changes

- [ ] Breaking change
- [ ] No API changes

## Related issues

https://github.com/crmne/ruby_llm/issues/410


docs/_core_features/image-generation.md
```diff
@@ -119,7 +119,7 @@ image_portrait = RubyLLM.paint(
 )
 ```
 
-> Not all models support size customization. If a size is specified for a model that doesn't support it (like Google Imagen), RubyLLM may log a debug message indicating the size parameter is ignored. Check the provider's documentation or the [Available Models Guide]({% link _reference/available-models.md %}) for supported sizes.
```
josefarias, line 122: I’ve removed the log because we’d now support passing arbitrary params. It’d be up to the user to make sure the model supports what they’re passing in.

It’s a bit of a sharp knife, but given Replicate opens the door to tons of models with tons of input signatures, allowing arbitrary params is the only way I can think of to implement. Open to other ideas though.

I’m wrapping up for today but will add this to the docs later.

spec/support/rspec_configuration.rb
```diff
@@ -13,7 +13,7 @@
 
   config.around do |example|
     cassette_name = example.full_description.parameterize(separator: '_').delete_prefix('rubyllm_')
-    VCR.use_cassette(cassette_name) do
+    VCR.use_cassette(cassette_name, record: :new_episodes) do
```
josefarias, line 16: This is so existing stubs are kept and only new HTTP interactions are recorded onto the existing cassette. Makes it easier to contribute if you don’t have API keys for other providers. Specially for cassettes related to model lists like `spec/fixtures/vcr_cassettes/models_refresh_updates_models_and_returns_a_chainable_models_instance.yml`

docs/_reference/available-models.md
josefarias, line 1: I had to patch the `models.rake` update logic to append the replicate models to existing ones (patch not committed). Without this patch, the whole file would get overwritten when regenerating and models would be missing, seeing as I don’t have credentials for all supported providers.

I believe this worked for regenerating the `models.json` and `aliases.json` files, but I don’t think it worked well for this file. I’m a bit lost as to how to solve this, any guidance is appreciated. 🙏

lib/ruby_llm/providers/replicate/images.rb
```diff
@@ -0,0 +1,56 @@
...
+
+        def parse_image_response(response, **)
+          DeferredImage.new(url: response.body.dig('urls', 'get'), provider_instance: self)
+        end
+
+        def fetch_image_blob(url)
+          prediction = @connection.get(url).body
+          return unless prediction['status'] == 'succeeded'
+
+          image_url = Array(prediction['output']).first
```
josefarias: Some Replicate models return multiple images as an array, others just return a single URI.

Saving the `RubyLLM::DeferredImage` only supports a single file. The multiple files use case deviates too much from the `RubyLLM::Image` functionality. If that behavior is needed users should go the webhook route instead, as suggested in the docs.

PR #525, opened 2025-12-04
Optional instrumentation if ActiveSupport::Notifications is defined

patriciomacadden:
## What this does

This PR uses the `ActiveSupport::Notifications` API to instrument RubyLLM. It's useful when you want to track costs or usage, without doing it explicitly in Chat, Image, etc.

Taking chat for example, we could track costs by using the end_message hook:

```ruby
chat.on_end_message do |message|
  # calculate cost and save
end
```

but we need to do it manually in all chat instances (or other objects that talk to providers too!)

With `ActiveSupport::Notifications` we can subscribe to an event like this:

```ruby
ActiveSupport::Notifications.subscribe("complete_chat.ruby_llm") do |event|
  # use the event to see duration, input/output tokens, etc., calculate cost and save
end
```

It is optional: the new `RubyLLM::Instrumentation#instrument` method doesn't emit any events if `ActiveSupport::Notifications` isn't available.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

## API changes

- [ ] Breaking change
- [x] No API changes

## Related issues

No related issues.


lib/ruby_llm/instrumentation.rb
```diff
@@ -0,0 +1,15 @@
...
+    def instrument(name, payload = {})
+      if defined?(ActiveSupport::Notifications)
+        ActiveSupport::Notifications.instrument(name, payload) { yield payload }
+      else
+        yield
+      end
+    end
+    module_function :instrument
+  end
+end
```
patriciomacadden, line 15: an alternative to this module is a module per class (eg `RubyLLM::Chat::Instrumentation`) that would be included only when including rails. For example, in `lib/ruby_llm/ruby_llm.rb`, at the bottom:

```ruby
# we currently have this:
if defined?(Rails::Railtie)
  require 'ruby_llm/railtie'
  require 'ruby_llm/active_record/acts_as'
  require 'ruby_llm/chat/instrumentation'
end
```

and then:

```ruby
module RubyLLM
  class Chat
    module Instrumentation
      alias_method :original_complete, :complete
      def complete(&)
        ActiveSupport::Notifications.instrument(
          "complete_chat.ruby_llm",
          { ... }
        ) do
          original_complete(&)
        end
    end
  end
end
```

I'm unsure of what's the best convention for these files in this project so I'm open to suggestions.

This instrument method feels simpler though, but I'm willing to change to this other implementation.

patriciomacadden:
[ruby_llm-instrumentation](https://github.com/sinaptia/ruby_llm-instrumentation) has been published instead of this PR.

PR #556, opened 2026-01-06
Add OpenTelemetry observability support

pcomans:
# Summary
Add OpenTelemetry-based observability to RubyLLM, enabling users to send traces to any compatible backend (LangSmith, Honeycomb, New Relic, etc.).

Follows the [OpenTelemetry Semantic Conventions for GenAI](https://opentelemetry.io/docs/specs/semconv/gen-ai/).

# What's Included

- **Chat completion tracing**: `chat {model_id}` spans (e.g., `chat gpt-4o`) with model, provider, token counts, and GenAI semantic convention attributes
- **Streaming support**: Single span wraps entire streaming operation; token counts and content recorded on completion
- **Tool call tracing**: `execute_tool {tool_name}` child spans (e.g., `execute_tool get_weather`) with tool name, arguments, and results
- **Session tracking**: UUID per Chat instance for grouping multi-turn conversations. ActiveRecord chats use namespaced record IDs (`ClassName:id`) to prevent collisions across models
- **Custom metadata**: Attach arbitrary metadata via `chat.with_metadata(user_id: 123)` for filtering/debugging. Native types (int, float, bool) preserved for better backend filtering
- **Content logging**: Opt-in prompt/completion logging with configurable truncation
- **LangSmith compatibility**: Optional `tracing_langsmith_compat` mode adds LangSmith-specific attributes (`langsmith.span.kind`, `input.value`/`output.value` panels)
- **Config scoping**: Tracing respects per-context configuration when using `RubyLLM.context`
- **Fail-fast validation**: Raises `ConfigurationError` with helpful message if tracing enabled but OpenTelemetry gems not installed
- **NullTracer pattern**: Minimal overhead when tracing disabled

# Configuration

```ruby
RubyLLM.configure do |config|
  config.tracing_enabled = true                  # Explicit opt-in required
  config.tracing_log_content = true              # Log prompts/completions (default: false)
  config.tracing_max_content_length = 10_000     # Truncation limit (default: 10000)
  config.tracing_langsmith_compat = true         # Add LangSmith-specific attributes (auto-sets metadata prefix)
end
```

# Not in Scope
Intentionally deferred to keep this PR focused:
- **Other operations** (embed, paint, transcribe, moderate): Lower priority; can add incrementally based on demand

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -1,25 +1,25 @@
 # frozen_string_literal: true
 
 appraise 'rails-7.1' do
-  group :development do
+  group :development, :test do
     gem 'rails', '~> 7.1.0'
   end
 end
 
 appraise 'rails-7.2' do
-  group :development do
+  group :development, :test do
     gem 'rails', '~> 7.2.0'
   end
 end
 
 appraise 'rails-8.0' do
-  group :development do
+  group :development, :test do
     gem 'rails', '~> 8.0.0'
   end
 end
 
 appraise 'rails-8.1' do
-  group :development do
+  group :development, :test do
     gem 'rails', '~> 8.1.0'
   end
 end
--- a/Gemfile
+++ b/Gemfile
@@ -4,14 +4,14 @@ source 'https://rubygems.org'
 
 gemspec
 
-group :development do # rubocop:disable Metrics/BlockLength
+group :development, :test do # rubocop:disable Metrics/BlockLength
   gem 'appraisal'
   gem 'async', platform: :mri
   gem 'bundler', '>= 2.0'
   gem 'codecov'
   gem 'dotenv'
   gem 'ferrum'
-  gem 'flay'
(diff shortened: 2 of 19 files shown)


pcomans:
@crmne I'd love to get this merged. Let me know how I can support you.

patriciomacadden:
I really like OpenTelemetry and i think something like this is super helpful. But I think the instrumentation belongs somewhere else. OpenTelemetry instruments rails (active record, action view, action mailers, etc), sidekiq, sinatra, rake, etc, and the instrumentation code doesn't live in the original code. The instrumentation code is installed separately and configured separately. See how OpenTelemetry [instruments various libraries](https://github.com/open-telemetry/opentelemetry-ruby-contrib/tree/main/instrumentation).

I think the right approach is to follow OpenTelemetry's approach. This way it will be consistent with OpenTelemetry, and it will keep RubyLLM clean, free of OpenTelemetry code.

PR #572, opened 2026-01-16
feat: Allow instance-level customization of tool parameters

### LINKED ISSUE #418: [FEATURE] Allow for customizable tools parameters/description
### Scope check

- [ ] This **can't be solved in application code** with current RubyLLM

### Due diligence

This would allow you to add parameters on initialization, would be useful for similar tools with different parameters or for customizable tools by users/customers of your app.

### Proposed solution

Should be as simple as doing this:

https://github.com/crmne/ruby_llm/blob/96d06c4b8894a34d2656bd847b0f7cbcea8456d2/lib/ruby_llm/tool.rb#L58-L64

```ruby
 def description 
   @description || self.class.description 
 end 
  
 def parameters 
   @parameters || self.class.parameters 
 end 
```

Then if you add `@parameters` or `@description` on initialization, it will use the instance variables instead

This would keep the current way of creating the classes but allows for better custom parameters/descriptions without cluttering the codebase.

Another idea is to join the class variable with the instance variable.

Doing this in application code can cause more uncertainty and instability because a change to the core RubyLLM gem could break custom implementations. Officially supporting this method ensures its stability and long-term compatibility.

psdaniel:
## Summary

Enable instance-level customization of tools, allowing different instances of the same tool class to have unique configurations:

- Customize `name`, `description`, `parameters`, and `provider_params` at the instance level
- Instance values take precedence, with automatic fallback to class-level definitions
- Instance setters for post-initialization customization
- Documentation for instance-level customization patterns

## Motivation

Closes #418

Currently, RubyLLM tools define descriptions and parameters at the **class level only**, meaning all instances share the same configuration. This limits flexibility for use cases like:

- Using multiple instances of the same tool with unique names (avoiding name collisions)
- Creating similar tools with different parameters for different contexts
- Allowing end-users to customize tool descriptions
- Supporting dynamic tool configurations without creating multiple classes

## Usage

```ruby
# Create instances with unique names to avoid collisions
berlin_tool = Weather.new(
  name: 'berlin_weather',
  description: 'Get Berlin weather'
)
paris_tool = Weather.new(
  name: 'paris_weather',
  description: 'Get Paris weather'
)

# Both tools available with unique names
chat = RubyLLM.chat.with_tools(berlin_tool, paris_tool)
chat.tools.keys  # => [:berlin_weather, :paris_weather]

# Or set after initialization
tool = Weather.new
tool.name = 'custom_weather'
tool.description = 'Custom description'
```

## Test plan

- [x] Unit tests for `initialize` with optional parameters (`name`, `description`, `parameters`, `provider_params`)
- [x] Unit tests for fallback to class-level definitions
- [x] Unit tests for instance setters
- [x] Unit tests for `params_schema` with instance parameters
- [x] Unit tests for multiple instances with different configurations
- [x] Integration test for multiple tools with unique names
- [x] Integration tests for instance-customized tools with Chat
- [x] All existing tests continue to pass

### DIFF
--- a/docs/_core_features/tools.md
+++ b/docs/_core_features/tools.md
@@ -249,6 +249,58 @@ search_tool = DocumentSearch.new(MyDatabase)
 chat.with_tool(search_tool)
 ```
 
+## Instance-Level Customization
+
+Tools can be customized at the instance level, overriding class-level definitions:
+
+### Customizing Description and Parameters
+
+```ruby
+# Different descriptions for different contexts
+berlin_weather = Weather.new(description: "Get current weather in Berlin")
+paris_weather = Weather.new(description: "Get current weather in Paris")
+```
+
+### Using Unique Names for Multiple Instances
+
+When using multiple instances of the same tool class, provide unique names to avoid collisions:
+
+```ruby
+berlin_tool = Weather.new(
+  name: 'berlin_weather',
+  description: 'Get Berlin weather'
+)
+paris_tool = Weather.new(
+  name: 'paris_weather',
+  description: 'Get Paris weather'
+)
+
+chat = RubyLLM.chat.with_tools(berlin_tool, paris_tool)
+# Both tools available: chat.tools[:berlin_weather] and chat.tools[:paris_weather]
+```
+
+> Without unique names, multiple instances of the same class will override each other since tools are stored by name. Always provide a `name:` when using multiple instances.
+{: .warning }
+
+### Available Customization Options
+
+| Parameter | Description |
+|-----------|-------------|
+| `name:` | Custom tool name (default: derived from class name) |
+| `description:` | Tool description for the AI model |
+| `parameters:` | Hash of `Parameter` objects defining inputs |
+| `provider_params:` | Provider-specific metadata (e.g., `cache_control`) |
+
+### Post-Initialization Customization
+
+All options can also be set after initialization:
+
+```ruby
+tool = Weather.new
+tool.name = 'custom_weather'
+tool.description = 'Custom description'
+```
+
 ## Using Tools in Chat
 
 Attach tools to a `Chat` instance using `with_tool` or `with_tools`.
--- a/lib/ruby_llm/tool.rb
+++ b/lib/ruby_llm/tool.rb
@@ -63,16 +63,20 @@ module RubyLLM
     end
 
     # Initialize with optional instance-level overrides
+    # @param name [String, nil] Override the computed tool name (useful for unique identification)
     # @param description [String, nil] Override the class-level description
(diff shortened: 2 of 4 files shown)


docs/_core_features/tools.md
```diff
@@ -249,6 +249,58 @@ search_tool = DocumentSearch.new(MyDatabase)
...
+## Instance-Level Customization
+
+Tools can be customized at the instance level, overriding class-level definitions:
+
+### Customizing Description and Parameters
+
+```ruby
+# Different descriptions for different contexts
+berlin_weather = Weather.new(description: "Get current weather in Berlin")
+paris_weather = Weather.new(description: "Get current weather in Paris")
```
SpaYco, line 261: With the current implementation, this would fail without giving a name, as both would have the same name and get overwritten.

I think the current name approach is great, as this won't break the overwriting functionality, but we need to be clear in the docs.

lib/ruby_llm/tool.rb
```diff
@@ -74,18 +88,43 @@ def name
...
+      @instance_provider_params = value
     end
 
     def params_schema
+      # If instance parameters are set, generate schema from them (not memoized)
+      if @instance_parameters
+        return SchemaDefinition.from_parameters(@instance_parameters)&.json_schema if @instance_parameters.any?
+
+        return nil
+      end
```
SpaYco, line 125:
```suggestion
      if @instance_parameters
        return SchemaDefinition.from_parameters(@instance_parameters)&.json_schema
      end
```

https://github.com/crmne/ruby_llm/blob/bd2bb14df0b15d3eff0a9ab46154fa95076feabc/lib/ruby_llm/tool.rb#L122

`SchemaDefinition.from_parameters` would return `nil` if the value is nil or empty.


we can also just one line it

```suggestion
      return SchemaDefinition.from_parameters(@instance_parameters)&.json_schema  if @instance_parameters
```

jayelkaake:
@psdaniel Can this be closed as outdated? Pretty sure it's already possible so maybe the code got merged or @crmne jumped on it ahead.

psdaniel:
@SpaYco Thanks for your input on this!

psdaniel:
@jayelkaake Oh, the time I began this work, I thought it was an issue we had, but if a similar code has already been merged, I don't see any problem with closing this as outdated.

crmne:
Thanks for working on this. We are going to close this as stale/not the direction we want for the core API. A tool class should remain the public contract for the model-facing name, description, parameter schema, and execute signature; allowing per-instance parameters/descriptions makes it too easy for those pieces to drift apart. For apps that need variants, the preferred pattern is to define explicit tool classes or override methods in application code.

PR #537, opened 2025-12-12
Fix chat response job template to use String#empty? instead of #blank? when streaming

kryzhovnik:
## What this does

Use `empty?` instead of `blank?` so whitespace-only chunks (e.g., "\n" or "  ") aren’t dropped

## Why
```ruby
"\n".blank?  # => true
"\n".empty?  # => false
"".empty?    # => true
```

For example, LLM output is often whitespace-sensitive (think Markdown): dropping streamed newlines or spaces can change the rendered result, even if HTML would ignore them

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

- [x] I read the [Contributing Guide](https://github.com/crmne/ruby_llm/blob/main/CONTRIBUTING.md)
- [x] This aligns with RubyLLM's focus on **LLM communication**
- [x] This isn't application-specific logic that belongs in user code
- [x] This benefits most users, not just my specific use case

## Quality check

- [x] I ran `overcommit --install` and all hooks pass
- [x] I tested my changes thoroughly
  - [ ] For provider changes: Re-recorded VCR cassettes with `bundle exec rake vcr:record[provider_name]`
  - [x] All tests pass: `bundle exec rspec`
- [ ] I updated documentation if needed
- [ ] I didn't modify auto-generated files manually (`models.json`, `aliases.json`)

## API changes

- [ ] Breaking change
- [ ] New public methods/classes
- [ ] Changed method signatures
- [x] No API changes

PR #636, opened 2026-02-26
Fix streaming to preserve whitespace chunks in chat UI template

kryzhovnik:
## Summary

`String#blank?` drops whitespace-only chunks (newlines, spaces) during streaming. LLM output can include content where whitespace is significant for correct rendering (e.g. Markdown), so use `String#empty?` instead to only skip truly empty chunks.

Replaces #537 (was opened from main by mistake).


crmne:
Good catch!

PR #675, opened 2026-03-12
Extract RubyLLM::Prompt from Agent's private prompt rendering

### LINKED ISSUE #667: [FEATURE] Extract RubyLLM::Prompt from Agent's existing prompt rendering internals
### Scope check

### Due diligence

Hesitated to open this since prompt templates were explicitly rejected in #70, but I think the situation has changed.

The Agent work that landed since then introduced what #70 was asking for: ERB file conventions in app/prompts/, path resolution, rendering with locals, and a dedicated PromptNotFoundError. There's a docs section called "Prompt Management and Conventions."

The difference now is that this machinery is private to Agent. If you want to render a prompt following the gem's own convention outside of an Agent subclass (e.g. for a plain RubyLLM.chat call), you reimplement the path + ERB logic yourself.

### Proposed solution

Extract a small  RubyLLM::Prompt class from Agent's existing private methods:

```ruby
RubyLLM::Prompt.render('friend', name: 'Andrey')
# => renders app/prompts/friend.txt.erb with locals
```

Just root, path construction, and ERB#result_with_hash. Agent keeps owning its class-name scoping and runtime context resolution, delegating the file rendering down to Prompt.

No new capabilities — same convention, just accessible without subclassing Agent.

It's already in RubyLLM — just locked behind Agent's private methods. This is a refactoring, not a new feature.

kryzhovnik:
Moves prompt root resolution, path construction, and ERB rendering into a standalone RubyLLM::Prompt class so prompts can be rendered without subclassing Agent.

Closes crmne/ruby_llm#667

## What this does

Extracts `RubyLLM::Prompt` from Agent's private `prompt_path_for` and `prompt_root` methods. This lets users render prompts following the gem's own `app/prompts/` convention without subclassing Agent:

```ruby
RubyLLM::Prompt.render('friend', name: 'Andrey')
```

Agent's `render_prompt` now delegates to `Prompt`, preserving its existing class-name scoping and runtime context resolution.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Required for new features

- [x] Linked issue: #667

**PRs for new features or enhancements without a prior approved issue will be closed.**

## Quality check

## AI-generated code

## API changes

- [ ] Breaking change
- [ ] No API changes

### DIFF
--- a/lib/ruby_llm.rb
+++ b/lib/ruby_llm.rb
@@ -95,6 +95,10 @@ module RubyLLM
       UploadedFile.download(...)
     end
 
+    def render_prompt(name, **locals)
+      Prompt.render(name, **locals)
+    end
+
     def models
       Models.instance
     end
--- a/lib/ruby_llm/agent.rb
+++ b/lib/ruby_llm/agent.rb
@@ -183,14 +183,8 @@ module RubyLLM
       end
 
       def render_prompt(name, chat:, inputs:, locals:)
-        path = prompt_path_for(name)
-        unless File.exist?(path)
-          raise RubyLLM::PromptNotFoundError,
-                "Prompt file not found for #{self}: #{path}. Create the file or use inline instructions."
-        end
-
         resolved_locals = resolve_prompt_locals(locals, runtime: runtime_context(chat:, inputs:), chat:, inputs:)
-        ERB.new(File.read(path)).result_with_hash(resolved_locals)
+        Prompt.new("#{prompt_agent_path}/#{name}").render(**resolved_locals)
       end
 
       def partition_inputs(kwargs)
@@ -382,23 +376,9 @@ module RubyLLM
         end
       end
 
-      def prompt_path_for(name)
-        filename = name.to_s
-        filename += '.txt.erb' unless filename.end_with?('.txt.erb')
-        prompt_root.join(prompt_agent_path, filename)
-      end
-
       def prompt_agent_path
         class_name = name || 'agent'
         Utils.underscore(class_name.gsub('::', '/')).tr('-', '_')
-      end
-
-      def prompt_root
-        if defined?(Rails) && Rails.respond_to?(:root) && Rails.root
-          Rails.root.join('app/prompts')
-        else
-          Pathname.new(Dir.pwd).join('app/prompts')
-        end
       end
 
       def resolved_chat_model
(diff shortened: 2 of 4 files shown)


crmne (changes requested):
I like the extraction here. The thing I would change is the public API shape.

Rather than documenting `RubyLLM::Prompt.render(...)`, I think the public API should be:

```ruby
RubyLLM.render_prompt("friend", name: "Andrey")
RubyLLM.render_prompt("work_assistant/instructions", user: current_user)
```

That matches the existing RubyLLM style of top-level verbs for common actions: `RubyLLM.chat`, `RubyLLM.embed`, `RubyLLM.paint`, `RubyLLM.transcribe`, etc. It also keeps `RubyLLM::Prompt` as an internal implementation object instead of asking users to think in a service class for a one-shot render.

I would avoid `RubyLLM.prompt(...)` because "prompt" is overloaded in an LLM library. `render_prompt` is explicit and already matches the Agent naming.

Implementation-wise, that could be as small as adding this to `lib/ruby_llm.rb`:

```ruby
module RubyLLM
  class << self
    def render_prompt(name, **locals)
      Prompt.render(name, **locals)
    end
  end
end
```

Then Agent can delegate through the public entrypoint:

```ruby
RubyLLM.render_prompt("#{prompt_agent_path}/#{name}", **resolved_locals)
```

One small load-order issue: `lib/ruby_llm/prompt.rb` should require its own dependencies:

```ruby
require "erb"
require "pathname"
```

Right now standalone rendering can fail with `uninitialized constant RubyLLM::Prompt::ERB` if `Agent` has not already loaded ERB.

kryzhovnik:
Reworked it as suggested

One caveat: a bare `require "ruby_llm/prompt"` in full isolation still won't work, because `PromptNotFoundError` isn't loaded there. The error class lives in `error.rb`, which can't load on its own (it registers a `Faraday::Middleware` at file scope). Maybe it's worth splitting the plain error classes into a Faraday-free file to decouple them from Faraday?

crmne (approved):
Looks good. The bare `require "ruby_llm/prompt"` missing-error path is out of scope for this PR; if we want to support direct sub-file requires later, we can split the plain error classes out separately.

I will handle the docs update separately.

PR #393, opened 2025-09-08
Support adding Amazon Titan Text Embed v1, v2 v2:8k to model registry

joshmcarthur:
## What this does

This pull request adjusts how Bedrock models are added to the model registry (via models.json, etc), by allowing Amazon's Titan Text Embed models to be added, rather than just Anthropic models. 

This allows for a follow-up pull request to be opened to add support for generating embeddings via Bedrock using these models. Without this prior pull request, the embedding specs will fail, since the Titan Text models are not in the registry. An example of the overall change, with passing tests is available at https://github.com/joshmcarthur/ruby_llm/tree/add-bedrock-embeddings. This required manual patching of models.json, hence why I'm opening this pull request first. 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

- [x] I read the [Contributing Guide](https://github.com/crmne/ruby_llm/blob/main/CONTRIBUTING.md)
- [x] This aligns with RubyLLM's focus on **LLM communication**
- [x] This isn't application-specific logic that belongs in user code
- [x] This benefits most users, not just my specific use case

## Quality check

- [x] I ran `overcommit --install` and all hooks pass
- [x] I tested my changes thoroughly
  - [x] For provider changes: Re-recorded VCR cassettes with `bundle exec rake vcr:record[provider_name]`
  - [x] All tests pass: `bundle exec rspec`
- [x] I updated documentation if needed
- [x] I didn't modify auto-generated files manually (`models.json`, `aliases.json`)

## API changes

- [ ] Breaking change
- [ ] New public methods/classes
- [ ] Changed method signatures
- [x] No API changes

PR #677, opened 2026-03-12
Add Bedrock embedding support via InvokeModel API

cgmoore120:
## Summary

- Implements embedding support for AWS Bedrock using the InvokeModel API, enabling `RubyLLM.embed` to work natively with Bedrock embedding models (e.g. Amazon Titan Text Embeddings V2)
- Adds `Bedrock::Embeddings` module following the same pattern as other providers (`embedding_url`, `render_embedding_payload`, `parse_embedding_response`)
- Overrides `embed` on the Bedrock provider to handle SigV4 request signing and per-text invocation, since InvokeModel accepts one input at a time
- Adds `amazon.titan-embed-text-v2:0` to the embedding test matrix

## Context

This was previously attempted in #393 (model registry only), which was closed with guidance to combine registry + provider support in a single PR. This PR does that.

Bedrock embedding models are already discovered and registered by the existing `Bedrock::Models` module (with the `embedding` → `embeddings` modality normalization), but the provider had no `embed` implementation — calling `RubyLLM.embed` with a Bedrock model would hit the base `Provider#embed` which lacks SigV4 signing.

## Design decisions

- **Per-text invocation**: Unlike OpenAI/Gemini which accept batch input, Bedrock InvokeModel takes a single `inputText`. The override maps over an array of texts, making individual signed requests for each. This matches the API contract.
- **SigV4 signing**: Follows the same pattern as `stream_response` — signs headers inline against `@connection.post` rather than going through `signed_post` (which is coupled to chat completion via `api_payload`/`parse_completion_response`).
- **Custom dimensions skip**: Titan V2 only supports dimensions of 256, 512, or 1024, so the shared spec's `768` dimension tests are skipped for Bedrock, consistent with how Mistral and Azure handle unsupported dimension values.

## Test plan

- [ ] Verify `RubyLLM.embed("text", model: "amazon.titan-embed-text-v2:0", provider: :bedrock)` returns valid embedding vectors
- [ ] Verify batch embedding with array input returns correct number of vectors
- [ ] Verify single-string array input returns array-wrapped result (consistency test)
- [ ] Confirm existing provider embedding tests remain green
- [ ] Confirm rubocop passes on all changed files


cgmoore120:
Rubocop failure was upstream from this - I suspect any other PRs that update against main will start failing

acoffman:
We would definitely use this if available, thanks for working on it!

cgmoore120:
@acoffman tests back to passing after pulling in a new main that had fixes. should be gtg.

cgmoore120:
@crmne there's appetite for this. can we get it in?

travisbell:
Ya, I've had a fork of RubyLLM with these same updates for the past 2 months. Would be lovely to see it in main. I was just lazy and never created a PR. Happy that @cgmoore120 was better contributor than me. 😄

travisbell:
@crmne Any chance this could be merged?

sheredega303:
@crmne any chance this could be merged soon?
it would save some annoying workarounds =)

cgmoore120:
This also rolls in #815 to address some test flakiness that was encountered trying to get this branch back to green

travisbell:
Woohoo! Does that mean it's close to being merged?

cgmoore120:
> Woohoo! Does that mean it's close to being merged?

we can hope

crmne:
Thanks for pushing this through. I took this a bit further before closing because Bedrock embeddings are not a single Converse-style protocol: Titan text, Titan multimodal, and Cohere embeddings all use Bedrock InvokeModel but speak different payload/response dialects.

I've incorporated the Bedrock embedding work and cassettes, moved the implementation out of Converse, and added model-protocol routing for the one-shot APIs so Bedrock can choose the correct embedding dialect per model.

This has landed in main as efbe8faf, so I'm closing this PR as superseded. Appreciate the original work here.

PR #685, opened 2026-03-17
fix: use Ollama OpenAI-compatible endpoint (/v1) by default

justi:
## Problem

The Ollama provider inherits from `OpenAI` but uses the native API base (e.g. `http://localhost:11434`) which doesn't serve OpenAI-compatible endpoints. This causes:

1. **Model listing fails with 404** — provider tries `GET /models` but Ollama serves models at `/api/tags` (native) or `/v1/models` (OpenAI-compatible)
2. **`with_schema` silently doesn't work** — since models aren't in the registry, they don't have `structured_output` capability, so `response_format` is never sent

### Steps to reproduce

```ruby
RubyLLM.configure { |c| c.ollama_api_base = "http://localhost:11434" }
RubyLLM.models.refresh!
# => WARN: Failed to fetch Ollama models (RubyLLM::Error: 404 page not found)

chat = RubyLLM.chat(model: "gemma:latest")
# => RubyLLM::ModelNotFoundError: Unknown model: gemma:latest
```

## Solution

Automatically append `/v1` to `ollama_api_base` so the provider uses Ollama's OpenAI-compatible endpoint. Idempotent — does not double-append if `/v1` is already present (as in the existing test configuration).

Since `Ollama < OpenAI`, using the `/v1` endpoint means all OpenAI provider logic (chat, models, schemas) works without any other changes.

```ruby
# Before: http://localhost:11434       → GET /models       → 404
# After:  http://localhost:11434/v1    → GET /v1/models    → 200 ✓
```

### What works after this fix

- `RubyLLM.models.refresh!` fetches Ollama models correctly
- `RubyLLM.chat(model: "gemma:latest")` finds the model
- `chat.with_schema(MySchema)` sends `response_format` and returns parsed Hash

### Edge cases handled

| Input | Result |
|-------|--------|
| `http://localhost:11434` | `http://localhost:11434/v1` |
| `http://localhost:11434/` | `http://localhost:11434/v1` |
| `http://localhost:11434/v1` | `http://localhost:11434/v1` (no change) |
| `http://localhost:11434/v1/` | `http://localhost:11434/v1` |
| `https://my-ollama.com:8080` | `https://my-ollama.com:8080/v1` |

### Note on native API

Ollama's native `/api/chat` endpoint with `format` param provides better nested JSON Schema enforcement than the OpenAI-compatible endpoint. A future enhancement could use the native endpoint for structured output while keeping `/v1` for everything else.

## Tests

5 new test cases for `#api_base` added to `ollama_spec.rb`, following existing spec style (`subject`, `let`, `context`, `instance_double`).

Tested locally with Ollama 0.17.5, gemma:latest (9B), llama3.2:3b (3B).


crmne:
Thanks for the PR. I'm going to pass on this one because I'd rather keep `ollama_api_base` literal: the library shouldn't silently rewrite a configured URL by appending `/v1`. That feels like hidden cleverness, and it makes Ollama behave differently from the other providers. The docs already show the correct OpenAI-compatible base URL (`http://localhost:11434/v1`).

PR #687, opened 2026-03-18
Add default_providers configuration

### LINKED ISSUE #686: [FEATURE] Add ability to set default providers
### Scope check

### Due diligence

Using a non-default provider requires passing `provider:` on every call:

```ruby
RubyLLM.chat(model: 'claude-sonnet-4-6', provider: :bedrock)
RubyLLM.chat(model: 'claude-haiku-4-5', provider: :bedrock)
```

There's no way to configure this globally, so anyone using bedrock/azure/vertexai as their primary provider ends up with a monkey patch or extra `provider:` keys everywhere.

This only covers chat — you'd need the same for embed, paint, transcribe, and moderate. It also bypasses the model resolution layer, so alias handling and bedrock region inference don't work correctly.

### Proposed solution
`default_providers`: Route models to specific providers based on model ID prefix. 
- More specific keys take precedence over shorter ones.
- Only applies when no explicit provider: argument is passed.

```ruby
RubyLLM.configure do |config|
  config.default_providers = {
    'claude' => :bedrock,                # all claude models route to bedrock
    'claude-haiku' => :anthropic,        # except haiku, which goes direct to anthropic
    'gpt' => :azure,                     # all gpt models route to azure
    'gemini' => :vertexai,               # all gemini models route to vertex ai
    'claude-sonnet-4-5' => :openrouter,  # one specific model via openrouter
  }
end
```

Specifying provider every time is a waste and unnecessary.

Right now I do this with a Monkey Patch because we always use bedrock for claude, would love to get rid of it!

```
module RubyLLMBedrockDefaultPatch
  def chat(model:, provider: nil, **options)
    provider ||= :bedrock if model.to_s.match?(/^claude-/)
    super
  end
end

# Prepend to RubyLLM module to default Claude models to Bedrock
RubyLLM.singleton_class.prepend(RubyLLMBedrockDefaultPatch)
```

llenodo:
## What this does

Screencast overview: https://www.loom.com/share/d044887130304d64b08eeab0eeee6f82

Adds a `default_providers` configuration option that routes models to specific providers by model ID prefix, eliminating the need to pass `provider:` on every call.

```ruby
RubyLLM.configure do |config|
  config.default_providers = {
    'claude' => :bedrock,            # all claude models route to bedrock
    'claude-haiku' => :anthropic,    # except haiku, which goes direct to anthropic
    'gpt' => :azure,                 # all gpt models route to azure
    'gemini' => :vertexai,           # all gemini models route to vertex ai
    'claude-sonnet-4-5' => :openrouter,  # one specific model via openrouter
  }
end
```

Longer (more specific) keys take precedence. Explicit `provider:` arguments always override. Hooks into `Models.resolve`, so it applies to all model types (chat, embedding, image, transcription, moderation).

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Required for new features

- [x] Linked issue: #686

## Quality check

## AI-generated code

## API changes

- [ ] Breaking change
- [ ] No API changes


crmne:
I’m not comfortable with this because it’s effectively an escape hatch around provider resolution.

Today, omitting `provider:` means RubyLLM resolves via aliases, the registry, and provider preference. This changes that to “force a provider if a global prefix rule matches,” which feels like app-specific routing policy leaking into the library.

If an app wants Claude to always go through Bedrock, I think it should pass or persist `provider: :bedrock` locally, not change RubyLLM’s resolution semantics.

PR #689, opened 2026-03-19
Fix ActiveRecord with_tool/with_tools resetting messages mid-conversation

ebeigarts:
When a tool dynamically registers new tools during execution (e.g., a ToolSearch tool that calls `chat.with_tool` to make discovered tools available), the ActiveRecord `with_tool`/`with_tools` methods were calling `to_llm` which reset all messages from the database. This inserted an extra empty assistant message between the tool_calls message and its tool result, violating the OpenAI API contract and causing a `RubyLLM::BadRequestError`.

```
An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages: call_bcf3b9FOG3G3wxc2OW7Zg3CI (RubyLLM::BadRequestError)
```

The solution is to add a `reset_messages` parameter to `to_llm` so that `with_tool`/`with_tools` can initialize the chat object without disrupting the in-flight message sequence.

To reproduce the issue: https://gist.github.com/ebeigarts/dae372091c42756c3bdf1fa9e2d5f2ef

### DIFF
--- a/lib/ruby_llm/active_record/chat_methods.rb
+++ b/lib/ruby_llm/active_record/chat_methods.rb
@@ -54,19 +54,23 @@ module RubyLLM
         model_association&.provider
       end
 
-      def to_llm
+      def to_llm(reset_messages: true)
         model_record = model_association
         @chat ||= (context || RubyLLM).chat(
           model: model_record.model_id,
           provider: model_record.provider.to_sym,
           assume_model_exists: assume_model_exists || false
         )
-        @chat.reset_messages!
 
-        ordered_messages = order_messages_for_llm(messages_association.to_a)
-        ordered_messages.each do |msg|
-          @chat.add_message(msg.to_llm)
+        if reset_messages
+          @chat.reset_messages!
+
+          ordered_messages = order_messages_for_llm(messages_association.to_a)
+          ordered_messages.each do |msg|
+            @chat.add_message(msg.to_llm)
+          end
         end
+
         reapply_runtime_instructions(@chat)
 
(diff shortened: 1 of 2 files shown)


lib/ruby_llm/active_record/chat_methods.rb
```diff
@@ -110,12 +114,12 @@ def with_runtime_instructions(instructions, append: false, replace: nil)
...
-        to_llm.with_tool(...)
+        to_llm(reset_messages: false).with_tool(...)
         self
       end
 
       def with_tools(...)
-        to_llm.with_tools(...)
+        to_llm(reset_messages: false).with_tools(...)
         self
       end
```
ebeigarts, line 103: Maybe it would be better to just use this instead?

```
  def with_tool(...)
    (@chat || to_llm).with_tool(...)
    self
  end

  def with_tools(...)
    (@chat || to_llm).with_tools(...)
    self
  end
```

jayelkaake:
Ran into this issue as well.

crmne:
Thanks for the original report and regression case here.

I reproduced this on current main and landed a smaller rebased fix directly on main in 2c446372. The fix keeps ActiveRecord with_tool/with_tools from rehydrating the chat while a tool call is in flight by reusing the live chat when one already exists.

That preserves the intended message order for dynamic tool registration during tool execution: user, assistant, tool instead of user, assistant, assistant, tool.

Closing this PR as superseded by 2c446372. Thanks again for tracking this down.

PR #694, opened 2026-03-22
Fix parent_tool_call foreign key inference

Gujiassh:
## Summary
- derive `parent_tool_call` from the tool-call class by default, but honor custom namespaced association prefixes when `tool_calls_foreign_key` is explicitly provided
- keep the fix scoped to the new `acts_as` API so it aligns with the already-correct legacy behavior without touching generators or docs
- add focused namespaced ActiveRecord regressions covering both the reflected foreign key and real `parent_tool_call` persistence with a custom `tool_calls` association name

## Validation
- `bundle exec rspec spec/ruby_llm/active_record/acts_as_spec.rb --example 'uses the tool_call_class-derived foreign key for parent_tool_call' --example 'persists parent_tool_call with a custom tool_calls association name'`
- `bundle exec rspec spec/ruby_llm/active_record/acts_as_spec.rb`
- `bundle exec rubocop lib/ruby_llm/active_record/acts_as.rb spec/ruby_llm/active_record/acts_as_spec.rb`

### DIFF
--- a/lib/ruby_llm/active_record/acts_as.rb
+++ b/lib/ruby_llm/active_record/acts_as.rb
@@ -106,6 +106,12 @@ module RubyLLM
           self.tool_call_class = (tool_call_class || tool_calls.to_s.classify).to_s
           self.model_class = (model_class || model.to_s.classify).to_s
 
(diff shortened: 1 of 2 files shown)


Gujiassh:

PR #698, opened 2026-03-22
Add Chat#step single-iteration execution

Gujiassh:
## Summary
- add `Chat#step` to run exactly one provider iteration plus any resulting tool execution without recursing through the full agent loop
- refactor `complete` to delegate to `step` so existing behavior stays unchanged while `Tool::Halt` semantics are preserved
- add focused chat tool regressions proving `step` executes one tool-calling turn and returns `Tool::Halt` immediately when a halting tool is invoked

## Validation
- `bundle exec rspec spec/ruby_llm/chat_tools_spec.rb --example 'step executes a single tool-calling iteration without recursing' --example 'step returns Halt when a tool halts'`
- `bundle exec rspec spec/ruby_llm/chat_tools_spec.rb`
- `bundle exec rubocop lib/ruby_llm/chat.rb spec/ruby_llm/chat_tools_spec.rb`

### DIFF
--- a/lib/ruby_llm/active_record/chat_methods.rb
+++ b/lib/ruby_llm/active_record/chat_methods.rb
@@ -227,6 +227,14 @@ module RubyLLM
         raise e
       end
 
+      def step(...)
+        to_llm.step(...)
+      rescue RubyLLM::Error => e
+        cleanup_failed_messages if @message&.persisted? && @message.content.blank?
+        cleanup_orphaned_tool_results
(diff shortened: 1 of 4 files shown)


Gujiassh:

abuisman:
This would be very nice to add. Maybe you should open it again.

This would allow us to work with rubyllm more as an adapter for various llm providers rather than having to port all of our code.

PR #700, opened 2026-03-23
Fix Attachment#save for binary IO

Gujiassh:
## Summary
- make `Attachment#save` rewind IO sources and write them in binary mode
- add regression coverage to ensure binary payloads are saved byte-for-byte without an added newline

## Testing
- `bundle exec rspec spec/ruby_llm/attachment_spec.rb`
- `bundle exec rubocop lib/ruby_llm/attachment.rb spec/ruby_llm/attachment_spec.rb`


Gujiassh:

PR #711, opened 2026-03-31
Fix streaming parallel tool calls concatenating JSON arguments

### LINKED ISSUE #710: [BUG] Streaming parallel tool calls: JSON fragments concatenated across different tool calls
### Basic checks

- [x] I searched existing issues - this has not been reported

### What is broken?

When Anthropic returns **multiple tool calls in a single streaming response**, the JSON argument fragments from different tool calls are concatenated into one invalid string, causing `JSON::ParserError` in `StreamAccumulator#tool_calls_from_stream` (called during `to_message`).

The root cause is in `Providers::Anthropic::Tools#extract_tool_calls` — `input_json_delta` events are returned with `id: nil`:

```ruby
if json_delta?(data)
  { nil => ToolCall.new(id: nil, name: nil, arguments: data.dig("delta", "partial_json")) }
```

Then `StreamAccumulator#accumulate_tool_calls` routes all `id: nil` fragments to `@latest_tool_call_id`:

```ruby
else
  existing = @tool_calls[@latest_tool_call_id]
  existing.arguments << fragment
```

When tool call A and tool call B stream with interleaved deltas, B fragments get appended to A argument string, producing:

```
{"symbol":"MNQM26","from":"2026-03-31","interval":"minute"}{"query":"market news March 31 2026"}
```

`JSON.parse` then raises `unexpected characters after the JSON document`.

### How to reproduce

The bug triggers whenever Anthropic emits multiple interleaved tool calls in a single streamed turn. One easy repro path:

1. Use Anthropic Claude with streaming enabled
2. Register two or more tools (e.g., a local tool + Anthropic server tool like `web_search`, or just multiple local tools)
3. Send a prompt that triggers parallel tool calls (e.g., "Look up my recent trades and search the web for market news about the price spike at 7:11 AM")
4. The streaming response includes interleaved `content_block_start` + `input_json_delta` events for both tool calls
5. `tool_calls_from_stream` raises `JSON::ParserError`

The bug is about multiple interleaved Anthropic tool calls in general — mixing local and server tools just makes it easy to trigger.

### Expected behavior

Each tool call JSON fragments should be accumulated separately, producing two valid JSON argument objects.

### Root cause and fix

Anthropic streaming API includes an `index` field on every `content_block_delta` event that identifies which content block the fragment belongs to:

```json
{"type":"content_block_delta","index":1,"delta":{"type":"input_json_delta","partial_json":"{\"symbol\":"}}
{"type":"content_block_delta","index":2,"delta":{"type":"input_json_delta","partial_json":"{\"query\":"}}
```

The fix: pass the `index` through `extract_tool_calls` and use it in `accumulate_tool_calls` to map each fragment to the correct tool call (via a `block_index -> tool_call_id` mapping built from `content_block_start` events).

### Current workaround

We are using `calls: :one` (`disable_parallel_tool_use: true`) to prevent Anthropic from emitting parallel tool calls, which avoids the bug entirely at the cost of sequential tool execution.

### Environment

- ruby_llm version: 1.13.2
- Ruby version: 3.4.7
- Provider: Anthropic (Claude Sonnet 4)
- Streaming: yes
- Tools: multiple

alexandrigo:
## What this does

Fixes a bug where parallel streaming tool calls from Anthropic get their JSON argument fragments concatenated into one invalid string, causing `JSON::ParserError` in `StreamAccumulator#tool_calls_from_stream`.

Fixes #710

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

- [x] This aligns with RubyLLM focus on LLM communication
- [x] This is not application-specific logic that belongs in user code

## Root cause

`Providers::Anthropic::Tools#extract_tool_calls` returns `input_json_delta` fragments with `id: nil`. `StreamAccumulator#accumulate_tool_calls` routes all `id: nil` fragments to `@latest_tool_call_id`. When two tool calls stream with interleaved deltas, fragments from tool B get appended to tool A arguments:

```
{"symbol":"MNQM26","interval":"minute"}{"query":"market news"}
```

## Fix

Anthropic streaming API includes an `index` field on every `content_block_delta` identifying which content block a fragment belongs to.

1. **`extract_tool_calls`**: Pass block index as the hash key for `input_json_delta` chunks. Also emit a registration entry from `content_block_start` events that maps block index to tool call ID.
2. **`accumulate_tool_calls`**: Build a `block_index -> tool_call_id` mapping from registration entries. Route delta fragments by block index instead of `@latest_tool_call_id`.
3. **Backward compatible**: Falls back to `@latest_tool_call_id` when no block index is present (non-Anthropic providers).

## Tests

- `keeps parallel tool call arguments separate when block index is provided` — simulates interleaved deltas from two tool calls, verifies each gets its own parsed arguments
- `falls back to latest_tool_call_id when no block index is provided` — verifies existing behavior for providers that do not send block index
- Existing `handles tool call deltas that omit arguments` still passes


alexandrigo:
(NOT ACTUAL)
CI failure is pre-existing — all 6 rubocop offenses are in `spec/ruby_llm/generators/chat_ui_generator_spec.rb` (Style/RedundantPercentQ), which this PR does not touch. Our 3 changed files pass rubocop cleanly.

crmne:
Superseded by https://github.com/crmne/ruby_llm/commit/ff19c933e092a9dc75224da273d256fb301b2db4

PR #715, opened 2026-04-01
Add Apple Intelligence provider — fully local on-device AI for macOS

parolkar:
## What

Adds a new provider that brings Apple Intelligence (Apple Foundation Models) to RubyLLM — completely local, on-device inference on macOS 26+ with Apple Silicon. No API keys, no cloud, no data leaves the machine.

## Why

Every other RubyLLM provider requires a cloud API. macOS 26 ships a capable on-device LLM through Apple Intelligence, but there is no way to use it from Ruby today. This provider bridges that gap as a drop-in addition alongside existing providers.

## How it works

The provider pipes JSON to the [`osx-ai-inloop`](https://github.com/inloopstudio-team/apple-intelligence-inloop) binary via `Open3.capture3` (stdin/stdout), completely bypassing HTTP/Faraday. The binary communicates with Apple's Foundation Models framework on-device.

- **Auto-downloads** the pre-built binary on first use (`~/.ruby_llm/bin/osx-ai-inloop`)
- **Tool calling** via two-pass argument extraction — the model extracts parameter values, Ruby constructs the tool call
- **Zero configuration** — no API keys, optional custom binary path

## Files (605 lines added)

| File | Purpose |
|------|---------|
| `providers/apple_intelligence.rb` | Provider class, `complete()` override, tool call orchestration |
| `providers/apple_intelligence/chat.rb` | Payload building, binary execution, response parsing, tool arg extraction |
| `providers/apple_intelligence/binary_manager.rb` | Binary download, caching, platform validation |
| `providers/apple_intelligence/models.rb` | `apple-intelligence` model registration |
| `providers/apple_intelligence/capabilities.rb` | Capability declarations |
| `docs/_getting_started/apple-intelligence.md` | Full guide: setup, usage, troubleshooting |
| Updated: `README.md`, `configuration.md`, `overview.md`, `lib/ruby_llm.rb` | Provider registration and docs |

## Usage

```ruby
# No API key needed
chat = RubyLLM.chat(model: "apple-intelligence", provider: :apple_intelligence)
chat.ask "Explain Ruby blocks"

# Tool calling works too
chat.with_tool(Weather).ask "What's the weather in Tokyo?"
```

## Requirements

- macOS 26+ (Tahoe), Apple Silicon, Apple Intelligence enabled
- Flags: `local? => true`, `assume_models_exist? => true`
- Text-only (no streaming, no vision, no embeddings)


parolkar:
I have fixed the errors that caused checks to fail.

The remaining 6 offenses in the CI output are in spec/ruby_llm/generators/chat_ui_generator_spec.rb —  are pre-existing in the  repo, not from my changes.

cc @crmne

parolkar:
@crmne : Any feedback?

parolkar:
Hi @crmne — pushed `107fae23` which addresses the RuboCop CI failure:

- **Removed redundant `Metrics/PerceivedComplexity` disable** on `apple_intelligence.rb#complete` — the method is no longer over threshold after extracting the `try_tool_call` helper
- **Removed redundant `Metrics/PerceivedComplexity` disable** on `chat.rb#resolve_tool_call` — also under threshold now thanks to extracted helpers
- **Fixed `Lint/NoReturnInBeginEndBlocks`** — replaced a `begin..end` assignment block containing `return nil` with a small `parse_json_safely` helper method

Also merged the latest changes from upstream `main` into this branch — no conflicts.

✅ RuboCop and gitleaks now pass. Test matrix is running.

Would love to get your thoughts on the overall approach when you have time — happy to make any changes needed.

lib/ruby_llm/providers/apple_intelligence/chat.rb
```diff
@@ -0,0 +1,221 @@
...
+          json_input = JSON.generate(payload)
+
+          stdout, stderr, status = Open3.capture3(bin, stdin_data: json_input)
+
+          handle_exit_code(status, stdout, stderr)
+          parse_binary_response(stdout)
+        end
+
+        def resolve_tool_call(tools, user_message, config)
+          return nil unless tools&.any?
```
fagnerpereira, line 90: All those explicit `return nil` statements are unnecessary; it’s fine to return a value when you want to, but when you just use `return` to exit the method early, Ruby already returns `nil` by default.

crmne:
Thanks for the PR, but I don’t think this belongs in RubyLLM core.

A provider that downloads, chmods, and executes a third-party binary at runtime changes the trust model of the gem in a way users won’t reasonably expect from selecting a provider. RubyLLM should not take responsibility for distributing and running an opaque native helper as a side effect of normal library use.

If this integration is useful, the right shape is a separate extension gem where users explicitly opt into installing and trusting the binary. The provider docs cover that path here: https://rubyllm.com/next/custom-providers/

parolkar:
Hi @crmne , Are you suggesting this to be companion gem, then?

PR #733, opened 2026-04-16
Handle nil response bodies consistently across providers

fidalgo:
Faraday JSON middleware can normalize empty HTTP bodies to nil, but ruby_llm still assumed response.body was always non-nil in several sync completion and error parsing paths. That leaked as NoMethodError from provider parsers instead of raising a RubyLLM::Error with a useful message.

Centralize the sync completion invariant in Provider by adding ensure_response_body! and calling it from Provider#sync_response. Apply the same guard to Bedrock's custom signed sync path so override implementations follow the same contract.

Harden provider-specific parsers that were still nil-fragile:
- OpenAI::Chat#parse_completion_response
- OpenRouter::Chat#parse_completion_response
- Anthropic::Chat#parse_completion_response
- Gemini::Chat#parse_completion_response
- Bedrock::Chat#parse_completion_response
- Provider#parse_error
- OpenRouter#parse_error
- Perplexity#parse_error

Standardize nil sync completion bodies to raise:
  RubyLLM::Error, "Provider returned an empty response body"

Add regression coverage for:
- Provider#sync_response nil-body handling
- Provider/OpenRouter/Perplexity parse_error nil-body handling
- OpenAI/OpenRouter/Anthropic/Gemini/Bedrock chat parser nil-body handling

This keeps empty-body failures inside the gem's error model and makes nil-body behavior consistent across the provider stack.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

- [x] I read the [Contributing Guide](https://github.com/crmne/ruby_llm/blob/main/CONTRIBUTING.md)
- [x] This aligns with RubyLLM's focus on **LLM communication**
- [x] This isn't application-specific logic that belongs in user code
- [x] This benefits most users, not just my specific use case

## Required for new features

- [ ] I opened an issue **before** writing code and received maintainer approval
- [ ] Linked issue: #___

**PRs for new features or enhancements without a prior approved issue will be closed.**

## Quality check

- [x] I ran `overcommit --install` and all hooks pass
- [ ] I tested my changes thoroughly
  - [ ] For provider changes: Re-recorded VCR cassettes with `bundle exec rake vcr:record[provider_name]`
  - [x] All tests pass: `bundle exec rspec`
- [ ] I updated documentation if needed
- [x] I didn't modify auto-generated files manually (`models.json`, `aliases.json`)

## AI-generated code

- [x] I used AI tools to help write this code
[...]

PR #734, opened 2026-04-16
Add Content#empty? to fix Anthropic NoMethodError

### LINKED ISSUE #729: [BUG] RubyLLM::Content missing #empty? method causes NoMethodError in Anthropic provider
## Description

`RubyLLM::Content` does not implement `#empty?`, but several Anthropic provider code paths call `msg.content.empty?` on message content. When a message has ActiveStorage attachments, `extract_content` (in `active_record/message_methods.rb`) wraps the content in a `Content` object instead of returning a plain String. This causes a `NoMethodError` when the message is later formatted for the Anthropic API.

## Steps to reproduce

1. Create a message with ActiveStorage attachments (e.g., via `has_many_attached :attachments` on the message model)
2. Have the agent respond with tool calls in the same conversation
3. On the next turn, `extract_content` returns a `RubyLLM::Content` for the user message with attachments
4. The Anthropic provider calls `msg.content.empty?` during `format_tool_call_with_thinking` or `format_tool_result`
5. **NoMethodError: undefined method 'empty?' for an instance of RubyLLM::Content**

## Affected code paths

- `lib/ruby_llm/providers/anthropic/chat.rb:179` — `msg.content.nil? || msg.content.empty?`
- `lib/ruby_llm/providers/anthropic/tools.rb:19` — `msg.content.nil? || msg.content.empty?`

## Expected behavior

`RubyLLM::Content` should respond to `#empty?` (returning true when text is nil and attachments are empty), consistent with how String and Array respond to it.

## Workaround

```ruby
RubyLLM::Content.class_eval do
  def empty?
    text.nil? && attachments.empty?
  end
end
```

## Environment

- ruby_llm 1.14.0
- Rails 8.1 with ActiveStorage
- Anthropic provider (Claude models)

55728:
## What this does

Fixes #729

### Summary
Adds `RubyLLM::Content#empty?` so `msg.content.nil? || msg.content.empty?` no longer raises `NoMethodError` when `msg.content` is a `Content` instance.

### Why `@text.to_s.strip.empty? && @attachments.empty?`
Three shapes were considered:

- `def empty?; false; end`, leaning on the constructor check `raise ... if @text.nil? && @attachments.empty?`. That check only rejects a `nil` text, though — `Content.new('')` and `Content.new('   ')` both construct successfully, so a blanket `false` would misrepresent those instances.
- Updating the two call sites in `anthropic/chat.rb:179` and `anthropic/tools.rb:19` to use the defensive `respond_to?(:empty?) && empty?` idiom already present in `anthropic/tools.rb:49`, `bedrock/media.rb:20`, and `gemini/tools.rb:19`. Workable, but leaves `Content` as the only nearby domain object without `#empty?`, which is a bit surprising for a container-like Ruby type.
- **Chosen:** give `Content` an `#empty?` with `String#empty?` / `Array#empty?` semantics — blank when there is no visible text and no attachments. The method name alone predicts the behavior, and the rule holds even if the constructor invariant is relaxed later.

### Side effect (bonus)
`bedrock/media.rb:20`, `gemini/tools.rb:19`, `gemini/tools.rb:50`, and `anthropic/tools.rb:49` already use the defensive `content.respond_to?(:empty?) && content.empty?` idiom. Until now `Content` did not respond to `empty?`, so those guards silently treated every `Content` instance as non-empty. With `#empty?` defined, those call sites behave as their authors intended for blank content.

### Out of scope (happy to fold in if preferred)
`anthropic/chat.rb:179` and `anthropic/tools.rb:19` still use bare `msg.content.empty?`. Commit 4371a1b introduced the defensive `respond_to?(:empty?)` pattern in `anthropic/tools.rb:49`; aligning these two lines with that convention would be a small follow-up — happy to handle it in a separate PR, or fold it into this one if that fits the repo style better.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

  - [ ] For provider changes: Re-recorded VCR cassettes — n/a, no provider code changed
  - [x] New spec `spec/ruby_llm/content_spec.rb` passes (5 examples, 0 failures)
- [ ] I updated documentation if needed — n/a, no public-API docs to update beyond the new method itself

## AI-generated code

- [x] I have reviewed and understand all generated code

## API changes

- [ ] Breaking change
- [x] New public methods/classes (`RubyLLM::Content#empty?`)
- [ ] No API changes


fidalgo:
I've opened a PR that aims to fix this on all providers and raising a meaningful error.
Have a look and let me know your thoughts: https://github.com/crmne/ruby_llm/pull/733

55728:
@fidalgo

Thanks for flagging this — I had a look at #733 and it’s a solid approach for the HTTP layer.
If I may, though, I wonder whether these two PRs might actually be complementary rather than overlapping? #733 guards against nil response bodies from Faraday, which is an important safety net at the transport level. This PR (#734), on the other hand, addresses Content as a domain object — it simply lacks #empty?, which means existing call sites like msg.content.empty? raise NoMethodError regardless of whether the response body was nil or not.

There’s also the matter of the existing respond_to?(:empty?) guards scattered across providers (bedrock/media.rb, gemini/tools.rb, anthropic/tools.rb) — without Content#empty?, those silently treat every Content instance as non-empty, which isn’t quite what their authors intended.

I’d be delighted if both could land — one hardens the transport, the other completes the domain model. But of course, happy to hear your thoughts.

fidalgo:
@55728 That's a broader decision: is an empty response an error or an acceptable case. My gut feeling tells me, we should have some content on the response, but I do not know every provider and that may pose an API change.

Edit: there's another PR open on the same: https://github.com/crmne/ruby_llm/pull/747

Flagging this to @crmne (sorry to flag you, but if we can have a decision here both PR's can be addressed and the issue fixed)

55728:
Good call flagging @crmne — thanks for that.

Agreed that the empty-response semantics are a broader design question. For what it’s worth, I believe this PR is independent of that decision: Content#empty? is about giving the domain object a method that Ruby containers conventionally have, and should be safe to land either way.

Happy to wait for a steer on the bigger picture.

55728:
Closing this as stale.

Thanks @fidalgo for the thoughtful discussion!

PR #745, opened 2026-04-24
Translate Anthropic tool search feature (defer_loading + native BM25)

swistaczek:
Hey Carmine — huge fan of ruby_llm. Thank you for all the work you put into it.

## A heads-up, not a demand

I'm keeping this as a **draft** and am aware `CONTRIBUTING.md` asks for an issue-first discussion for features. I went ahead and built a working prototype because I thought a concrete artifact would be easier to evaluate. **Please treat this as a proposal, not a request for merge** — happy to close it, rework it, or move to a fresh issue first. No pressure.

## The use case (real-world, not hypothetical)

I'm building an app that plugs many MCP servers into a single chat. Past ~40 tools the system-prompt prefix balloons and prompt caching breaks when the tool set shifts. Anthropic shipped a first-party fix in late 2025 — `defer_loading: true` per tool plus the server-side `tool_search_tool_bm25_20251119` primitive — that keeps deferred schemas out of the cached prefix. It works beautifully in Claude Code.

Since ruby_llm already translates provider features like prompt caching, thinking, and tool use one-to-one, this felt like the same category of work.

## Scope — narrowed to just the Anthropic translation

Per your rejection of the earlier full-scope version, I've trimmed this PR to exactly the Anthropic feature translation and nothing else. The client-side emulation pieces (pure-Ruby BM25, `Tools::SearchTools` function tool, OpenAI/Gemini/Bedrock stub emission, custom-ranker hook) are preserved locally on a separate branch and can ship as a follow-up PR or a standalone gem once this lands — whichever you prefer.

**Opt-in, non-breaking.** If you don't use `defer:` or `deferred`, nothing changes.

```ruby
# Class-level DSL (intrinsically heavy tools)
class DeepResearchTool < RubyLLM::Tool
  description 'Runs a multi-step research task'
  deferred
  param :query, desc: '...'
  def execute(query:); ...; end
end

# Per-registration (natural for MCP bulk registration)
chat.with_tools(*mcp_client.tools, defer: true)

# Observe promotions (Claude loaded these server-side)
chat.on_tool_search { |event| logger.info("loaded: #{event.results}") }
```

## What this PR does

- Forwards `defer_loading: true` per deferred tool and appends `tool_search_tool_bm25_20251119` to the payload.
- Parses `tool_search_tool_result` blocks in the response and promotes referenced tools from `chat.tool_catalog` into `chat.tools` so the normal dispatch path can call them on the next turn.
- Exposes `chat.on_tool_search` for observability, `chat.tool_catalog` for inspection.
- On non-Anthropic providers `defer: true` is silently coerced to regular registration with a one-time warning; same behavior when `RubyLLM.config.tool_search_enabled = false`.

## Public API (trimmed per maintainer feedback)

- `RubyLLM::Tool.deferred` class DSL + `.deferred?` / `#deferred?`
- `RubyLLM::Chat#with_tool(..., defer:)` / `#with_tools(..., defer:)` kwarg
- `RubyLLM::Chat#on_tool_search`
- `RubyLLM::Chat#tool_catalog` reader
- `RubyLLM::Tool::SearchEvent` (callback payload)
- `RubyLLM::ToolCatalog`
- `RubyLLM::Message#tool_references`
- `RubyLLM.config.tool_search_enabled` (default `true`)

That's 7 new surfaces (down from ~14 in the earlier scope). `promote_from_tool_references` is private (called only from `Chat#complete` after parsing the response); `Tool::Registration` is tagged `@api private`.

## Tests

- 2 VCR cassettes against Claude (2 requests and 1 request — no retry storm).
- Deterministic 40+ tool scale spec (payload shape, native primitive emission, catalog promotion).
- Unit specs for `ToolCatalog`, `Message#tool_references`, the Anthropic parser's `tool_search_tool_result` handling, the kill-switch and unsupported-provider warning paths.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

- [x] I read the Contributing Guide
- [x] Translates a named Anthropic API feature one-to-one (same category as prompt caching, thinking, tool use)
- [x] Not application-specific logic
- [x] Benefits every user with a large tool catalog, especially MCP-heavy setups

## Quality check

- [x] `bundle exec rubocop` — 260 files, 0 offenses
- [x] `bundle exec rspec` — all affected specs green (104 examples in the tool-search surface, 0 failures)
- [x] VCR cassettes re-recorded against live Anthropic API, API keys scrubbed
- [x] Documentation added (`docs/_core_features/tool-search.md` + upgrade-guide entry + cross-link from tools.md)
- [x] Did not modify auto-generated files

## AI-generated code

- [x] I have reviewed and understand all generated code

Happy to iterate further or split/drop as you see fit.

### DIFF
--- a/docs/_advanced/upgrading.md
+++ b/docs/_advanced/upgrading.md
@@ -28,9 +28,11 @@ This guide focuses on upgrade-impacting changes: migrations, token semantics, de
 
 ## How to Upgrade
 
-No generator is required for the token and cost API changes in 1.15.
+No generator is required for the token, cost, and tool-search API changes in 1.15.
 
 If you use the Rails integration and already ran the v1.9 migration, no new columns are needed. The new `cache_read_tokens` and `cache_write_tokens` helpers use the existing `cached_tokens` and `cache_creation_tokens` columns.
+
+Tool search is fully additive — if you don't use `defer:` or `deferred`, nothing changes.
 
 ## Token Semantics Changed
 
@@ -78,6 +80,10 @@ Cost helpers are available from 1.15 onward. They return `nil` for any cost buck
 `tokens.thinking` remains available from 1.10. From 1.15 onward, `tokens.output` is normalized as the billable output bucket. Do not add `tokens.thinking` to `tokens.output` yourself; RubyLLM includes thinking in output when the provider bills it as output, and exposes `cost.thinking` only for models with distinct reasoning-token pricing.
 
 See [Tracking Token Usage]({% link _core_features/chat.md %}#tracking-token-usage) for the provider comparison table and the exact normalized token semantics RubyLLM exposes.
+
+## Tool Search (Anthropic)
+
+1.15 adds tool search in a fully additive way. `RubyLLM::Chat#with_tool` / `#with_tools` accept a new `defer:` keyword argument, and `RubyLLM::Tool` exposes a class-level `deferred` DSL. On Anthropic this translates to the native `defer_loading: true` flag plus the `tool_search_tool_bm25_20251119` primitive: deferred tools stay out of the system-prompt prefix and Claude loads the ones it actually needs server-side. On other providers `defer:` is ignored with a one-time warning. If you don't use `defer:` or `deferred`, nothing changes. See [Tool Search]({% link _core_features/tool-search.md %}).
 
 # Upgrade to 1.14
 
--- a/docs/_core_features/tool-search.md
+++ b/docs/_core_features/tool-search.md
@@ -1,0 +1,113 @@
+---
+layout: default
+title: Tool Search
+nav_order: 9
+description: Keep large tool catalogs out of Claude's prompt prefix. Mark tools as deferred and let Anthropic's server-side tool-search primitive load them on demand.
+redirect_from:
+  - /guides/tool-search
+---
+
+# {{ page.title }}
+{: .d-inline-block .no_toc }
+
+New in 1.15
+{: .label .label-green }
+
+{{ page.description }}
+{: .fs-6 .fw-300 }
+
+## Table of contents
+{: .no_toc .text-delta }
+
+1. TOC
+{:toc}
+
+---
+
+After reading this guide, you will know:
+
+*   When deferred tool loading helps.
+*   How to mark tools as deferred.
+*   How Anthropic loads deferred tools at runtime.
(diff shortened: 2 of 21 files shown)


swistaczek:
@crmne — rebased on fresh `main` (now sitting on top of `5bdda1a5`) and aligned with conventions that landed since the PR was opened:

- **Additive callbacks** — `on_tool_search` now routes through `set_legacy_callback` (matching `on_new_message`/`on_tool_call`/etc.), and a new `after_tool_search` is the additive primary form. Dispatch in `promote_from_tool_references` goes through `run_callbacks(:after_tool_search, :tool_search, event)`. Docs and specs updated to use `after_tool_search`.
- **Provider feature check** — added `supports_deferred_loading? = false` default on `Provider` (parallel to `local?`/`remote?`/`assume_models_exist?`) so the `chat.rb` defer gate is `@provider.supports_deferred_loading?` instead of a `respond_to?` probe.
- Explicit `require 'set'` in `chat.rb` (Ruby 3.1 target).
- Conflict resolutions in `chat.rb` and `docs/_advanced/upgrading.md` keep both 1.15 features (token/cost normalization + tool search) coherent under a single upgrade section.

Rubocop clean, 45 tool-search-related specs green locally. Curious what you think.

crmne:
Thank you for working through this. I'm going to close this PR, because this is exactly the kind of feature that needs an issue/design discussion first.

The problem is real, but I don't like the API shape here. It's too Anthropic-specific, and it adds a lot of public surface (`deferred`, `defer:`, catalog, callbacks, config) before we've agreed on the abstraction.

Looking at this more, OpenAI has native tool search now too, and Bedrock exposes Claude's version through InvokeModel, though not Converse. So I think this wants to be a RubyLLM-level tool search/tool loading feature with provider adapters underneath, not an Anthropic-only API.

Could you please open an issue for the design? I'll add my thoughts there and we can work out the right shape before code.

benjaminwood:
An issue for this [exist](https://github.com/crmne/ruby_llm/issues/748) and has a proposed solution but no discussion yet.

I'm going to take a stab at this myself, but will not open a PR until you've given some direction there.

_Edit: Actually, that issue is anthropic specific so I've opened a new issue [here](https://github.com/crmne/ruby_llm/issues/839) for the generalized solution_

PR #749, opened 2026-04-25
Fix deepseek-reasoner multi-turn requests missing reasoning_content

iuhoay:
## Summary

DeepSeek's thinking-capable models (`deepseek-reasoner`, the newer `deepseek-v4-flash` / `deepseek-v4-pro`, and presumably anything else DeepSeek ships with thinking mode) reject multi-turn requests when any prior assistant message is missing a `reasoning_content` key:

> The `reasoning_content` in the thinking mode must be passed back to the API.

The inherited OpenAI `format_thinking` (`lib/ruby_llm/providers/openai/chat.rb:111-124`) returns `{}` when `msg.thinking` is nil and omits both `reasoning` / `reasoning_content` when `thinking.text` is empty. Any assistant turn that emitted only tool calls — or that simply didn't capture thinking text — therefore breaks the next request.

This PR overrides `format_thinking` on `RubyLLM::Providers::DeepSeek::Chat` to always emit `reasoning_content` on assistant messages (empty string when no thinking text was captured), preserving the existing `reasoning` / `reasoning_signature` semantics for everything else.

The override is provider-level (not gated on model id), so it applies to every DeepSeek model. For non-thinking models like `deepseek-chat` the API ignores the empty key.

## Test plan

- [x] New spec at `spec/ruby_llm/providers/deep_seek/chat_spec.rb` covers: thinking text + signature present, no thinking captured (the regression case), and non-assistant messages
- [x] `bundle exec rspec spec/ruby_llm/providers/deep_seek/chat_spec.rb` — 3 examples, 0 failures
- [x] `bundle exec rubocop lib/ruby_llm/providers/deepseek/chat.rb spec/ruby_llm/providers/deep_seek/chat_spec.rb` — no offenses
- [x] Reviewer to confirm the existing `deepseek-chat` VCR-backed flows still pass (cassettes were recorded before this change; the empty `reasoning_content` should be ignored, but worth a re-record if the live API tightens validation)


ashishbista:
+1

PR #761, opened 2026-05-07
Add hosted OpenAI-compatible gateway example

haocn-ops:
This adds a small hosted OpenAI-compatible gateway example to the existing custom endpoint section.

API NODE uses the same `openai_api_base` path as other OpenAI-compatible endpoints, so the example is meant to make that hosted-gateway setup easier to recognize without changing the RubyLLM API surface.

Docs-only change.


crmne:
This looks like promotion.

Issue #785, opened 2026-05-29
OpenAI Responses API: enable reasoning_effort + function tools (gpt-5.x / o-series)


andrew-woblavobla:
## Problem

OpenAI reasoning models (gpt-5, gpt-5.1/5.2/5.5, o1/o3/o4) **reject `reasoning_effort` together with function tools** on `/v1/chat/completions`:

> 400 — Function tools with reasoning_effort are not supported for gpt-5.5 in /v1/chat/completions. Please use /v1/responses instead.

The OpenAI provider hardcodes `completion_url => 'chat/completions'` and `render_payload` always sets `payload[:reasoning_effort]`, so `chat.with_thinking(effort:).with_tools(...)` is impossible for the entire gpt-5 reasoning family. It's API-wide, not RubyLLM-specific (same break tracked in langchain-ai/langchain#35584 and BerriAI/litellm#23156).

### Repro
```ruby
RubyLLM.chat(model: 'gpt-5.5', provider: :openai, assume_model_exists: true)
       .with_tools(SomeTool)
       .with_thinking(effort: :low)
       .ask('...')   # => RubyLLM::BadRequestError ("use /v1/responses instead")
```

## Proposal

Add OpenAI Responses API (`/v1/responses`) support. I have a **working prototype** and would like your steer on shape — two options:

- **(A) Auto-route within the existing OpenAI provider** (what I prototyped): `render_payload` detects `thinking && tools`, sets a flag, and `completion_url` / `parse_completion_response` branch to the Responses request/response shapes. Zero config — `with_thinking + tools` just works; the default chat/completions path is unchanged (gated by the flag).
- **(B) A dedicated `:openai_responses` provider** subclassing `OpenAI`.

### Verified
`gpt-5.5` + `with_thinking(:high)` + a function tool now completes the full tool loop with **88 reasoning tokens** (previously a 400).

### Scope handled in the prototype
Request: `input` items (incl. `function_call` / `function_call_output` round-trip), flat `{type:"function",…}` tools, top-level `reasoning:{effort:}`, `text.format` for structured output, `store:false`. Response: `output[]` → `Message` / `ToolCall` / `Thinking` + usage (incl. `reasoning_tokens`).

### Not yet done (would complete before merge)
- Responses **streaming** (different SSE event set) — currently guarded with a clear error
- `input_image` multimodal
- Reasoning-item round-trip across turns (`include: ["reasoning.encrypted_content"]`)
[...]

PR #786, opened 2026-05-29
feat(openai): route reasoning + tools to the Responses API (refs #785)

andrew-woblavobla:
## What

Transparently routes `with_thinking(effort:) + tools` for OpenAI to `/v1/responses` — the only endpoint that accepts `reasoning` together with function tools for gpt-5.x / o-series. The default `/v1/chat/completions` path is unchanged (gated by `@openai_responses_mode`).

## Why

OpenAI reasoning models 400 on `reasoning_effort` + function tools via chat/completions ("use /v1/responses instead"), so `chat.with_thinking(effort:).with_tools(...)` is impossible for the whole gpt-5 reasoning family. Details + repro in #785.

## How (auto-route within the OpenAI provider)

- `OpenAI#render_payload` sets `@openai_responses_mode = instance_of?(OpenAI) && responses_api?(tools:, thinking:)` (true only when *both* are present) and renders a Responses payload; `completion_url` / `parse_completion_response` branch on it. The `Chat` module's `render_payload` stays pure chat/completions, and the `instance_of?` guard keeps subclasses (Azure/OpenRouter/Mistral/Perplexity/xAI/GPUStack) on chat/completions — they have no `/v1/responses`.
- New `OpenAI::Responses` module: request translation — `input` items (incl. `function_call` / `function_call_output` round-trip), flat `{type:"function",…}` tools, top-level `reasoning:{effort:}`, `text.format` for structured output, `store:false`; response parsing — `output[]` → `Message` / `ToolCall` / `Thinking` + usage (incl. `reasoning_tokens`).
- `stream_response` raises a clear error in responses mode (Responses SSE streaming not implemented yet).

## Verified

- Live: `gpt-5.5` + `with_thinking(:high)` + a function tool completes the tool loop with reasoning tokens (previously a 400).
- 18 keyless unit specs for the request/response translation + routing; the provider suite passes; RuboCop clean.

## Known follow-ups

- Responses **streaming** (different SSE event set)
- `input_image` multimodal
- Reasoning-item round-trip across turns (`include: ["reasoning.encrypted_content"]`)

I'm **open to design changes** — e.g. extracting this into a dedicated `:openai_responses` provider rather than auto-routing within `OpenAI`, or any other shape you'd prefer.

Refs #785.

### DIFF
--- a/lib/ruby_llm/providers/openai.rb
+++ b/lib/ruby_llm/providers/openai.rb
@@ -5,6 +5,7 @@ module RubyLLM
     # OpenAI API integration.
     class OpenAI < Provider
       include OpenAI::Chat
+      include OpenAI::Responses
       include OpenAI::Embeddings
       include OpenAI::Models
       include OpenAI::Moderation
@@ -13,6 +14,18 @@ module RubyLLM
       include OpenAI::Images
       include OpenAI::Media
       include OpenAI::Transcription
+
+      # Streaming over the Responses API uses a different SSE event set than
+      # chat/completions and is not implemented yet. Reasoning+tools turns route
+      # to Responses (see OpenAI::Responses#responses_api?), so guard streaming
+      # there explicitly instead of mis-parsing chat/completions events.
+      def stream_response(connection, payload, additional_headers = {}, &)
+        if @openai_responses_mode
+          raise RubyLLM::Error.new(nil, 'OpenAI Responses streaming not implemented yet; call without a block')
+        end
+
+        super
+      end
 
       def api_base
         @config.openai_api_base || 'https://api.openai.com/v1'
--- a/lib/ruby_llm/providers/openai.rb
+++ b/lib/ruby_llm/providers/openai.rb
@@ -15,6 +15,25 @@ module RubyLLM
       include OpenAI::Media
       include OpenAI::Transcription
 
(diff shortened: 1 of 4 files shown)


reeganviljoen:
Hi does anyone know how close this is to being landed as this is an issue that breaks this libraries amazing interface with openai models unless it is fixed

zavan:
@reeganviljoen My guess is it won't be merged until streaming is supported (in my case that is essential).

crmne:
Responses API requires us to make a new layer in RubyLLM's architecture: protocols. It will be done that way.

https://github.com/crmne/ruby_llm/issues/213#issuecomment-4025483596

crmne:
@zavan streaming is supported since 1.0

zavan:
@crmne I was talking specifically about streaming support in this pull request:

Known follow-ups
- Responses streaming (different SSE event set)

reeganviljoen:
@crmne thanks, I didn't see that comment, is there any assistance I can provide to help get that working ?

crmne:
Following up here too: Responses support landed on main via the new protocols layer (0875ce2d), including the semantic streaming events, so the streaming follow-up discussed in this thread is covered. Thanks for pushing on this @andrew-woblavobla.

PR #792, opened 2026-06-04
Support Anthropic adaptive thinking

### LINKED ISSUE #782: [BUG] Anthropic: thinking.type.enabled not supported, should use adaptive format for new models
### Basic checks

### What's broken?

When using RubyLLM with newer Anthropic models (e.g. Opus 4.7), requests fail with the error:

Accumulating error chunk: {"type":"error","error":{"type":"invalid_request_error","message":"\"thinking.type.enabled\" is not supported for this model. Use \"thinking.type.adaptive\" and \"output_config.effort\" to control thinking behavior."},"request_id":"..."}

The code currently always sets `thinking: {type: "enabled"}` in the payload (see `lib/ruby_llm/providers/anthropic/chat.rb`).

https://platform.claude.com/docs/en/build-with-claude/adaptive-thinking

### How to reproduce

1. Configure RubyLLM to use a newer Anthropic Claude model (e.g. Opus 4.7)
2. Call `chat.with_thinking(...)`
3. Request fails with the error above

### Expected behavior

RubyLLM should send `thinking: {type: "adaptive"}` with `output_config.effort`, as required by new Anthropic models. It should detect/model-switch as needed to remain compatible with all supported Claude versions.

### What actually happened

RubyLLM sends the legacy format and Anthropic rejects the call. Error: "'thinking.type.enabled' is not supported for this model. Use 'thinking.type.adaptive' and 'output_config.effort' to control thinking behavior."

### Environment

- Provider: Anthropic

ebeigarts:
Support Anthropic adaptive thinking

Newer Claude models (Opus 4.7+) reject thinking.type "enabled" and
require type "adaptive" with output_config.effort instead. Send the
adaptive format when an effort is given, keep the budget_tokens format
when a budget is given, and merge output_config so structured output
and adaptive thinking compose.

Fixes https://github.com/crmne/ruby_llm/issues/782

### DIFF
--- a/docs/_core_features/thinking.md
+++ b/docs/_core_features/thinking.md
@@ -116,8 +116,8 @@ end
 
 ## Provider Notes
 
-- Claude uses a thinking budget and can return both text and signature.
-- Anthropic requires a thinking budget.
+- Claude supports budget or effort depending on the model, and can return both text and signature.
+- Anthropic requires a thinking budget or effort, depending on the model.
 - Bedrock thinking params are model-dependent; models may accept budget, effort, or provider-specific fields.
 - Gemini 2.5 uses a token budget; Gemini 3 uses effort levels.
 - OpenAI reasoning models accept `effort` but may not return thinking text or signatures.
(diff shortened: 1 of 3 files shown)


ebeigarts:
Fixed in https://github.com/crmne/ruby_llm/commit/3f27b864ea79f0556c6296557b72f6a041b268cb

PR #811, opened 2026-06-12
Document TrustedRouter OpenAI-compatible setup

jperla:
## Summary
- document TrustedRouter as an OpenAI-compatible routing option
- use `https://api.trustedrouter.com/v1` as the base URL
- include the `trustedrouter/zdr` model alias for privacy-sensitive agent workloads

## Why
TrustedRouter is useful for workflows that involve private code, customer data, or other sensitive context. It provides open-source, verifiable attested routing and does not log prompts or outputs by default.

## Validation
- `git diff --check`

### DIFF
--- a/docs/_getting_started/configuration.md
+++ b/docs/_getting_started/configuration.md
@@ -154,6 +154,19 @@ end
 chat = RubyLLM.chat(model: 'my-custom-model', provider: :openai, assume_model_exists: true)
 ```
 
+For privacy-sensitive applications, you can point the same OpenAI-compatible configuration at TrustedRouter:
+
+```ruby
+RubyLLM.configure do |config|
+  config.openai_api_key = ENV['TRUSTEDROUTER_API_KEY']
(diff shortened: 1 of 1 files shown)


crmne:
Looks like promotion.

jperla:
Thanks for the feedback. I agree the current shape can read as promotion. Is there any format you would accept, such as a neutral OpenAI-compatible router setup example, a broader custom base URL guide, or an issue first? TrustedRouter is built as an end-to-end encrypted, open-source router, and I am happy to rewrite this so it is useful without sounding like an ad.

jperla:
One clarification on why I am asking about an acceptable format: TrustedRouter is built to be the #1 end-to-end encrypted open-source router, not just another provider listing. The value for users is a verifiable private OpenAI-compatible route for sensitive prompts. If the original PR felt promotional, would a neutral custom OpenAI-compatible router example be acceptable instead?

PR #813, opened 2026-06-16
Fix Bedrock streaming dropping all chunks when Faraday env is nil

chen-anders:
## What this does

Fixes a bug where **Bedrock ConverseStream responses come back completely empty** (zero chunks, blank content, nil token counts) even though the HTTP request succeeds with status 200.

The Faraday 2 `on_data` callback gates chunk parsing on `env&.status == 200`. With the `net_http` adapter, `env` is **nil** during streaming (the status is not yet known when chunks arrive), so this condition is false for every chunk. Each valid AWS event-stream frame is then routed to the failed-response handler, which tries to `JSON.parse` binary eventstream bytes, fails, logs a "failed stream error chunk" debug line, and silently drops it. The result is an empty streamed response.

This affects any app on Faraday 2.x + the net_http adapter (Faraday's default) using Bedrock streaming.

**Fix:** invert the guard so a nil env (status unknown) falls through to normal chunk parsing, and only a present env reporting a non-200 status is treated as a failure. Applied in both spots that build the v2 `on_data` proc:

- `RubyLLM::Streaming::FaradayHandlers.v2_on_data` (shared SSE handler)
- `RubyLLM::Protocols::Converse::Streaming#stream_response` (Bedrock ConverseStream)

### Reproduction

```ruby
chat = RubyLLM.chat(model: "us.anthropic.claude-sonnet-4-6", provider: "bedrock", assume_model_exists: true)

# Non-streaming — works:
chat.ask("Say BANANA only.").content          # => "BANANA"

# Streaming — broken before this fix:
chunks = []
resp = chat.ask("Say BANANA only.") { |c| chunks << c }
chunks.size    # => 0   (expected: > 0)
resp.content   # => ""   (expected: "BANANA")
```

With Faraday 2.x + net_http, the debug log shows the data arriving fine (`{"delta":{"text":"BANANA"}}`, usage tokens, status 200) but every frame hitting "Failed Bedrock stream error chunk". After the fix: `chunks.size => 5`, `resp.content => "BANANA"`, tokens populated.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

  - [x] Added unit tests covering nil-env / 200 / non-200 routing for both code paths (`spec/ruby_llm/streaming_spec.rb`, `spec/ruby_llm/protocols/converse/streaming_spec.rb`)

## AI-generated code

## API changes

- [ ] Breaking change
- [x] No API changes

### DIFF
--- a/lib/ruby_llm/protocols/converse/streaming.rb
+++ b/lib/ruby_llm/protocols/converse/streaming.rb
@@ -33,10 +33,15 @@ module RubyLLM
               end
             else
               req.options.on_data = proc do |chunk, _bytes, env|
-                if env&.status == 200
-                  parse_stream_chunk(decoder, chunk, accumulator, &block)
-                else
+                # A nil env means the status is not yet known (Faraday 2 with the
+                # net_http adapter passes nil during streaming) — process the chunk
+                # normally. Only a present env reporting a non-200 status is a real
+                # failure. Gating on `env&.status == 200` would discard every chunk
+                # whenever env is nil, yielding an empty streamed response.
+                if env && env.status != 200
                   handle_failed_stream(chunk, env)
+                else
+                  parse_stream_chunk(decoder, chunk, accumulator, &block)
                 end
               end
             end
--- a/lib/ruby_llm/protocols/converse/streaming.rb
+++ b/lib/ruby_llm/protocols/converse/streaming.rb
@@ -22,28 +22,22 @@ module RubyLLM
           decoder = event_stream_decoder
           body = JSON.generate(payload)
 
+          faraday_v1 = Faraday::VERSION.start_with?('1')
+          on_data = RubyLLM::Streaming::FaradayHandlers.build(
+            faraday_v1: faraday_v1,
+            on_chunk: ->(chunk, _env) { parse_stream_chunk(decoder, chunk, accumulator, &block) },
+            on_failed_response: ->(chunk, env) { handle_failed_stream(chunk, env) }
+          )
+
           response = @connection.post(stream_url, payload) do |req|
             req.headers.merge!(@provider.sign_headers('POST', stream_url, body))
             req.headers.merge!(additional_headers) unless additional_headers.empty?
             req.headers['Accept'] = 'application/vnd.amazon.eventstream'
 
-            if Faraday::VERSION.start_with?('1')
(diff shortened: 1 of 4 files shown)


lib/ruby_llm/streaming.rb
```diff
@@ -165,10 +165,16 @@ def v1_on_data(on_chunk)
 
       def v2_on_data(on_chunk, on_failed_response)
         proc do |chunk, _bytes, env|
-          if env&.status == 200
-            on_chunk.call(chunk, env)
-          else
+          # A nil env means the status is not yet known (Faraday 2 with the
```
crmne: Let's keep this short, and treat unknown status the same whether `env` is nil or `env.status` is nil.

```ruby
# Some adapters do not expose response status during on_data.
status = env&.status

if status && status != 200
  on_failed_response.call(chunk, env)
else
  on_chunk.call(chunk, env)
end
```

lib/ruby_llm/protocols/converse/streaming.rb
```diff
@@ -33,10 +33,15 @@ def stream_response(payload, additional_headers = {}, &block)
               end
             else
               req.options.on_data = proc do |chunk, _bytes, env|
-                if env&.status == 200
-                  parse_stream_chunk(decoder, chunk, accumulator, &block)
-                else
+                # A nil env means the status is not yet known (Faraday 2 with the
```
crmne: Converse needs its own AWS Event Stream parser, but it can still use the existing Faraday `on_data` wrapper so this routing rule lives in one place.

```ruby
faraday_v1 = Faraday::VERSION.start_with?('1')
on_data = RubyLLM::Streaming::FaradayHandlers.build(
  faraday_v1: faraday_v1,
  on_chunk: ->(chunk, _env) { parse_stream_chunk(decoder, chunk, accumulator, &block) },
  on_failed_response: ->(chunk, env) { handle_failed_stream(chunk, env) }
)

if faraday_v1
  req.options[:on_data] = on_data
else
  req.options.on_data = on_data
end
```

crmne (changes requested):
The bug fix and test coverage look right. Please tighten the shape before merging: keep the Faraday status-routing rule in one helper, and keep the comment/predicate small.

PR #815, opened 2026-06-17
Fix flaky acts_as_model specs caused by RubyLLM.logger spec leaving config nil

cgmoore120:
## Problem

The `acts_as_model` "model registry integration" specs (`loads models from database when configured`, `finds models from database`) fail intermittently under the parallel test queue (`bin/rspec-queue`), with errors like:

```
Failure/Error: expect(models.all.map(&:id)).to include('test-model')
```

The failures are order-dependent: they only appear when certain specs land in the same worker, so they surface non-deterministically depending on how `test-queue` distributes examples (more likely on slower Ruby versions).

## Root cause

The `RubyLLM.logger` spec resets the global config/logger in its `after` hook by setting them to `nil` instead of restoring the originals:

```ruby
after do
  described_class.instance_variable_set(:@config, nil)
  described_class.instance_variable_set(:@logger, nil)
end
```

Once `@config` is left `nil`, the next access lazily rebuilds a fresh `Configuration`. But `config.model_registry_source` is only ever set once — in `RubyLLM::ActiveRecord::ActsAs.included` at load time — so the rebuilt config no longer has it. Any later spec in the same worker that relies on the database-backed model registry (`Models#load_models` reads `config.model_registry_source`) then silently falls back to the JSON registry and fails to find the test records.

This reproduces deterministically by running the two specs in order:

```
rspec spec/ruby_llm_spec.rb spec/ruby_llm/active_record/acts_as_model_spec.rb --order defined
# => acts_as_model registry integration: 2 failures
```

## Fix

Use an `around` hook that saves and restores the original `@config`/`@logger`, so the logger spec leaves global state untouched. After the fix the same ordered run passes (22 examples, 0 failures), and `ruby_llm_spec` itself stays green.


crmne:
Thanks for this. This is now covered on main via efbe8faf: the logger spec saves and restores the original @config and @logger in an around hook instead of leaving either ivar nil, so later specs in the same worker keep the ActiveRecord model registry config.

Verified with:

```sh
bundle exec rspec spec/ruby_llm_spec.rb
```

2 examples, 0 failures.

Closing this PR as superseded by the main-branch fix.

PR #822, opened 2026-06-25
Add first-class LLMGateway.io provider

### LINKED ISSUE #821: [FEATURE] Add LLMGateway.io as a built-in provider
### Scope check

### Due diligence

[LLMGateway.io](https://llmgateway.io) is an OpenAI-compatible gateway that routes one endpoint and key to 35+ upstream providers (OpenAI, Anthropic, Google, etc), similar to OpenRouter.

Today the only way to reach it through RubyLLM is to repoint the OpenAI provider:

```ruby
RubyLLM.configure do |config|
  config.openai_api_base = "https://api.llmgateway.io/v1"
  config.openai_api_key  = ENV["LLM_GATEWAY_API_KEY"]
end
```

That works, but it globally hijacks the `openai` provider: you can't talk to OpenAI directly and the gateway at the same time.

### Proposed solution

Add a first-class `:llm_gateway` provider, exactly like the existing OpenRouter provider

Talking to a provider (HTTP API, host, auth headers, model catalog, error envelope) is core LLM communication, not application logic.

goianiense:
## Summary
- Add a dedicated `:llm_gateway` provider (modeled on OpenRouter) for [LLMGateway.io](https://llmgateway.io), an OpenAI-compatible gateway for 35+ upstream providers.
- Register config options `llm_gateway_api_key` / `llm_gateway_api_base` (default `https://api.llmgateway.io/v1`) so the gateway can be used alongside the real OpenAI provider instead of hijacking `openai_api_base`.
- Include chat completions, models catalog parsing, streaming, image generation, and files support, with unit coverage for config, API base resolution, and error parsing.

Closes #821.

## Usage

```ruby
RubyLLM.configure do |config|
  config.openai_api_key = ENV["OPENAI_API_KEY"]
  config.llm_gateway_api_key = ENV["LLM_GATEWAY_API_KEY"]
end

# Gateway and OpenAI work independently
RubyLLM.chat(model: "gpt-4o", provider: :llm_gateway)
RubyLLM.chat(model: "gpt-4o", provider: :openai)
```

## Test plan
- [x] `bundle exec rspec spec/ruby_llm/providers/llm_gateway/parse_error_spec.rb spec/ruby_llm/provider_spec.rb spec/ruby_llm/configuration_spec.rb` (25 examples, 0 failures)
- [x] Smoke test: separate API bases/keys for `:llm_gateway` and `:openai`
- [ ] Optional: live chat against LLMGateway.io with `LLM_GATEWAY_API_KEY` set

### DIFF
--- a/lib/ruby_llm.rb
+++ b/lib/ruby_llm.rb
@@ -23,6 +23,7 @@ loader.inflector.inflect(
   'deepseek' => 'DeepSeek',
   'gpustack' => 'GPUStack',
   'llm' => 'LLM',
+  'llm_gateway' => 'LLMGateway',
   'mistral' => 'Mistral',
   'openai' => 'OpenAI',
   'openrouter' => 'OpenRouter',
@@ -127,6 +128,7 @@ RubyLLM::Provider.register :bedrock, RubyLLM::Providers::Bedrock
 RubyLLM::Provider.register :deepseek, RubyLLM::Providers::DeepSeek
 RubyLLM::Provider.register :gemini, RubyLLM::Providers::Gemini
 RubyLLM::Provider.register :gpustack, RubyLLM::Providers::GPUStack
+RubyLLM::Provider.register :llm_gateway, RubyLLM::Providers::LLMGateway
 RubyLLM::Provider.register :mistral, RubyLLM::Providers::Mistral
 RubyLLM::Provider.register :ollama, RubyLLM::Providers::Ollama
 RubyLLM::Provider.register :openai, RubyLLM::Providers::OpenAI
--- a/lib/ruby_llm/models.rb
+++ b/lib/ruby_llm/models.rb
@@ -35,6 +35,7 @@ module RubyLLM
       vertexai
       bedrock
       openrouter
+      llm_gateway
       azure
       ollama
       gpustack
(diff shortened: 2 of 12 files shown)


crmne:
Thanks for opening this and for the linked issue.

I understand the OpenRouter comparison, but I don't think LLMGateway.io clears the bar for a built-in RubyLLM provider right now. The provider guidance in `CONTRIBUTING.md` keeps core providers to a high acceptance bar, and smaller or emerging providers/gateways should generally live as community gems rather than expanding RubyLLM core.

This seems like a good fit for a community provider gem, especially since the separate provider wrapper can preserve its own API key/base while staying outside core. We're working on a provider generator for RubyLLM 2.0 to make community providers easier to create and maintain, and we'll add docs/guidelines for community provider gems, including discovery conventions such as linking to a GitHub topic/tag for RubyLLM providers.

So my recommendation is to package this outside core for now rather than continue this as a built-in provider PR.

steebchen:
hey @crmne Luca here, co-founder of LLMGateway. can you clarify what the bar is? note that we are SOC2 Type 2 certified and our revenue is public - https://trustmrr.com/startup/llm-gateway, let me know if there are other requirements to fulfil so we can work on these. we're also 100% OSS: https://github.com/theopenco/llmgateway

PR #823, opened 2026-06-25
Expose persisted assistant message during streaming

### LINKED ISSUE #814: [FEATURE] Expose persisted assistant message during streaming
### Scope check
- [x] This **can't be solved in application code** with current RubyLLM (cleanly — see below)
- [x] I read the Contributing Guide

### Due diligence

When streaming a response with `acts_as_chat` and a block, the most common need is to broadcast/render chunks against the assistant message that RubyLLM is currently persisting. Today that message is reachable via:

1. `conversation.messages.last` — relies on the implicit invariant that the `has_many` association is still loaded (it is, because `to_llm` calls `messages_association.to_a` before `complete`). Brittle: a stray `reload` and `.last` issues an SQL `ORDER BY created_at ASC LIMIT 1` that can return another job's record in a multi-worker setup.
2. `conversation.before_message { @msg = conversation.messages.last }` + capture in an outer ivar — works, but adds ceremony to every streaming call site and still leans on the same `.last` lookup.
3. `conversation.instance_variable_get(:@message)` — exactly the record `persist_new_message` just created, but reaching into a private ivar (`chat_methods.rb#L371`).

All three work; none feel idiomatic. Looking at the gem README and the chat-UI generator, broadcasting from inside the streaming block is the canonical pattern, so the friction hits most users that do streaming + Rails persistence.

### Proposed solution

Add a public reader for the currently-persisting assistant message on the chat record:

```ruby
# lib/ruby_llm/active_record/chat_methods.rb
attr_reader :message
```

(or `current_message` if `message` collides with anyone's column — the existing ivar is `@message`.)

That turns the canonical streaming-broadcast pattern into:

```ruby
conversation.complete do |chunk|
  conversation.message.broadcast_append_chunk(chunk.content) if chunk.content.present?
end
```

No association reload, no manual capture, no private-ivar access. The reader simply exposes what `persist_new_message` already maintains, including across tool-flow turns where `before_message` re-fires and `@message` is reassigned.

The persistence callbacks that create/update `@message` are part of `acts_as_chat`'s contract with the host application — but the application has no public way to reference the record those callbacks just produced. The workarounds either (a) rely on undocumented in-memory association state or (b) reach into a private ivar that can be renamed without a deprecation cycle. A one-line `attr_reader` formalises something the gem already does internally, costs nothing in maintenance, and removes a sharp edge from the most common streaming use case.

Happy to send a PR with the reader + a spec if this direction is acceptable.

goianiense:
## Summary

- Add a public `attr_reader :message` on `acts_as_chat` models that exposes the assistant record currently being persisted during `complete`/`ask` (including across tool-flow turns).
- Lets streaming blocks broadcast against the stable in-memory record without relying on `messages.last`, `before_message` capture, or private ivar access.
- Update the chat UI generator job template and streaming docs to use the new reader.

Closes #814

## Usage

```ruby
chat.complete do |chunk|
  chat.message.broadcast_append_chunk(chunk.content) if chunk.content.present?
end
```

## Test plan

- [x] Specs for `chat.message` nil before completion and set afterwards
- [x] Specs for `chat.message` available inside the streaming block
- [x] Specs for reassignment on tool-flow assistant turns
- [x] Generator job expectations updated and passing
- [ ] CI green on this PR

### DIFF
--- a/docs/_advanced/rails-streaming.md
+++ b/docs/_advanced/rails-streaming.md
@@ -88,11 +88,8 @@ class ChatStreamJob < ApplicationJob
     chat = Chat.find(chat_id)
 
     chat.complete do |chunk|
-      # Get the assistant message record (created before streaming starts)
-      assistant_message = chat.messages.last
-      if chunk.content && assistant_message
-        assistant_message.broadcast_append_chunk(chunk.content)
-      end
+      # chat.message is the assistant record created before streaming starts
+      chat.message.broadcast_append_chunk(chunk.content) if chunk.content.present?
     end
   end
 end
--- a/lib/generators/ruby_llm/chat_ui/templates/jobs/chat_response_job.rb.tt
+++ b/lib/generators/ruby_llm/chat_ui/templates/jobs/chat_response_job.rb.tt
@@ -4,8 +4,7 @@ class <%= chat_job_class_name %> < ApplicationJob
 
     <%= chat_variable_name %>.ask(content) do |chunk|
       if chunk.content && !chunk.content.empty?
-        <%= message_variable_name %> = <%= chat_variable_name %>.<%= message_table_name %>.last
-        <%= message_variable_name %>.broadcast_append_chunk(chunk.content)
+        <%= chat_variable_name %>.message.broadcast_append_chunk(chunk.content)
       end
     end
   end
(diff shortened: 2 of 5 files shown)


crmne:
Thanks for putting this together.

I do not think we should add a public chat.message API here. In the Rails streaming flow RubyLLM creates the assistant row before yielding chunks, so chat.messages.last is already the public Rails-facing way to access the row being streamed.

Adding chat.message would duplicate the association API with a name that is too ambiguous, so I am closing this one.

PR #825, opened 2026-06-26
Add first-class per-call metadata for observability (#807)

### LINKED ISSUE #807: [FEATURE] First-class per-call metadata context for observability attribution
### Scope check

### Due diligence

RubyLLM already exposes great instrumentation hooks (`chat.ruby_llm`, `request.ruby_llm`, `embedding.ruby_llm`, etc.) and includes the originating object in the payload (e.g. `chat: self`). That is enough to *observe* LLM calls, but not enough to *attribute* them in a consistent, library-native way.

In production apps, every LLM call needs caller context for observability: which feature triggered it, what tags to attach to metrics, which trace/span to correlate with, etc. Today there is no first-class place to attach that context on a `Chat` (or other entry points) and have it travel with the object through instrumentation.

We currently work around this by monkey-patching `RubyLLM::Chat`:

```ruby
module RubyLLM::Telemetry
  Context = Data.define(:namespace, :tags)

  attr_reader :telemetry

  def with_telemetry(namespace, tags: {})
    @telemetry = Context.new(namespace:, tags:)
    self
  end
end

RubyLLM::Chat.include RubyLLM::Telemetry
```

Usage at call sites:

```ruby
RubyLLM.chat
  .with_telemetry(:chat_session)
  .with_instructions(instructions)
  .ask(message)

Ai::SummarizeAgent.chat
  .with_telemetry(:summary, tags: { type: summarizable_type })
  .ask(text)
```

And in an `ActiveSupport::Subscriber`:

```ruby
def chat(event)
  telemetry = event.payload[:chat].telemetry
  return unless telemetry

  Appsignal.increment_counter("#{telemetry.namespace}_requests", 1, telemetry.tags)
  Appsignal.add_distribution_value("#{telemetry.namespace}_input_tokens", event.payload[:input_tokens], telemetry.tags)
end
```

This works, but it is fragile:

- It relies on reopening `RubyLLM::Chat`, which can break across gem upgrades.
- It only covers `Chat`; `Embedding`, `Image`, `Transcription`, `Moderation`, and `Agent` have no equivalent hook.
- It is easy to forget in tests/fakes (we also include the module in test doubles).
- The name collides conceptually with `RubyLLM::Context` / `Chat#with_context`, which already mean *configuration* scoping — not caller metadata.

Related: #414 (observability) and #799 (extensibility) were closed with instrumentation improvements, but there is still no supported way to attach per-call metadata without patching core classes.

### Proposed solution

Add a first-class, opt-in **shared metadata** API on LLM entry points, following the existing fluent `with_*` builder style.

Suggested API (names open to bikeshedding — `metadata`, `tags`, or `telemetry` all work):

```ruby
chat
  .with_metadata(namespace: :chat_session, tags: { academy_id: 42 })
  .with_instructions(instructions)
  .ask(message)
```

Implementation sketch:

1. Introduce a small value object, e.g. `RubyLLM::Metadata = Data.define(:namespace, :tags)` (or a plain Hash if you prefer maximum flexibility).
2. Add `attr_reader :metadata` and `with_metadata(namespace: nil, tags: {})` to `Chat` (and parallel entry points where instrumentation already fires).
3. Merge metadata into instrumentation payloads explicitly, e.g. `metadata: chat.metadata`, so subscribers do not need to reach into arbitrary ivars on core objects.
4. Optionally support `with_metadata(tags: { ... })` to merge/append tags when chaining multiple calls.
5. Document the intended use: observability attribution, not provider request params (distinct from `with_params` / `with_safety_identifier`).

Example subscriber after the change:

```ruby
def chat(event)
  metadata = event.payload[:metadata]
  return unless metadata

  Appsignal.increment_counter("#{metadata.namespace}_requests", 1, metadata.tags)
end
```

If you want a narrower first iteration, `Chat` alone would already remove the most common monkey-patch.

Per-call attribution metadata is a cross-cutting concern of LLM communication — the same category as instrumentation itself. RubyLLM already owns the instrumentation contract and includes the chat object in event payloads; extending that contract with an official metadata slot is a small, library-level addition that prevents every app from patching `RubyLLM::Chat`.

This is not application business logic. The *values* (which namespace, which tags) are app-specific, but the *mechanism* for carrying them through the library belongs in RubyLLM — just like `RubyLLM::Context` carries per-call configuration, and instrumentation carries per-call telemetry events.

Without first-class support, teams either monkey-patch core classes or build wrapper objects around `RubyLLM.chat`, both of which fight the library's fluent API and miss non-Chat entry points.

goianiense:
## Summary

- Adds `RubyLLM::Metadata` (`namespace` + `tags`) and `Chat#with_metadata` for opt-in caller attribution on instrumentation events
- Includes `metadata` in `chat.ruby_llm` and `tool_call.ruby_llm` payloads; one-shot APIs (`embed`, `paint`, `moderate`, `speak`, `transcribe`) accept `metadata:`
- Delegates through `Agent` and ActiveRecord `ChatMethods`; documents usage in the instrumentation guide

Closes #807

## Test plan

- [x] `bundle exec rspec spec/ruby_llm/instrumentation_spec.rb`
- [x] Confirm subscribers can read `payload[:metadata]` without reaching into chat ivars
- [x] Smoke-test `with_metadata` chaining merges tags and replaces namespace when set

### DIFF
--- a/docs/_advanced/instrumentation.md
+++ b/docs/_advanced/instrumentation.md
@@ -25,6 +25,7 @@ After reading this guide, you will know:
 
 *   How to subscribe to RubyLLM events in Rails.
 *   How to connect RubyLLM instrumentation outside Rails.
+*   How to attach per-call metadata for observability attribution.
 *   Which events RubyLLM emits.
 *   Which payload fields may contain sensitive application data.
 
@@ -45,6 +46,36 @@ end
 ```
 
 When an instrumented block raises, Rails adds the standard `:exception` and `:exception_object` payload keys.
+
+## Per-call metadata
+
+Attach caller attribution with `with_metadata` on chat (and agents / ActiveRecord chat models). Metadata is **not** sent to providers — it only rides along on instrumentation payloads so subscribers can attribute metrics and traces without monkey-patching core classes.
+
+```ruby
+RubyLLM.chat
+  .with_metadata(namespace: :chat_session, tags: { academy_id: 42 })
+  .with_instructions(instructions)
+  .ask(message)
+```
+
+Successive `with_metadata` calls merge tags and replace `namespace` when a new one is given. One-shot APIs (`embed`, `paint`, `moderate`, `speak`, `transcribe`) accept a `metadata:` keyword (`RubyLLM::Metadata` or a hash with `:namespace` / `:tags`).
+
+```ruby
+# config/initializers/ruby_llm_instrumentation.rb
+ActiveSupport::Notifications.subscribe('chat.ruby_llm') do |_name, _start, _finish, _id, payload|
+  metadata = payload[:metadata]
+  next unless metadata
+
+  Appsignal.increment_counter("#{metadata.namespace}_requests", 1, metadata.tags)
+  Appsignal.add_distribution_value(
+    "#{metadata.namespace}_input_tokens",
+    payload[:input_tokens],
+    metadata.tags
+  )
+end
+```
+
+Use metadata for observability attribution only. Provider request options still go through `with_params` / keyword arguments; configuration scoping still uses [contexts]({% link _getting_started/configuration-connection.md %}#contexts-isolated-configurations).
 
 ## Outside Rails
 
--- a/lib/ruby_llm/active_record/chat_methods.rb
+++ b/lib/ruby_llm/active_record/chat_methods.rb
@@ -122,6 +122,11 @@ module RubyLLM
         self
       end
 
+      def with_metadata(...)
+        to_llm.with_metadata(...)
+        self
+      end
+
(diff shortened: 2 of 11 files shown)


crmne:
Thanks for the PR.

We took the core observability idea, but not this API shape. Main now supports simple metadata on the one-shot APIs and includes it directly in the instrumentation payload. That landed in 1bc6fc03.

The smaller shape intentionally avoids a RubyLLM::Metadata value object and does not add chat, agent, or ActiveRecord metadata APIs in this PR's form. Provider request customization stays under params.

Closing this PR as superseded. The linked issue remains open for the separate question of whether chat and tool-call instrumentation should get a minimal metadata API too.

PR #844, opened 2026-07-13
Send strict: false for function tools on the Responses API to preserve Chat Completions behavior

### LINKED ISSUE #843: [BUG] Responses API silently enables strict tool validation — optional tool parameters are always filled in
### Basic checks

### What's broken?

With OpenAI defaulting to the Responses API in v2, function tools silently switch to **strict** parameter validation, so optional tool parameters stop being optional: the model is grammar-constrained to fill in **every** property of the schema on every call.

The cause is a server-side default mismatch that RubyLLM doesn't compensate for:

- Chat Completions `function.strict`: default **`false`** (OpenAI OpenAPI spec: "Whether to enable strict schema adherence... default: false")
- Responses `FunctionTool.strict`: default **`true`** ("Whether to enforce strict parameter validation. Default `true`.")

`Protocols::Responses::Tools.tool_for` renders `{type:, name:, description:, parameters:}` and never sends `strict`, so the server default applies. On 1.x / Chat Completions that meant non-strict; on the Responses API it means strict. Under strict mode, OpenAI requires every property to be listed in `required` (optionality is expressed as nullable types instead), so schemas with genuinely optional parameters no longer behave as designed.

The [upgrading guide](https://rubyllm.com/reference/upgrading/) covers the Responses API switch but doesn't mention this behavior change, so it's easy to hit in production without noticing — tool calls keep succeeding, just with every optional filter populated.

### How to reproduce

1. Define a tool with optional parameters:

```ruby
class SearchTool < RubyLLM::Tool
  description "Searches calls with optional filters"

  parameters type: "object",
    properties: {
      keyword:   { type: "string",  description: "Optional title filter. Omit unless the user asks." },
      date_from: { type: "string",  description: "Optional ISO date filter. Omit unless the user asks." },
      page:      { type: "integer", description: "Omit unless you need the next page." }
    },
    required: [],
    additionalProperties: false

  def execute(keyword: nil, date_from: nil, page: nil)
    { keyword:, date_from:, page: }
  end
end
```

2. `RubyLLM.chat(model: "gpt-5.4").with_tool(SearchTool).ask("search for calls about pricing")`
3. Inspect the tool call arguments (`RUBYLLM_DEBUG=true`, or the `on_tool_call` callback).

### Expected behavior

The model omits parameters it doesn't need — e.g. `{"keyword": "pricing"}` — as it did on Chat Completions (RubyLLM ≤ 1.x behavior for the same tool, same model).

### What actually happened

Every property is populated on every call, e.g. `{"keyword": "pricing", "date_from": "2026-01-01", "page": 1}` — the model invents values for filters it was told to omit, changing search results. Same tool + same model on `protocol: :chat_completions` behaves as expected, confirming it's the strict default and not the model.

### Environment

- Ruby version: 3.4.8
- RubyLLM version: main (`b229895`)
- Provider (OpenAI, Anthropic, etc.): OpenAI (Responses API), also reproduced on Azure OpenAI v1
- OS: Linux

### Proposed fix

Send `strict: false` explicitly in `Responses::Tools.tool_for` to preserve the Chat Completions behavior, and let tools opt in with `provider_options strict: true` (which already merges over the definition). I have a patch ready with specs and docs updates (upgrading guide note + a strict-mode section in tool-parameters) — happy to open the PR if you agree with defaulting to non-strict. If you'd rather keep strict-by-default, this at least deserves a call-out in the upgrading guide, since properly strict schemas also require every property in `required` and nullable types for optionals, which most existing tools won't have.

bdegomme:
Fixes #843

This is AI generated but I understand every line. I'm also fine with accepting the new default strict true, but then it should be mentioned in the upcoming 2.0 [upgrading guide](https://rubyllm.com/upgrading/).

## What this PR does

Explicitly sends `strict: false` in function tool definitions rendered by `Protocols::Responses::Tools.tool_for`, restoring the tool-calling behavior applications had on Chat Completions. Tools can opt into strict validation per class with `provider_options strict: true`.

## The problem

OpenAI's two APIs have opposite server-side defaults for function tool `strict`:

- Chat Completions `FunctionObject.strict`: **default `false`** ([OpenAPI spec](https://github.com/openai/openai-openapi))
- Responses `FunctionTool.strict`: **default `true`** ("Whether to enforce strict parameter validation. Default `true`.")

RubyLLM never sends the `strict` field, so when OpenAI switched to defaulting to the Responses API, every existing tool silently flipped into strict mode. Under strict mode the model's output is grammar-constrained to emit **every** property in the schema — parameters left out of `required` stop being optional (strict mode expresses optionality as nullable types + membership in `required` instead).

Concretely: a search tool with a dozen optional filters that previously received only the relevant ones now gets a value for every filter on every call, changing tool behavior without any application code change. The [upgrade guide](https://rubyllm.com/reference/upgrading/) covers the Responses API switch but doesn't mention this side effect, and there's no way to notice it except by observing degraded tool calls.

## The fix

`tool_for` now always includes `strict: false` in the flat function definition. `provider_options` is merged after, so `provider_options strict: true` on a tool class opts back in for schemas that meet the strict-mode requirements.

Docs updated:

- `docs/_reference/upgrading.md` — note under "OpenAI now defaults to the Responses API" explaining tools stay non-strict and how to opt in.
- `docs/_core_features/tool-parameters.md` — new "Strict Mode on the OpenAI Responses API" subsection with an opt-in example (nullable optional params, all properties in `required`).

## Tests

- Updated the flat-function-definition expectation in `spec/ruby_llm/protocols/responses/chat_spec.rb` and added an opt-in test for `provider_options strict: true`.
- Existing OpenAI VCR cassettes still match (default VCR matching is method + URI), verified with `CI=true bundle exec rspec spec/ruby_llm/chat_tools_spec.rb -e openai` — 18 examples, 0 failures.

### DIFF
--- a/docs/_core_features/tool-parameters.md
+++ b/docs/_core_features/tool-parameters.md
@@ -237,17 +237,18 @@ The Responses API defaults function tools to strict schema validation, forcing t
 To opt into [strict validation](https://platform.openai.com/docs/guides/function-calling#strict-mode) per tool:
 
 ```ruby
+class LookupParams < RubyLLM::Schema
+  string :sku, description: "Product SKU"
+
+  any_of :locale, description: "Country code" do
+    string
+    null
+  end
+end
+
 class Lookup < RubyLLM::Tool
   description "Performs catalog lookups"
-
-  parameters type: "object",
-    properties: {
-      sku: { type: "string", description: "Product SKU" },
-      locale: { type: %w[string null], description: "Country code" }
-    },
-    required: %w[sku locale],
-    additionalProperties: false
-
+  parameters LookupParams
   provider_options strict: true
 
   def execute(sku:, locale: nil)
--- a/docs/_core_features/tool-parameters.md
+++ b/docs/_core_features/tool-parameters.md
@@ -230,6 +230,34 @@ Provider metadata is passed through verbatim. Turn on `RUBYLLM_DEBUG=true` if yo
 
 Pass `nil` to `provider_options` to clear provider-specific tool options.
 
+### Strict Mode on the OpenAI Responses API
+
+The Responses API defaults function tools to strict schema validation, forcing the model to emit **every** property on each call. RubyLLM sends `strict: false` so tools keep the Chat Completions behavior: parameters not in `required` can simply be omitted.
+
+To opt into [strict validation](https://platform.openai.com/docs/guides/function-calling#strict-mode) per tool:
+
+```ruby
+class Lookup < RubyLLM::Tool
+  description "Performs catalog lookups"
+
+  parameters type: "object",
+    properties: {
+      sku: { type: "string", description: "Product SKU" },
+      locale: { type: %w[string null], description: "Country code" }
+    },
+    required: %w[sku locale],
+    additionalProperties: false
+
+  provider_options strict: true
+
+  def execute(sku:, locale: nil)
+    # ...
+  end
+end
+```
+
+Strict mode requires every property in `required`; make optional parameters nullable instead.
+
 ## Next Steps
 
 *   [Controlling Tool Execution]({% link _core_features/tool-execution.md %}) - Steer tool choice, call counts, concurrency, and callbacks.
--- a/docs/_reference/upgrading.md
+++ b/docs/_reference/upgrading.md
@@ -506,6 +506,8 @@ RubyLLM.chat(model: 'gpt-5.4', protocol: :chat_completions)
 
 If you pass Chat Completions-only options via `with_provider_options` (like `response_format`), either switch those chats to `:chat_completions` or use the Responses API equivalents (`text: { format: ... }`).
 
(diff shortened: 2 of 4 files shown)


docs/_core_features/tool-parameters.md
```diff
@@ -230,6 +230,34 @@ Provider metadata is passed through verbatim. Turn on `RUBYLLM_DEBUG=true` if yo
...
+class Lookup < RubyLLM::Tool
+  description "Performs catalog lookups"
+
+  parameters type: "object",
+    properties: {
+      sku: { type: "string", description: "Product SKU" },
+      locale: { type: %w[string null], description: "Country code" }
+    },
+    required: %w[sku locale],
+    additionalProperties: false
```
crmne: could you use `RubyLLM::Schema` here?

crmne (dismissed):
Can you resolve the conflicts and change this?

PR #845, opened 2026-07-13
Finalize open thinking blocks on messageStop for Bedrock Converse streaming

mattwebbio:
## Summary
- Bedrock's Converse stream doesn't reliably send a `content_block_stop` for every reasoning-content index before the stream ends — observed for a redacted thinking block immediately superseded by the next block's `content_block_start` with no stop in between.
- That left `thinking.blocks` empty for such turns, forcing replay onto `format_thinking_blocks`' lossy single-block reconstruction, which Anthropic rejects on the next request with `Invalid \`data\` in \`redacted_thinking\` block`.
- `messageStop` now finalizes any block still open at that point, but only when it's structurally complete (has a signature or redacted-content marker present) — a genuinely truncated block with neither is still dropped, unchanged from existing behavior.

Root-caused live against agents_app staging: a fresh MCP conversation making a single GitHub tool call reproduced this exact error on every attempt. DB inspection of the persisted `llm_messages` row confirmed the failure shape — `thinking_signature` present, `thinking_blocks` and `thinking_text` both null — i.e. the reconstruction-fallback path this PR closes.

## Test plan
- [x] `bundle exec rspec spec/ruby_llm/protocols/converse/streaming_spec.rb spec/ruby_llm/stream_accumulator_spec.rb` — 18 examples, 0 failures (includes new regression test + the pre-existing truncation-drop test, both green)
- [x] `bundle exec rspec spec/ruby_llm/protocols/` — 309 examples, 0 failures
- [x] `bundle exec rubocop lib/ruby_llm/protocols/converse/streaming.rb spec/ruby_llm/protocols/converse/streaming_spec.rb` — no offenses
- [ ] Bump agents_app's `Gemfile.lock` to this commit and re-run the live repro against staging


mattwebbio:
Opened against the wrong repo by mistake (tooling defaulted to the fork parent). Closing — correct PR is on appfolio/ruby_llm.

PR #848, opened 2026-07-15
Fix streaming error parsing crashing on non-Hash JSON error bodies

### LINKED ISSUE #837: Streaming: parse_streaming_error crashes with 'TypeError: String does not have #dig' on a non-Hash JSON error body
## Summary

When a **streaming** request gets a non-200 response whose body parses to
a JSON value that is *not* a Hash (e.g. a bare JSON string), the OpenAI
provider's `parse_streaming_error` raises `TypeError: String does not
have #dig` instead of surfacing the provider's error message. The
identical body is handled correctly on the non-streaming path, so the
real error message is lost only when streaming.

Version: **ruby_llm 1.16.0** (latest), Ruby 3.3.11, faraday 2.14.2.

## Root cause

Two co-located issues in the streaming error path:

1. `RubyLLM::Streaming#handle_failed_response` (`lib/ruby_llm/streaming.rb`)
   calls `handle_parsed_error` with no `is_a?(Hash)` guard — unlike the
   success path `handle_data`, which does `return parsed unless parsed.is_a?(Hash) && parsed.key?('error')`.

2. `RubyLLM::Providers::OpenAI::Streaming#parse_streaming_error`
   (`lib/ruby_llm/providers/openai/streaming.rb:39`) assumes a Hash:

   ```ruby
   def parse_streaming_error(data)
     error_data = JSON.parse(data)
     return unless error_data['error']       # ← intended as a key-presence check
     case error_data.dig('error', 'type')    # ← line 43: String#dig => TypeError
     ...
   end
   ```

   The guard `return unless error_data['error']` is meant to test for a
   `"error"` key, but when `error_data` is a `String`, `String#[]('error')`
   does **substring matching**. A body that parses to a string *containing*
   the substring `error` slips past the guard and reaches `.dig`, which
   `String` doesn't implement.

   (If the string does *not* contain `error`, the guard instead returns
   `nil` and the real error is silently swallowed — a quieter variant of
   the same missing-type-check.)

By contrast, the non-streaming `RubyLLM::Provider#parse_error`
(`lib/ruby_llm/provider.rb`) is robust — it `case`s on `Hash` / `Array` /
`else`, returning the raw body as the message in the `else` arm.

## Reproduction

```ruby
require 'ruby_llm'

# A gateway/proxy returning a bare JSON string for an unavailable model,
# Content-Type: application/json, HTTP 404 — this reaches the streaming
# failed-response path.
body = %q("The model foo is not available in your region (error).")

RubyLLM::Providers::OpenAI::Streaming.parse_streaming_error(body)
# => TypeError: String does not have #dig method
#      .../providers/openai/streaming.rb:43:in `parse_streaming_error'

# The non-streaming path handles the identical body fine:
RubyLLM.configure { |c| c.openai_api_key = 'x' }
prov = RubyLLM::Providers::OpenAI.new(RubyLLM.config)
resp = Struct.new(:body, :status).new(body, 404)
prov.parse_error(resp)
# => "The model foo is not available in your region (error)."
```

Real-world trigger: pointing an OpenAI-compatible client at a model/region
the endpoint rejects (in my case Grok via an OpenAI-compatible endpoint,
not yet available in Europe) while streaming. The user sees a cryptic
Ruby `TypeError` in place of the endpoint's actual "not available"
message.

Full backtrace from the app:

```
TypeError: String does not have #dig method
  ruby_llm-1.16.0/lib/ruby_llm/providers/openai/streaming.rb:43:in `dig'
  ruby_llm-1.16.0/lib/ruby_llm/providers/openai/streaming.rb:43:in `parse_streaming_error'
  ruby_llm-1.16.0/lib/ruby_llm/streaming.rb:128:in `handle_parsed_error'
  ruby_llm-1.16.0/lib/ruby_llm/streaming.rb:90:in `handle_failed_response'
  ruby_llm-1.16.0/lib/ruby_llm/streaming.rb:54:in `block in build_on_data_handler'
  ...
```

## Suggested fix

Guard for a non-Hash parsed body, mirroring the non-streaming
`Provider#parse_error`. Minimal version in the OpenAI provider:

```ruby
def parse_streaming_error(data)
  error_data = JSON.parse(data)
  return [500, error_data] unless error_data.is_a?(Hash) && error_data['error']

  case error_data.dig('error', 'type')
  when 'server_error'
    [500, error_data['error']['message']]
  when 'rate_limit_exceeded', 'insufficient_quota'
    [429, error_data['error']['message']]
  else
    [400, error_data['error']['message']]
  end
end
```

and/or add the `is_a?(Hash)` guard in `Streaming#handle_failed_response`
so the failed-response path matches `handle_data`'s existing check. Happy
to open a PR if the direction looks right.

jonthedecepticon:
## What this does

Fixes #837.

When a **streaming** request fails with a body that parses to something other than the expected `{"error": {...}}` shape, `ChatCompletions::Streaming#parse_streaming_error` crashed instead of surfacing the provider's error message:

- `{"error": "The model is not available in your region."}` (string-valued `error`) → `TypeError: String does not have #dig method`
- a bare JSON string body containing the substring `error` → slips past the `error_data['error']` guard via `String#[]` substring matching → `NoMethodError: undefined method 'dig' for an instance of String`

The identical bodies are handled fine on the non-streaming path by `Provider#parse_error`, which already guards Hash/Array/String shapes — so the real error message was lost only when streaming.

The fix guards the parsed body shape before digging into it:
- non-Hash bodies return no status hint (the failed HTTP status wins in `build_stream_error_response` anyway, and `Provider#parse_error` still extracts the message downstream)
- string-valued `error` keys are surfaced as the error message directly

No change to the happy path or to correctly-shaped error objects.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

## Tests

- Unit specs for `parse_streaming_error` covering all classification branches (server_error → 500, rate limit → 429, other → 400, no error key → nil) plus the two crashing shapes from the issue.
- End-to-end streaming specs (webmock-stubbed, no cassettes needed) on a ChatCompletions provider for both body shapes, on both Faraday 1 and 2, asserting the proper `RubyLLM::ServerError` with the provider's message. One caveat encoded in the spec: Faraday 1 streams the error body away before `on_complete`, so for bare-string bodies only the status (and thus error class) is available there — the message assertion applies to Faraday 2.

## AI-generated code

## API changes

- [ ] Breaking change
- [x] No API changes

### DIFF
--- a/lib/ruby_llm/protocols/chat_completions/streaming.rb
+++ b/lib/ruby_llm/protocols/chat_completions/streaming.rb
@@ -47,15 +47,18 @@ module RubyLLM
 
         def parse_streaming_error(data)
           error_data = JSON.parse(data)
-          return unless error_data['error']
+          return unless error_data.is_a?(Hash) && error_data['error']
 
-          case error_data.dig('error', 'type')
+          error = error_data['error']
+          return [500, error] unless error.is_a?(Hash)
+
+          case error['type']
           when 'server_error'
-            [500, error_data['error']['message']]
+            [500, error['message']]
           when 'rate_limit_exceeded', 'insufficient_quota'
-            [429, error_data['error']['message']]
+            [429, error['message']]
           else
-            [400, error_data['error']['message']]
+            [400, error['message']]
           end
         end
       end
--- a/spec/ruby_llm/chat_streaming_spec.rb
+++ b/spec/ruby_llm/chat_streaming_spec.rb
@@ -106,5 +106,50 @@ RSpec.describe RubyLLM::Chat do
         end
       end
     end
+
+    context 'with non-Hash error bodies (deepseek/deepseek-chat)' do
+      let(:chat) { RubyLLM.chat(model: 'deepseek-chat', provider: :deepseek) }
+      let(:url) { 'https://api.deepseek.com/chat/completions' }
+
+      { '1' => '1.10.0', '2' => '2.0.0' }.each do |major, faraday_version|
+        describe "Faraday version #{major}" do # rubocop:disable RSpec/NestedGroups
+          before do
+            stub_const('Faraday::VERSION', faraday_version)
+          end
+
+          it 'surfaces the message when the error value is a string' do
+            stub_request(:post, url).to_return(
+              status: 500,
+              body: { error: 'The model is not available in your region.' }.to_json,
+              headers: { 'Content-Type' => 'application/json' }
+            )
+
+            expect do
+              chat.ask('Count from 1 to 3') { |chunk| chunk }
+            end.to raise_error(RubyLLM::ServerError, /not available in your region/)
+          end
+
+          it 'raises a server error when the body is a bare JSON string' do
+            stub_request(:post, url).to_return(
+              status: 500,
+              body: 'The upstream provider returned an error.'.to_json,
+              headers: { 'Content-Type' => 'application/json' }
+            )
+
+            # Faraday 1 streams the error body away before on_complete, so only
+            # the HTTP status is available to build the error message.
+            expected = if major == '1'
+                         [RubyLLM::ServerError]
+                       else
+                         [RubyLLM::ServerError, /upstream provider returned an error/]
+                       end
+
+            expect do
+              chat.ask('Count from 1 to 3') { |chunk| chunk }
+            end.to raise_error(*expected)
+          end
+        end
+      end
+    end
   end
 end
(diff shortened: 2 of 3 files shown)


crmne:
Closing in favor of #840. That PR covers Chat Completions, Anthropic, and Gemini streaming errors and preserves the real HTTP status when the error value is a string instead of replacing it with a synthetic 500. Thanks for working on this.

PR #849, opened 2026-07-15
Raise a clear error when a provider returns no completion message

### LINKED ISSUE #847: [BUG] `Chat#add_message(nil)` crashes with an unexplained `NoMethodError`, obscuring the real cause
### Basic checks

### What's broken?

When an OpenAI-family provider returns a response body with no data or `choices[0].message`, such as an empty or content-filtered upstream response, `parse_completion_response` returns `nil`.
`Chat#complete_once` then calls `add_message(nil)` without a nil check:
```ruby
gems/ruby_llm-1.16.0/lib/ruby_llm/chat.rb:229:in `complete_once'
  → add_message(nil)
```

`Message#initialize` blindly calls `options.fetch(:role)` on the nil argument:

```ruby
# ruby_llm/message.rb:12
@role = options.fetch(:role).to_sym
```

This raises `NoMethodError: undefined method 'fetch' for nil`, which escapes as an unhandled exception and crashes the caller.

### How to reproduce

I'm not sure this is a useful repro, but...

```ruby
chat = RubyLLM::Chat.new(model: "some-openai-family-model")
chat.add_message(nil) 
```

I'm not sure how to trigger a provider response that is lacking a `choices[0].message`.

### Expected behavior

When a provider returns no completion message, RubyLLM should, at the bare minimum, raise a `RubyLLM::Error`. There's no detail to log or data to introspect in to, so it's not possible to determine if this is recoverable or not.

### What actually happened

In our case this happened deep inside `Chat#handle_tool_calls` → `complete` → `complete_once` during a real agentic tool-call loop against OpenRouter, when the upstream response had no message body:

```
gems/ruby_llm-1.16.0/lib/ruby_llm/message.rb:12:in `RubyLLM::Message#initialize': undefined method 'fetch' for nil
      @role = options.fetch(:role).to_sym
    from gems/ruby_llm-1.16.0/lib/ruby_llm/chat.rb:166:in `RubyLLM::Chat#add_message'
    from gems/ruby_llm-1.16.0/lib/ruby_llm/chat.rb:229:in `RubyLLM::Chat#complete_once'
    ...
    from gems/ruby_llm-1.16.0/lib/ruby_llm/chat.rb:340:in `RubyLLM::Chat#handle_tool_calls'
    from gems/ruby_llm-1.16.0/lib/ruby_llm/chat.rb:233:in `RubyLLM::Chat#complete_once'
```

Digging deeper,

1. `RubyLLM.chat` builds a chat with `provider: :openrouter` 

2. **`RubyLLM::Provider#complete` → `sync_response`** (`provider.rb:44-66, 274-279`)  `Chat#provider_completion` (`chat.rb:280`) calls `@provider.complete(...)`. Since no block is given (non-streaming), that hits `sync_response`, which POSTs to `completion_url` and hands the raw response straight to `parse_completion_response`.

3. **`OpenRouter` inherits `parse_completion_response` from `OpenAI::Chat` unchanged** (`providers/openrouter.rb:6` — `class OpenRouter < OpenAI`, and `providers/openai/chat.rb:54-61`):

   ```ruby
   def parse_completion_response(response)
     data = response.body
     return if data.empty?                       # fully empty body → nil
     raise Error.new(...) if data.dig('error', 'message')
     message_data = data.dig('choices', 0, 'message')
     return unless message_data                  # <-- no message in the response → nil
     ...
     Message.new(role: :assistant, ...)
   end
   ```

Two separate paths return `nil` here: a truly empty body, or a body that parses fine but has no `choices[0].message` (e.g. `finish_reason: "content_filter"` with no message payload, or a `choices` array that's empty/malformed). OpenRouter doesn't override this method, so whatever quirk in its response shape triggered this just falls through the OpenAI-shaped parser silently.

4. **`Chat#complete_once` blindly forwards that `nil`** (`chat.rb:222-229`):

   ```ruby
   def complete_once(&)
     response = provider_completion(&)   # <- nil
     ...
     add_message response                # <- add_message(nil)
   ```

5. **`Chat#add_message(nil)` → `Message.new(nil)`** (`chat.rb:165-169`,
   `message.rb:11-12`):

   ```ruby
   def add_message(message_or_attributes)
     message = message_or_attributes.is_a?(Message) ? message_or_attributes : Message.new(message_or_attributes)
   ```

Since `nil` isn't a `Message`, it calls `Message.new(nil)`. `Message#initialize(options = {})` has a default of `{}`, but that default only applies when no argument is passed — here `nil` is passed explicitly, so `options` is `nil`, and `options.fetch(:role)` blows up with the `NoMethodError`.

This bug is still present in `main` despite a refactor.

### Environment

- Ruby version: 4.0.5
- Provider: Openrouter
- RubyLLM version: 1.16.0

jonthedecepticon:
## What this does

Fixes #847.

When an OpenAI-family provider returns a response body without `choices[0].message` (e.g. an empty or content-filtered upstream response), `parse_completion_body` returns `nil` and `Chat` ended up calling `add_message(nil)`, which crashed deep inside `Message#initialize` with an unexplained `NoMethodError: undefined method 'fetch' for nil` — obscuring the real cause.

Two small guards:

- **Protocol seam** (`Protocol#parse_completion_response`): when `parse_completion_body` produces no message, raise `RubyLLM::Error.new('Provider returned no completion message', response:)` — mirroring the existing empty-body guard two lines above, and carrying the response so callers can introspect status/body as the issue requests.
- **Domain guard** (`Chat#coerce_message`): a direct `chat.add_message(nil)` (the issue's repro) now raises `ArgumentError, 'Message cannot be nil'` instead of the baffling `NoMethodError`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

## Tests

- Protocol-level spec in `protocol_spec.rb` alongside the existing empty-body examples: a protocol whose `parse_completion_body` returns `nil` now raises the new error.
- End-to-end spec in `chat_error_spec.rb`: a stubbed 200 response with `{"choices": []}` raises `RubyLLM::Error` with the response attached (previously `NoMethodError`).
- Direct-repro spec: `chat.add_message(nil)` raises `ArgumentError` with a clear message.

## AI-generated code

## API changes

- [ ] Breaking change
- [x] No API changes

### DIFF
--- a/lib/ruby_llm/chat.rb
+++ b/lib/ruby_llm/chat.rb
@@ -630,6 +630,8 @@ module RubyLLM
     end
 
     def coerce_message(message_or_attributes)
+      raise ArgumentError, 'Message cannot be nil' if message_or_attributes.nil?
+
       message = if message_or_attributes.respond_to?(:to_llm)
                   message_or_attributes.to_llm
                 else
--- a/lib/ruby_llm/protocol.rb
+++ b/lib/ruby_llm/protocol.rb
@@ -267,7 +267,10 @@ module RubyLLM
         raise Error.new('Provider returned an empty response body', response:)
       end
 
-      parse_completion_body(body, raw: response)
+      message = parse_completion_body(body, raw: response)
+      raise Error.new('Provider returned no completion message', response:) unless message
+
+      message
     end
   end
 end
--- a/lib/ruby_llm/protocols/chat_completions/chat.rb
+++ b/lib/ruby_llm/protocols/chat_completions/chat.rb
@@ -69,7 +69,7 @@ module RubyLLM
           raise Error.new(data.dig('error', 'message'), response: raw) if data.dig('error', 'message')
 
           message_data = data.dig('choices', 0, 'message')
-          return unless message_data
+          raise no_completion_message_error(data, raw) unless message_data
 
           usage = data['usage'] || {}
           thinking_tokens = thinking_tokens(usage)
@@ -94,6 +94,13 @@ module RubyLLM
             model: data['model'],
             raw: raw
           )
+        end
+
+        def no_completion_message_error(data, raw)
+          finish_reason = data.dig('choices', 0, 'finish_reason')
+          message = 'Provider returned no completion message'
+          message = "#{message} (finish_reason: #{finish_reason})" if finish_reason
+          Error.new(message, response: raw)
         end
 
         def input_tokens(usage)
--- a/spec/ruby_llm/chat_error_spec.rb
+++ b/spec/ruby_llm/chat_error_spec.rb
@@ -109,6 +109,18 @@ RSpec.describe RubyLLM::Chat do
         expect(e.response).to be_present
       end
     end
+
+    it 'surfaces the finish_reason when the provider gives one' do
+      stub_request(:post, 'https://api.deepseek.com/chat/completions').to_return(
+        status: 200,
+        body: { choices: [{ finish_reason: 'content_filter' }] }.to_json,
+        headers: { 'Content-Type' => 'application/json' }
+      )
+
+      expect { chat.ask('Hello') }.to raise_error(
+        RubyLLM::Error, 'Provider returned no completion message (finish_reason: content_filter)'
+      )
+    end
   end
 
   describe '#add_message with nil' do
--- a/spec/ruby_llm/chat_error_spec.rb
+++ b/spec/ruby_llm/chat_error_spec.rb
@@ -94,4 +94,28 @@ RSpec.describe RubyLLM::Chat do
       end
     end
   end
+
+  describe 'responses without a completion message' do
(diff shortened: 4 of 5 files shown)


lucasmo:
Thanks for this! It solves the problem but gives no indication of the cause. I'm not familiar enough with wire protocol to understand why this would happen - perhaps the provider is returning some kind of error? Would it be possible to surface some kind of `finish_reason` to help diagnose why this happens? (again, no idea of wire protocol here)

jonthedecepticon:
In the OpenAI wire format the finish_reason lives on the choice, so when the provider returns a choice without a message (the content-filter shape) we can surface it. Pushed a commit that appends it: `Provider returned no completion message (finish_reason: content_filter)`. When the provider returns an empty `choices` array there is no finish_reason to show, but the error still carries the full HTTP response, so `e.response.body` shows whatever the provider actually sent. @lucasmo

crmne (approved):

PR #867, opened 2026-08-10
Detect transcription-capable models

### LINKED ISSUE #864: [FEATURE] Add `supports_transcription?` to model capability detection
### Scope check

### Due diligence

In a multi-model/provider environment, `Model::Info` has no way to ask whether a model supports transcription. This means we have to hardcode the allow-lists or else end up with runtime `RubyLLM.transcribe` errors.

In [Consul democracy's PR](https://github.com/consuldemocracy/consuldemocracy/pull/6481), we are using RubyLLM as the backing for Speech-to-Text service, enabling institutions freedom of choice over the provider / model backend, and not having to worry about vendor lock-in. Transcription detection would make supporting this feature much easier, as we don't have to poke around gem's internals / hardcode the list of capable models.

I'm happy to implement a small PR for the proposed solution.

### Proposed solution

**Public API** on `RubyLLM::Model::Info`:

- Add `transcription` to the dynamic capability predicates → `transcription?`
- Add `supports_transcription?` as an alias of `transcription?` (same style as
  `supports_functions?`)

**How the capability is populated:**

1. **OpenAI** - Helpers like `supports_vision?(model_id)` / `supports_structured_output?(model_id)` already return true/false based on `model_family(model_id)`, so we can do similarly here for `'whisper'`, `'gpt4o_transcribe'`, `'gpt4o_mini_transcribe'`.
2. **Gemini** - Derive `'transcription'` when `modalities.input` includes `audio` and `modalities.output` includes `text` (using `models_dev_capabilities`, mirroring how vision is derived from image/video/pdf input).
3. **Mistral** already emits `'transcription'` for `/voxtral.*transcribe/`, this is what prompted me to open the proposal in the first place

Capability detection already lives in the gem - `supports_vision?`, `streaming?`, `structured_output?`... - to name a few.
Mistral already tags some `voxtral.*transcribe` models already, and detection using family / capabilities for other providers should be straightforward enough, as shown above

crmne:
## What this does

Uses the existing `Model#supports?(:transcription)` API instead of adding another predicate.

- Tags Whisper and GPT-4o transcribe families from OpenAI model metadata
- Derives transcription for audio-input/text-output Gemini models on Gemini and Vertex AI
- Avoids false positives for other providers and multimodal embedding models
- Updates the bundled registry and generated model documentation

Fixes #864

## Type of change

- [x] New feature

## Quality check

- [x] Full pre-commit hooks pass
- [x] Registry schema and focused model specs pass
- [x] No public API changes
- [x] AI-assisted and reviewed

### DIFF
--- a/docs/_reference/available-models.md
+++ b/docs/_reference/available-models.md
@@ -2,7 +2,7 @@
 layout: default
 title: Available Models
 nav_order: 2
-description: Browse 1183 AI models across 11 remote providers. Updated 2026-07-15.
+description: Browse 1183 AI models across 11 remote providers. Updated 2026-08-10.
 redirect_from:
   - /guides/available-models
 ---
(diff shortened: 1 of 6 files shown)


crmne:
Closing this PR so the already-tested worktree commit can be merged directly into main, as requested.

PR #878, opened 2026-08-17
Add MiniMax text to video generation

octo-patch:
Reason: MiniMax has no video generation implementation, so `RubyLLM.animate` cannot reach its text-to-video endpoints.

RubyLLM already ships the whole asynchronous video lifecycle: `animate`, `animate_later`, `VideoJob#refresh!`, `wait!`, and `video`. This adds the MiniMax dialect behind those seams.

MiniMax renders video behind two API versions, and the model id decides which one a job uses for submission, polling, and download:

* `MiniMax-H3` uses `POST /v2/video_generation` with a multimodal `content` array, `resolution`, and `duration`, is polled at `GET /v2/query/video_generation/{task_id}`, and reports the finished clip as a hosted URL on `task.content.url`.
* `MiniMax-Hailuo-2.3`, `MiniMax-Hailuo-2.3-Fast`, `MiniMax-Hailuo-02`, `T2V-01-Director`, and `T2V-01` use `POST /v1/video_generation` with a flat `prompt`, are polled at `GET /v1/query/video_generation?task_id=...`, and report a `file_id` that is exchanged for a download URL at `GET /v1/files/retrieve`.

## Changes

* `lib/ruby_llm/providers/minimax.rb`: new provider with `minimax_api_key` and `minimax_api_base`, bearer auth, and a `VideoGeneration` protocol. `video_api_base` rebuilds each API version against the configured host, so pointing `minimax_api_base` at another regional host moves both versions with it. Models are assumed to exist, since MiniMax has no model listing endpoint.
* `lib/ruby_llm/providers/minimax/videos.rb`: the six video seams. v2 statuses `queued`, `running`, `succeeded`, `failed`, and `cancelled` and v1 statuses `Preparing`, `Queueing`, `Processing`, `Success`, and `Fail` map onto `:pending`, `:completed`, and `:failed`; failures surface `task.error.message` on v2. A non-zero `base_resp.status_code` becomes a `RubyLLM::Error` even on HTTP 200, which is how v1 reports rejected requests. Durations, aspect ratios, and every other request field travel through `provider_options`.
* `lib/ruby_llm.rb`: registers the provider and its inflection.
* `spec/ruby_llm/provider_spec.rb`: the API base table entry the "covers every registered provider" example requires.
* `spec/ruby_llm/providers/minimax/videos_spec.rb`: 24 examples over both versions, covering endpoint selection for the whole catalog, the regional host, request payloads, status parsing, downloads, and error paths.
* Docs: the provider row in the video generation guide, its `provider_options` vocabulary, the regional host, and the configuration keys.

Reference images and first/last frame inputs are deliberately not part of this change: `with:` keeps raising `RubyLLM::UnsupportedAttachmentError`, so image-to-video can land separately without reworking these seams.

## Checks

* `bundle exec rspec spec/ruby_llm/providers/minimax/videos_spec.rb spec/ruby_llm/provider_spec.rb spec/ruby_llm/provider_contract_spec.rb spec/ruby_llm/protocol_contract_spec.rb` — 216 examples, 0 failures.
* `bundle exec rubocop` on every file this change touches — no offenses.
* Drove `RubyLLM.animate` and `RubyLLM.animate_later` against stubbed MiniMax endpoints for both API versions: submit, poll, download, `to_blob`, a task that fails mid-render, and the regional host.

No live cassette is recorded here, so no row was added to `VIDEO_GENERATION_MODELS`.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -50,6 +50,7 @@ RSpec/SpecFilePathFormat:
     DeepSeek: deepseek
     ElevenLabs: elevenlabs
     GPUStack: gpustack
+    MiniMax: minimax
     OpenAI: openai
     OpenRouter: openrouter
     VertexAI: vertexai
--- a/docs/_core_features/video-generation.md
+++ b/docs/_core_features/video-generation.md
@@ -100,6 +100,7 @@ These providers generate video through RubyLLM:
 | xAI | `grok-imagine-video`, `grok-imagine-video-1.5` | Clips up to 15 seconds, priced per second |
 | OpenRouter | `x-ai/grok-imagine-video`, `google/veo-3.1-lite`, `openai/sora-2-pro`, and the rest of its video catalog | One API across many video models |
 | Azure OpenAI | Sora deployments such as `sora-2` | Uses your deployment name as the model id |
+| MiniMax | `MiniMax-H3`, `MiniMax-Hailuo-2.3`, `MiniMax-Hailuo-2.3-Fast`, `MiniMax-Hailuo-02`, `T2V-01-Director`, `T2V-01` | Clips of 4 to 15 seconds, priced per second |
 
 Refer to the [Working with Models Guide]({% link _reference/models.md %}) for finding and filtering models, and [Model Resolution]({% link _reference/model-resolution.md %}) for how a model name and provider resolve.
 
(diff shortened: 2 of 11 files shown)


crmne:
Thanks, but closing. MiniMax stays a community provider. The acceptance bar and the community-gem path are both in [CONTRIBUTING.md](https://github.com/crmne/ruby_llm/blob/main/CONTRIBUTING.md), which also asks for an approved issue before a provider PR.

You don't need core for this any more. `bundle exec ruby_llm provider-gem minimax` scaffolds a provider gem you own and release yourself, see [Custom Providers and Protocols](https://rubyllm.com/custom-providers/). One rule to carry over: wire formats belong in a protocol, not a provider. CI flags it here.

The generator is new. Please open issues if it gets in your way.

Issue #872, opened 2026-08-13
[BUG] Agent delegates every Chat callback except before_request


danielefrisanco:
### Basic checks

- [x] I searched existing issues - this hasn't been reported
- [x] I can reproduce this consistently
- [x] This is a RubyLLM bug, not my application code

### What's broken?

`Agent`'s `def_delegators` list covers six of the seven Chat callback
registrations — `before_message`, `after_message`, `before_tool_call`,
`after_tool_result`, `before_fallback`, `after_fallback` — but not
`before_request`, so `agent.before_request { ... }` raises `NoMethodError` while
its six siblings work.

This isn't blocking anything: an agent keeps the chat you hand it, so configuring
the chat before constructing the agent works fine (workaround under "What actually
happened"). It's the asymmetry I'm reporting, since the class documentation
describes the delegating behaviour without an exception:

- "Agent instances delegate the Chat API (#ask, #complete, #with_tools, and so
  on) to the wrapped chat, which is available via #chat."
- On `GUARDED_OPERATIONS`: "The chat operations an agent instance runs through
  its ::rescue_from handlers. **Every other Chat method is delegated untouched.**"

Calling each callback on an agent, on `main`:

| callback | `respond_to?` | returns the wrapped chat |
|---|---|---|
| `before_message` | true | yes |
| `after_message` | true | yes |
| `before_tool_call` | true | yes |
| `after_tool_result` | true | yes |
| `before_fallback` | true | yes |
| `after_fallback` | true | yes |
| `before_request` | **false** | — |

It reads like an oversight rather than a decision: the commit that added
`before_request` touched `chat.rb` and the protocols but not `agent.rb`, and the
existing `'delegates callback hooks to the underlying chat'` example enumerates
the six callbacks that existed at the time.

The argument against delegating it, in case it was deliberate: `before_request`
edits the provider's wire format, and agents support `with_fallbacks` across
providers, so a hook written for one payload shape could misfire after a
fallback. If that's the reason, a line in the Agent docs would settle it and I'm
happy to send that instead.

Either way I have a patch ready — one word in `agent.rb` plus a test that extends
the existing delegation example rather than adding a spec file (two files, eight
lines). Glad to open the PR, or to close this if the omission is intended.

### How to reproduce

No network needed — `Chat#render` applies the hooks without sending.

```ruby
require 'ruby_llm'
RubyLLM.configure { |c| c.anthropic_api_key = ENV['ANTHROPIC_API_KEY'] }

[...]

PR #884, opened 2026-08-19
Delegate before_request, count_tokens and render to the chat

### LINKED ISSUE #883: [BUG] acts_as_chat records don't expose before_request, count_tokens, or render
### Basic checks

### What's broken?

`RubyLLM::Chat` exposes `before_request`, `count_tokens`, and `render`, but records using `acts_as_chat` expose none of them, so calling any of the three on a persisted chat raises `NoMethodError`.

`CHAT_DELEGATES` in `lib/ruby_llm/active_record/chat_methods.rb:212` lists 19 methods and omits all three. There is no `method_missing` fallback, so nothing picks them up.

Current state across the three chat entry points:

| method | `RubyLLM::Chat` | `Agent` | `acts_as_chat` |
|---|---|---|---|
| `before_request` | yes | yes | **no** |
| `count_tokens` | yes | **no** | **no** |
| `render` | yes | **no** | **no** |

This is the same omission as #872 ("Agent delegates every Chat callback except `before_request`"), which was fixed in 48a7e751, but that commit only touched `lib/ruby_llm/agent.rb`, so the Rails integration still has the gap.

Two docs point users at methods that don't exist on a persisted chat:

- `docs/_advanced/rails-advanced-config.md:126` sends Rails users to the
  `before_request` hook for provider-specific request payloads.
- `docs/_core_features/cost-and-usage-tracking.md:83-104` documents
  `chat.count_tokens` for enforcing a context budget before sending.

What ac

Environment
- RubyLLM: main (b44504a9)
- Ruby 3.2.2
- Rails 8.1.3.1

### How to reproduce

1. In a Rails app with the RubyLLM install generator run:

```ruby
class Chat < ApplicationRecord
  acts_as_chat
end
```

2. In a console:

```ruby
chat = Chat.create!(model: "gpt-4.1-nano")

chat.respond_to?(:before_request)   # => false
chat.respond_to?(:count_tokens)     # => false
chat.respond_to?(:render)           # => false

chat.before_request { |payload| payload[:metadata] = { user_id: "u-1" } }
# NoMethodError: undefined method `before_request'

chat.count_tokens("Summarize this contract.")
# NoMethodError: undefined method `count_tokens'
```

3. The same calls on a plain chat work:

```ruby
RubyLLM.chat(model: "gpt-4.1-nano").respond_to?(:before_request)  # => true
```

### Expected behavior

A record using `acts_as_chat` should expose the same request hooks and inspection methods as `RubyLLM::Chat`  matching the behavior `Agent` got in 48a7e751.

Note for whoever fixes this: only `before_request` fits `CHAT_DELEGATES`. That
loop returns `self` so calls chain, which would discard the return values of
`count_tokens` and `render`. Those two need value-returning delegation like the
existing `generate` (line 340) and `step` (line 357).

### What actually happened

`NoMethodError` for all three. `respond_to?` returns `false`, and there is no
`method_missing` in `ChatMethods` to catch them.

### Environment

- RubyLLM: main (b44504a9)
- Ruby 3.2.2
- Rails 8.1.3.1

toluola:
## What this does

Fixes #883.

`RubyLLM::Chat` exposes `before_request`, `count_tokens` and `render`, but records using `acts_as_chat` expose none of them, and `Agent` is missing `count_tokens` and `render`. Each call raises `NoMethodError`.

| method | `RubyLLM::Chat` | `Agent` | `acts_as_chat` |
|---|---|---|---|
| `before_request` | yes | yes (#872) | **no** |
| `count_tokens` | yes | **no** | **no** |
| `render` | yes | **no** | **no** |

This is the same omission as #872, fixed in 48a7e751 — but that commit only touched `agent.rb`, so the Rails integration kept the gap. The issue's table also shows `count_tokens` and `render` missing from `Agent`, so this PR closes both rows.

Three docs already describe the behavior that doesn't work:

- `docs/_advanced/rails-advanced-config.md:126` sends Rails users to the `before_request` hook for provider-specific payloads.
- `docs/_core_features/cost-and-usage-tracking.md:83-104` documents `chat.count_tokens` for enforcing a context budget.
- `docs/_advanced/agents.md:301` states agents "delegate the full `RubyLLM::Chat` instance API".

No doc changes are included because the docs already say what this makes true.

### Implementation note

`before_request` joins `CHAT_DELEGATES`. `count_tokens` and `render` cannot: that loop returns `self` so calls chain, which would swallow their return values. They delegate like the existing `generate` and `step`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

  - [x] All tests pass: `bundle exec rspec --tag ~live` (2642 examples, 0 failures) and `bundle exec archspec check`

Both new spec groups fail with `NoMethodError` when the library change is reverted, so they are load-bearing rather than restating the implementation. No provider behavior changed, so no cassettes were re-recorded.

## AI-generated code

## API changes

- [ ] Breaking change
- [ ] No API changes

### DIFF
--- a/lib/ruby_llm/active_record/chat_methods.rb
+++ b/lib/ruby_llm/active_record/chat_methods.rb
@@ -221,7 +221,7 @@ module RubyLLM
         with_max_output_tokens with_thinking with_citations with_caching
         with_end_user with_compaction
         with_provider_options with_headers with_schema
-        before_message after_message before_tool_call after_tool_result
+        before_request before_message after_message before_tool_call after_tool_result
         before_fallback after_fallback
       ].freeze
 
@@ -310,6 +310,27 @@ module RubyLLM
       def cost
         records = ruby_llm_usages.to_a
         RubyLLM::Cost.aggregate(records.map(&:cost), complete: records.all?(&:usage_available?))
+      end
+
+      # Returns the number of input tokens the next request would carry,
+      # counted by the provider over the persisted conversation. Pass a
+      # message to count it alongside that history without persisting it.
+      # See RubyLLM::Chat#count_tokens.
+      #
+      #   chat.count_tokens "Summarize this contract."
+      #
+      def count_tokens(...)
+        to_llm.count_tokens(...)
+      end
+
+      # Returns the request payload this chat would send to the provider for
+      # its next completion, with #before_request hooks applied. See
+      # RubyLLM::Chat#render.
+      #
+      #   chat.render[:messages]
+      #
+      def render(...)
+        to_llm.render(...)
       end
 
       # Persists +message+ as a user message, then runs the completion loop
--- a/lib/ruby_llm/agent.rb
+++ b/lib/ruby_llm/agent.rb
@@ -768,7 +768,7 @@ module RubyLLM
                    :with_headers, :with_schema, :with_fallbacks, :before_request, :before_message, :after_message,
                    :before_tool_call, :after_tool_result, :before_fallback, :after_fallback, :each, :complete?,
                    :cancel!, :cancelled?, :approve!, :deny!, :awaiting_approval?, :pending_approvals,
-                   :add_message, :add_completion, :tokens, :cost
+                   :add_message, :add_completion, :tokens, :cost, :count_tokens, :render
 
     ##
     # :method: ask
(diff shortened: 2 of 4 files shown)


toluola:
@crmne reminder to review this. Thanks.

PR #903, opened 2026-09-01
[FEATURE] Support block/proc for dynamic context setup in Agent

### LINKED ISSUE #900: [FEATURE] Support block/proc for dynamic context setup in Agent
### Scope check

### Due diligence

`RubyLLM::Agent` has helpful macros for defining chat properties used (`model`, `params`, etc). There is a macro for `context`, but unlike the other macros, it only handles static values, and not procs/blocks evaluated at runtime. For cases where units of work are defined as subclasses of `RubyLLM::Agent`, there's no easy way to inject instance-specific details (eg: retry limits, timeout limits, per-tenant API keys).  

### Proposed solution

Following the conventions as the other macros, make the `context` macro take either a static value or a proc, which can read values from input and other sources at runtime. Happy to provide a PR if it helps.

Making the `context` macro follow the same value-or-proc-based pattern as the other macros makes this work without having disassemble the Agent into its constituent parts in applications (which would be a shame, it's a nicely designed class).

mikemikimike:
Closes #900

Summary:
- Allow Agent#context to accept a static Context or a deferred zero-argument block/Proc.
- Resolve deferred contexts from declared inputs before building plain Agent chats and instances.
- Reuse the resolved Context during configuration so each build evaluates the deferred context once.
- Preserve the existing class-definition form `context { |config| ... }` for static isolated contexts.
- Cover deferred context behavior for plain Ruby and Rails create/find flows.

Compatibility notes:
- The runtime gemspec pins `json < 3` because the supported Faraday/Rails stack still uses JSON 2 positional options; the development and appraisal Gemfiles stay aligned with that runtime constraint.
- The workflow-spec environment overrides preserve the repository's bundle settings when the workflow shell runs from a temporary directory and clear inherited Ruby options so the harness remains isolated.

Validation:
- `bundle exec rspec spec/ruby_llm/agent_dsl_spec.rb --tag ~live` (33 examples, 0 failures)
- `bundle exec rubocop lib/ruby_llm/agent.rb spec/ruby_llm/agent_dsl_spec.rb spec/ruby_llm/agent_rails_spec.rb spec/tasks/models_workflow_spec.rb`
- `bundle exec archspec check`
- `git diff --check`


crmne:
The concept in #900 is accepted for after 2.0. Keeping this PR open for that work; implementation review and any rebase can happen after the release.

PR #911, opened 2026-09-09
Return {} for empty-string tool-call arguments in the interactions protocol

pcbeingused333:
### Problem

`Protocols::Interactions::Tools.parse_interaction_arguments` runs `JSON.parse` on any `String`:

```ruby
def parse_interaction_arguments(arguments)
  arguments.is_a?(String) ? JSON.parse(arguments) : arguments || {}
rescue JSON::ParserError => e
  raise ToolCallParseError.new(finish_reason: :tool_calls), cause: e
end
```

A `function_call` step whose `arguments` is an empty string (`""`) makes `JSON.parse("")` raise, so the whole run fails with `ToolCallParseError` for what is a valid zero-argument tool call.

The sibling parser already handles this — `Protocols::ChatCompletions::Tools.parse_tool_call_arguments`:

```ruby
if arguments.nil? || arguments.empty?
  {}
else
  JSON.parse(arguments)
end
```

So the two tool-call parsers disagree on the same input.

### Fix

Guard empty (and nil) arguments in `parse_interaction_arguments` the same way, before `JSON.parse`.

```ruby
return {} if arguments.nil? || (arguments.is_a?(String) && arguments.empty?)
```

### Tests

New `spec/ruby_llm/protocols/interactions/tools_spec.rb`: empty string → `{}`, missing key → `{}`, Hash passthrough, valid JSON object, malformed JSON → `ToolCallParseError`. The empty-string case raises on `main`.

`bundle exec rspec spec/ruby_llm/protocols/interactions{_spec.rb,/tools_spec.rb} spec/ruby_llm/protocols/chat_completions/tools_spec.rb` — green; `rubocop` clean on both files.

### DIFF
--- a/lib/ruby_llm/protocols/interactions/tools.rb
+++ b/lib/ruby_llm/protocols/interactions/tools.rb
@@ -31,7 +31,9 @@ module RubyLLM
         end
 
         def parse_interaction_arguments(arguments)
-          arguments.is_a?(String) ? JSON.parse(arguments) : arguments || {}
+          return {} if arguments.nil? || (arguments.is_a?(String) && arguments.empty?)
+
+          arguments.is_a?(String) ? JSON.parse(arguments) : arguments
         rescue JSON::ParserError => e
           raise ToolCallParseError.new(finish_reason: :tool_calls), cause: e
         end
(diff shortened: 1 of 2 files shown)


crmne:
Integrated in https://github.com/crmne/ruby_llm/commit/c4b3afc3. Thanks for the fix and regression specs.

PR #915, opened 2026-09-10
Fix instruction declarations not inheriting to Agent subclasses

allenjiji:
## Problem

Every other piece of Agent class configuration — tools, server tools, tool options, provider options, headers, inputs, fallbacks, rescue handlers, thinking, temperature — is copied to subclasses on `inherited`. Instruction declarations are not:

```ruby
class BaseAgent < RubyLLM::Agent
  instructions "Be terse.", cache_until_here: true
end

class ChildAgent < BaseAgent; end

BaseAgent.instructions
# => [{value: "Be terse.", append: false, persist: true, cache_until_here: true}]
ChildAgent.instructions
# => []
```

Verified on 2.0.0.rc1, 2.0.0.rc2, and current main; also verified ivar-by-ivar that every other `DUPED_INHERITED_CONFIG` and `COPIED_INHERITED_CONFIG` entry does inherit correctly — instructions is the only one that doesn't.

The failure is silent in the common Rails setup: when the subclass has a conventional prompt template, `instructions_config` falls back to the template default, which hardcodes `append: false, persist: true, cache_until_here: false`. Instructions still appear, so nothing looks broken, while every option on the base class's declaration quietly vanishes — in our app, an Anthropic cache boundary declared on an `ApplicationAgent` base class evaporated in every concrete agent, a cost regression with no error.

## Root cause

`DUPED_INHERITED_CONFIG` lists `:@instructions`, an ivar nothing writes — the `instructions` macro stores declarations in `@instruction_declarations` (agent.rb:186–195, reader at :781–783). `copy_inherited_config_to` therefore dups an always-empty ivar and never copies the real declarations.

## Fix

Rename the table entry to `:@instruction_declarations`. The existing `.dup` semantics keep subclass declarations independent of the parent (a subclass appending doesn't mutate the base class), matching how `tools` accumulates.

Two specs added under the existing `describe 'inheritance'` block: declarations (with their options) reach the subclass, and a subclass declaration stays out of the parent.

Ruby 4.0.6 / Rails 8.1.3; found upgrading a Rails app from 1.16.0. Full non-live suite passes locally (3,135 examples, 0 failures).

### DIFF
--- a/docs/_advanced/agents.md
+++ b/docs/_advanced/agents.md
@@ -254,7 +254,7 @@ RubyLLM looks for:
 
 * `app/prompts/work_assistant/instructions.txt.erb`
 
-If the file exists, it is rendered and used as instructions automatically. If it does not exist and you did not set `instructions`, the agent starts without system instructions. To require a prompt and fail loudly when it is missing, reference it explicitly:
+If the file exists, it is rendered and used as instructions automatically. If it does not exist and you did not set `instructions` on the class or an ancestor, the agent starts without system instructions. Subclasses inherit instruction declarations, and inherited declarations suppress the conventional prompt the same way the class's own declarations do; declare `instructions(append: true) { prompt("instructions") }` to use both.
 
 ```ruby
 class WorkAssistant < RubyLLM::Agent
--- a/lib/ruby_llm/agent.rb
+++ b/lib/ruby_llm/agent.rb
@@ -54,7 +54,7 @@ module RubyLLM
       :@fallbacks => [],
       :@fallback_options => {},
       :@rescue_handlers => [],
-      :@instructions => []
+      :@instruction_declarations => []
     }.freeze
     # Simple value options: a class-level getter/setter macro whose value the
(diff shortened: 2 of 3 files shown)


allenjiji:
This fix implicitly settles a precedence question. Before it, a subclass with its own conventional template (`app/prompts/support_agent/instructions.txt.erb`) that declared nothing used that template — but only because declaration list was wrongly empty. With inheritance working, the subclass carries the parent's declarations, so `instructions_config` uses those and the template fallback does not run.

If inherited declarations should instead compose with the subclass's template automatically, that is a design decision beyond this bug fix.
I'm happy to open an issue for it if needed.

crmne:
Integrated in https://github.com/crmne/ruby_llm/commit/6186b390 with fallback precedence: the child's declarations, then its conventional prompt, then inherited declarations. Empty child prompts suppress inheritance. Inherited persistence and cache settings are preserved when used, with regression coverage for plain Ruby and Rails.

PR #943, opened 2026-09-18
Honor model mappings that follow a generator class option

### LINKED ISSUE #942: [BUG] ruby_llm:upgrade ignores model mappings when a class option comes first
### Basic checks

### What's broken?

`ruby_llm:upgrade` silently drops the model mappings you pass on the command line
whenever a class option appears before them. The generator falls back to the
default `Chat` / `Message` / `Model` / `ToolCall` names and writes migrations
against tables the application does not have.

Affected forms (Thor gives the generator an empty positional list):

```bash
bin/rails generate ruby_llm:upgrade --mode copy chat:AI::Chat message:AI::Chat::Message
bin/rails generate ruby_llm:upgrade --mode=copy chat:AI::Chat message:AI::Chat::Message
bin/rails generate ruby_llm:upgrade --phase prepare chat:AI::Chat message:AI::Chat::Message
bin/rails generate ruby_llm:upgrade --mode rename chat:AI::Chat message:AI::Chat::Message
```

Putting the mappings first works:

```bash
bin/rails generate ruby_llm:upgrade chat:AI::Chat message:AI::Chat::Message --mode copy
```

The upgrade guide tells you to pass the same mappings and `--mode` for every
phase without fixing the order, so this is easy to hit. The same
`argument :model_mappings, type: :array, default: []` declaration is used by
`ruby_llm:install`, `ruby_llm:chat_ui`, and `ruby_llm:tool`.

### How to reproduce

1. Create a Rails application with custom, namespaced chat models, e.g.
   `AI::Chat`, `AI::Chat::Message`, `AI::LLMModel`, `AI::Chat::ToolCall`, and a
   RubyLLM 1.16 schema (`ai_chats`, `ai_chat_messages`, `ai_llm_models`,
   `ai_chat_tool_calls`).

2. Generate the upgrade:

   ```bash
   bin/rails generate ruby_llm:upgrade --mode copy \
     chat:AI::Chat \
     message:AI::Chat::Message \
     model:AI::LLMModel \
     tool_call:AI::Chat::ToolCall
   ```

3. The generator reports the default names and writes migrations for the wrong
   tables:

   ```text
   models  chat:Chat -> chats, message:Message -> messages, model:Model -> models, tool_call:ToolCall -> tool_calls
   ```

   ```ruby
   # db/migrate/..._prepare_ruby_llm_v2_upgrade.rb
   move_table(:models, :ruby_llm_models)
   move_table(:tool_calls, :ruby_llm_tool_calls)
   ```

   ```ruby
   # config/initializers/ruby_llm_upgrade.rb
   RubyLLMUpgrade.install(chat: Chat, message: Message, tool_call: tool_call)
   ```

4. Run `bin/rails db:migrate`.

### Expected behavior

The mappings on the command line are honored wherever `--mode` / `--phase`
appear, so the generator resolves `AI::Chat -> ai_chats` and writes
`move_table(:ai_llm_models, :ruby_llm_models)`.

### What actually happened

The mappings were ignored with no error at generation time. `bin/rails db:migrate`
then aborts on boot with an error that does not point at the cause:

```text
NameError: uninitialized constant Chat

  RubyLLMUpgrade.install(chat: Chat, message: Message, tool_call: tool_call)
config/initializers/ruby_llm_upgrade.rb:12:in 'block in <top (required)>'
```

If the application happens to define a `Chat` constant, the prepare migration
then fails on `validate_table_move(:models, :ruby_llm_models)`.

Root cause: `Thor::Arguments.split` collects positional arguments only until the
first token matching `/^-/`, and `Thor::Group.dispatch` uses that split
(`thor-1.5.0/lib/thor/group.rb:234`). With `--mode copy chat:AI::Chat …` the
positional list is empty. `upgrade_generator.rb:32` declares
`argument :model_mappings, type: :array, default: []`, so an empty list is
accepted silently and `parse_model_mappings` starts from
`MODEL_MAPPING_DEFAULTS`. A required argument would have raised instead.

### Environment

- Baseline commit for the reported behavior: `e7bd2c9a1a7fa49870bf4ed26f4e1480ad098e80`
- Original report environment: Ruby 4.0.2, Rails 8.1.3.1; the report also listed Rails 7.1.6
- Current local validation: Ruby 3.3.5, Rails 8.1.3.1, SQLite
- Provider: N/A
- PostgreSQL was not exercised in the current validation; the failure is in Thor argument splitting before database access
- Mode: both `--mode rename` and `--mode copy`
- Phase: default multi-phase and every `--phase`
- Model mapping: custom namespaced (`chat:AI::Chat message:AI::Chat::Message model:AI::LLMModel tool_call:AI::Chat::ToolCall`)
- Inflections: `inflect.acronym "AI"`, `inflect.acronym "LLM"`
- OS: macOS arm64

## Scope

This starts from a supported RubyLLM 1.16 setup and follows the documented 2.0
upgrade path. It is a command-line parsing defect, not an application
configuration problem: the same mappings resolve correctly when they precede the
options.

## Local validation

- `Thor::Arguments.split` reproduces the empty positional list for the affected command order.
- `bundle exec rspec spec/ruby_llm/generators/upgrade_generators_spec.rb`: 12 examples, 0 failures.
- `bundle exec rspec --tag generator`: 95 examples, 0 failures with the RubyLLM 1.16 compatibility gem.
- The shared helper also has regression coverage for dasherized option names, option aliases, and boolean options.
- PostgreSQL and a full production-style multi-phase database migration were not run; the existing generator suite does cover UUID paths on SQLite.

yorzi:
## What this does

Fixes #942

`ruby_llm:upgrade` silently ignored the model mappings passed on the command line
whenever a class option preceded them:

```bash
bin/rails generate ruby_llm:upgrade --mode copy chat:AI::Chat message:AI::Chat::Message
# => models  chat:Chat -> chats, message:Message -> messages, ...
```

The generator fell back to the default model names and wrote migrations against
tables the application does not have, so `db:migrate` later aborted with an
unrelated-looking `NameError: uninitialized constant Chat` from the generated
compatibility initializer.

Thor collects positional arguments only until the first switch
(`Thor::Arguments.split`), and `Thor::Group.dispatch` uses that split, so
`--mode copy chat:AI::Chat …` yields an empty positional list. Because
`model_mappings` is declared with `default: []`, the empty list is accepted
silently and mapping resolution starts from the defaults.

`GeneratorHelpers` now reorders the incoming command line before Thor parses it,
moving the mappings ahead of the switches while keeping each switch with its own
value:

```ruby
["--mode", "copy", "chat:AI::Chat"] -> ["chat:AI::Chat", "--mode", "copy"]
```

The override lives in `GeneratorHelpers`, so `ruby_llm:install`,
`ruby_llm:chat_ui`, and `ruby_llm:tool` get the same fix: they share the
`argument :model_mappings, type: :array, default: []` declaration.

Booleans (`--force`, `--quiet`, `--discard-incomplete-tool-calls`) never consume
the next token, `--mode=copy` is self-contained, and unknown switches leave the
following token positional so mappings are never swallowed. Argument order is
unchanged when the mappings already come first.

### Regression coverage

`spec/ruby_llm/generators/upgrade_generators_spec.rb` — "resolves mappings that
follow a class option on the command line" drives the real CLI entry point
(`UpgradeGenerator.start`) and asserts the prepare migration targets
`ai_llm_models` and `ai_chat_tool_calls`. The direct Thor split reproduces the
pre-fix failure mode, and the example passes after the fix.

`spec/ruby_llm/generators/generator_helpers_spec.rb` covers dasherized option
names, short option aliases, boolean options, equals-form options, and unknown
switches. The InstallGenerator, ChatUIGenerator, and ToolGenerator specs each
exercise mappings after options through their real Rails generator entry point.

### Neighbor matrix

| Case | Result |
|---|---|
| Rails 8.1 generator suite, including the new shared-generator regressions | PASS — 97 examples, 0 failures |
| Rails 7.1 / 7.2 / 8.0 targeted compatibility | NOT RUN for the new shared-generator cases; the older fixture uses `load_defaults 8.1` |
| UUID generator/migration coverage on SQLite | PASS within the 97-example generator suite; not PostgreSQL |
| PostgreSQL 17 real adapter and full production-style copy-mode migration | NOT RUN |
| `bundle exec rspec --tag '~live'` | PASS — 3272 examples, 0 failures with loopback access enabled |
| `bundle exec rubocop --cache false` | PASS — 703 files, 0 offenses |
| `bundle exec archspec check` | PASS — 327 files, 1491 constants, 24293 facts checked |
| `CI=1 bundle exec overcommit --run --diff origin/main` | PASS — TrailingWhitespace, Gitleaks, RuboCop, ArchSpec, AppraisalUpdate, and RSpec |
| GitHub Actions [test / Test (4.0, rails-8.1)](https://github.com/crmne/ruby_llm/actions/runs/35322102290/job/105526689900?pr=943) | The linked run failed because the new namespaced ChatUI fixture touched an empty `app/models/billing/message.rb`; the fixture now defines `Billing::Message`, but the follow-up has not been pushed or re-run remotely |
| Flay on `lib/generators` | 138 observed; unchanged versus the parent was not independently established |

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

- [x] I ran the complete `overcommit --run` gate in the CI-compatible environment — `CI=1 bundle exec overcommit --run --diff origin/main` passed all configured hooks, including AppraisalUpdate and RSpec
  - [x] For provider changes: Re-recorded VCR cassettes with `bundle exec rake vcr:record[provider_name]` — N/A, no provider changes
  - [x] All non-live local tests pass: `bundle exec rspec --tag '~live'` passed with 3272 examples and 0 failures
  - [ ] Full local RSpec hook — one unrelated VertexAI live example still requires local ADC credentials; this change does not touch providers
- [x] I updated documentation if needed — no documentation change was needed; the
      documented command now behaves as written

## AI-generated code

## API changes

- [ ] Breaking change
- [x] No API changes


crmne:
Found one edge case: Thor accepts `--discard-incomplete-tool-calls false`, but the reordering moves `false` into the model mappings and turns the flag on. A previously valid command now fails validation. Could we preserve explicit boolean values and add a regression test for this?

yorzi:
Fixed argument reordering so `--discard-incomplete-tool-calls` false retains its explicit value instead of enabling the flag and treating `false` as a model mapping.

Also fixed related handling of boolean aliases, grouped short flags, `negated/skip_*` options, and the `--` end-of-options boundary.

Added regression coverage, including actual upgrade generation with namespaced models. All 3,175 non-live tests and generator cases passed after resolving missing test dependencies.

PR #949, opened 2026-09-21
Drop foreign provider raw_content on replay

### LINKED ISSUE #948: [BUG] Provider-native `raw_content` is replayed across provider/protocol boundaries
### Basic checks

### What's broken?

A provider-tool turn (`with_provider_tools(:web_search)`, `:code_execution`, etc.) keeps
the provider's native wire blocks in `Message#raw_content` so the same protocol can
replay the turn exactly on the next request. Anthropic's `format_message`,
the Responses protocol's `format_assistant_items` / `format_tool_items`, and Gemini's
`format_parts` all take an unconditional branch:

```ruby
# lib/ruby_llm/protocols/anthropic/chat.rb
def format_message(msg, ...)
  if msg.role == :assistant && msg.raw_content
    format_raw_assistant_message(msg, caching:)   # replays msg.raw_content verbatim
  ...
```

```ruby
# lib/ruby_llm/protocols/responses/chat.rb
def format_assistant_items(msg)
  return msg.raw_content if msg.raw_content   # replays msg.raw_content verbatim
  ...
```

```ruby
# lib/ruby_llm/protocols/gemini/chat.rb
def format_parts(msg)
  if msg.role == :assistant && msg.raw_content
    msg.raw_content   # replays msg.raw_content verbatim
  ...
```

None of these branches check which provider or protocol produced the message.
When a conversation moves to another provider with `with_model` (or a
configured fallback), or to another protocol within the same provider, the
source protocol's blocks are inserted into the new request unchanged.

### How to reproduce

These are deterministic request-rendering reproductions using representative
provider-native blocks. They do not call a provider API.

```ruby
entry = RubyLLM::Accounting::Usage::Entry.new(
  operation: :chat, provider: "anthropic", model: "claude-haiku-4-5", status: :succeeded
)

raw_content = [
  { "type" => "server_tool_use", "id" => "srvtoolu_1", "name" => "web_search",
    "input" => { "query" => "latest Ruby" } },
  { "type" => "web_search_tool_result", "tool_use_id" => "srvtoolu_1", "content" => [] },
  { "type" => "text", "text" => "Ruby 3.x is the latest stable line." }
]

message = RubyLLM::Message.new(
  role: :assistant, content: "Ruby 3.x is the latest stable line.",
  model: "claude-haiku-4-5", raw_content:, usage_entries: [entry]
)

chat = RubyLLM.chat(model: "claude-haiku-4-5", provider: :anthropic)
chat.add_message(role: :user, content: "What is the latest Ruby?")
chat.add_message(message)
chat.add_message(role: :user, content: "And now?")

payload = chat.with_model("gpt-5-nano", provider: :openai).render
payload[:input]
```

The same-provider protocol switch is Vertex AI Claude (Anthropic protocol) to
Vertex AI Gemini (Gemini protocol):

```ruby
raw_content = [
  { "type" => "server_tool_use", "id" => "srvtoolu_1", "name" => "web_search",
    "input" => { "query" => "latest Ruby" } },
  { "type" => "text", "text" => "Ruby 3.x is the latest stable line." }
]

message = RubyLLM::Message.new(
  role: :assistant,
  content: "Ruby 3.x is the latest stable line.",
  model: "claude-haiku-4-5",
  raw_content:,
  raw_content_protocol: :anthropic,
  usage_entries: [
    RubyLLM::Accounting::Usage::Entry.new(
      operation: :chat, provider: "vertexai", model: "claude-haiku-4-5", status: :succeeded
    )
  ]
)

chat = RubyLLM.chat(model: "claude-haiku-4-5", provider: :vertexai, protocol: :anthropic)
chat.add_message(role: :user, content: "Search for the latest Ruby version.")
chat.add_message(message)
payload = chat.with_model("gemini-2.5-flash", provider: :vertexai, protocol: :gemini).render
payload[:contents]
```

The baseline ignored `raw_content_protocol`, so Gemini rendering included the
Anthropic block in its parts. On the patched branch, that marker lets Gemini
render the normalized assistant text instead.

### Expected behavior

When provider or protocol provenance proves the raw blocks foreign, replay the
normalized content, attachments, and RubyLLM tool calls through the target
protocol's existing renderer. Do not treat provider-native `citations` or
`server_tool_calls` as portable wire history; those message fields are not the
normalized request transcript. Preserve exact raw replay when provider and
protocol match. If stripping foreign blocks would leave no normalized content,
attachment, or local tool call, raise a clear error rather than silently
dropping the turn.

### What actually happened

The local rendered `payload[:input]` contains Anthropic's native `server_tool_use` /
`web_search_tool_result` blocks unchanged, sent to the OpenAI Responses
endpoint:

```ruby
[
  { role: "user", content: "What is the latest Ruby?" },
  { "type" => "server_tool_use", "id" => "srvtoolu_1", "name" => "web_search",
    "input" => { "query" => "latest Ruby" } },
  { "type" => "web_search_tool_result", "tool_use_id" => "srvtoolu_1", "content" => [] },
  { "type" => "text", "text" => "Ruby 3.x is the latest stable line." },
  { role: "user", content: "And now?" }
]
```

The reverse direction reproduces the same way in the renderer: an OpenAI
Responses turn's native items are inserted into an Anthropic request's
`content` array once the chat moves to Anthropic.

## Same-provider control

Same-provider and same-protocol replay must remain exact. Provider identity by
itself is insufficient: Vertex AI can select Anthropic for Claude and Gemini
for Gemini, so that protocol switch must drop the old raw blocks.

### Environment

- Baseline commit: `fc66b9f0ab403fcaec630e0773f44ece6f078015`
- Ruby: reproduced identically on both 3.3.5 and 4.0.2
- Rails: not required for the renderer reproduction; persistence is covered by a dummy-app spec
- Provider/protocol cases: Anthropic → OpenAI Responses, OpenAI Responses → Anthropic, and Vertex AI Anthropic → Vertex AI Gemini
- Reproduction: deterministic local request rendering against baseline commit `fc66b9f0ab403fcaec630e0773f44ece6f078015`
- Live API call or production reproduction: not run / not claimed

## Related architecture

Provider identity is available from successful usage entries, but it cannot
distinguish two protocols selected by the same provider. New raw-content
messages therefore need source-protocol provenance. Existing persisted rows
have no such marker; do not infer their source protocol from the model ID.
Their same-provider protocol-switch behavior remains unresolved unless the
original protocol is known through another trusted source.

yorzi:
## What this does

Fixes #948 

A provider-tool turn keeps the producing provider's native content blocks in
`Message#raw_content` so that provider can replay the turn exactly (Anthropic
`server_tool_use` / `web_search_tool_result`, the Responses protocol's
`web_search_call` / reasoning / `message` items, Gemini's server-tool parts).
`format_message` (Anthropic), `format_assistant_items` / `format_tool_items`
(Responses), and `format_parts` (Gemini) all replayed those blocks whenever
`raw_content` was present, with **no check for which provider or protocol
produced them**. Once a chat moved to another provider with `with_model` (or
a configured fallback), the previous provider's wire-format blocks were
inserted into the new provider's request verbatim.

This mirrors the `thinking` cross-provider replay bug fixed in #935. That fix
identified the producer from the message's successful `ruby_llm_usage_entries`
and stripped foreign `thinking` before rendering. This PR adds the same guard
for `raw_content` — and goes one step further than a pure provider check,
because provider identity alone is not always enough.

### Why provider identity alone isn't enough

Vertex AI serves Claude, Gemini, and Chat-Completions-compatible models
through **one provider slug** (`vertexai`) but **three different wire
protocols** (`VertexAI::Anthropic`, `VertexAI::Gemini`,
`VertexAI::ChatCompletions`), selected per model. A message produced on a
Claude-on-Vertex turn and a message produced on a Gemini-on-Vertex turn both
report `provider: "vertexai"` in their usage entry — so a producer-only check
(matching the merged `thinking` fix's granularity) would treat a
Claude-shaped `raw_content` moving to a Gemini-on-Vertex turn as "same
producer" and still replay Anthropic content blocks into Gemini's `parts`.
This is caught with real, tested reproduction in this PR's regression suite.

### The fix

`Message` gains `raw_content_protocol` — the registered name of the protocol
that produced `raw_content` (`"anthropic"`, `"responses"`, `"gemini"`, etc.).
`Provider#complete`, `#tool_approval_response`, `#compact`, and
`Batch#deliver_message` tag it with the resolved protocol's name (reusing the
existing `batch_protocol_name` reverse lookup already used to name protocols
for batch storage — no new naming scheme).

`Protocol#preprocess_message` drops `raw_content` (and, as before, `thinking`)
when the message is foreign by **either** signal:

```ruby
def foreign_protocol_state?(message)
  producer = producer_slug(message)
  return true if producer && producer != @provider.slug

  source_protocol = message.raw_content_protocol
  target_protocol = @provider.batch_protocol_name(self.class)
  source_protocol && target_protocol && source_protocol != target_protocol
end
```

An unknown producer *and* an untagged protocol (a message from before this
change, or one built by hand without `raw_content_protocol:`) both replay
unchanged — the same "preserve current behavior when unsure" rule already
established for `thinking`. This is required for backward compatibility:
existing persisted rows do not carry a protocol tag, and this PR does not
retroactively guess one from the model ID.

### Raising instead of silently losing a turn

If dropping `raw_content` would leave nothing to continue from — no
normalized text, no attachments, no function tool calls — the turn would
otherwise silently vanish from the target request. `preprocess_message` now
raises instead:

```ruby
def ensure_replayable_raw_content!(message)
  return if message.content.to_s.strip != '' || message.attachments.any? || message.tool_call?

  raise Error, "Cannot replay provider-specific assistant state through #{@provider.name}: ..."
end
```

`server_tool_calls` (the normalized record of what a provider tool did) is
deliberately not checked here: no renderer serializes it back onto the wire,
so its presence alone would not make the message replayable.

### Rails persistence

`raw_content`'s existing `jsonb`/`json` column has no `raw_content_protocol`
column, so this PR does not add a migration. Instead,
`Message#raw_content_for_storage` wraps the tagged value in a small envelope
only when a protocol is known:

```ruby
{ "__ruby_llm_raw_content_storage__" => { "version" => 1, "protocol" => "anthropic" }, "content" => raw_content }
```

`Message.raw_content_from_storage` unwraps it on `to_llm`, and falls back to
treating the column as bare `raw_content` with no protocol when the shape
doesn't match — the legacy pre-upgrade case. A message with no protocol tag
falls back to producer-only detection, which existing rows already have
(their usage entries were already persisted).

### Before

```ruby
payload[:input]
# => [..., {"type"=>"server_tool_use", "id"=>"srvtoolu_1", ...},
#     {"type"=>"web_search_tool_result", ...}, {"type"=>"text", "text"=>"Ruby 3.x is the latest stable line."}, ...]
```

### After

```ruby
payload[:input]
# => [..., { role: "assistant", content: [{ type: "output_text", text: "Ruby 3.x is the latest stable line." }] }, ...]
```

### Controls

- Same-provider, same-protocol replay is unchanged: Anthropic → Anthropic and
  OpenAI Responses → OpenAI Responses still replay `raw_content` exactly, so
  Anthropic's encrypted search-result state still round-trips as required.
- Same provider, different protocol (Claude-on-Vertex → Gemini-on-Vertex) now
  also drops foreign `raw_content` — this is the gap a provider-only check
  would miss.
- A message with no known producer, or a protocol-tagged message replayed
  through an untagged (legacy) path, still replays unchanged.
- The producer is never guessed from `message.model` or a shared model id —
  only from `ruby_llm_usage_entries`; the protocol is never guessed from the
  model id either — only from the tag set at creation time.
- Foreign `thinking` and foreign `raw_content` on the same message are both
  handled together.
- A foreign turn with no normalized content, attachments, or tool calls left
  raises a clear error instead of silently disappearing from the request.
- Persisted (`acts_as_chat`) messages keep their stored `raw_content` (and
  protocol tag) after reload; only the per-request copy used to build the
  payload is sanitized. Rows written before this change keep replaying
  exactly within the same provider — verified with a dedicated regression
  spec.
- Tool-result messages are unaffected: `foreign_raw_content?` only inspects
  `role == :assistant`, matching `foreign_thinking?`'s existing scope and the
  plan's guidance not to guess a tool-result message's provenance in this
  first pass.
- Batch-delivered messages (`RubyLLM::Batch#messages`) are tagged too, since
  they never pass through `Provider#complete`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Quality check

- [x] I ran `overcommit --run` and all hooks pass except one pre-existing,
      unrelated failure: `spec/ruby_llm/protocols/vertexai/ranking_spec.rb`
      requires local Google Application Default Credentials and fails the
      same way on `main` before this change.
  - [ ] For provider changes: no wire format changed; nothing to re-record
        with `bundle exec rake vcr:record[provider_name]`
  - [x] All tests pass: `bundle exec rspec --tag "~live"` — 3362 examples, 0
        failures, verified on both Ruby 3.3.5 and Ruby 4.0.2
- [x] I updated documentation if needed — updated the cross-provider replay
      note in `docs/_core_features/provider-tools.md` to describe protocol
      provenance, the raise-instead-of-drop behavior, and the legacy-row
      fallback.

### Regression coverage

- `spec/ruby_llm/chat_raw_content_replay_spec.rb` — Anthropic → OpenAI and
  OpenAI → Anthropic drop foreign `raw_content`; same-provider/same-protocol
  controls; **same-provider, different-protocol** (Vertex AI Claude → Vertex
  AI Gemini) drops it too; unknown producer; no guessing from a shared model
  id; foreign thinking and foreign `raw_content` together; raising on a
  raw-only turn with nothing to continue from; *not* raising when a function
  tool call or an attachment remains; the original transcript stays untouched.
- `spec/ruby_llm/active_record/chat_methods_thinking_replay_spec.rb` — a
  persisted message with Anthropic server-tool `raw_content` drops it when
  the chat moves to OpenAI while the stored row keeps it; a **legacy row
  persisted before this change** (no protocol tag) still replays unchanged
  when the chat continues with the same provider; a protocol-tagged row
  drops `raw_content` after a same-provider protocol switch, and the stored
  envelope shape is asserted directly.
- `spec/ruby_llm/message_spec.rb` — `raw_content_protocol` round-trips
  through `to_h`/`JSON`; `raw_content_for_storage`/`.raw_content_from_storage`
  round-trip with and without a protocol tag, read a legacy (unwrapped) row,
  read a `nil` column, and don't misclassify a coincidentally similar-shaped
  Hash as the storage envelope.
- `spec/ruby_llm/provider_spec.rb` — `#complete`, `#tool_approval_response`,
  and `#compact` each tag their result's `raw_content_protocol` with the
  resolved protocol's name.
- `spec/ruby_llm/batch_spec.rb` — `Batch#messages` tags delivered messages
  with the batch's protocol.
- `spec/ruby_llm/chat_fallbacks_spec.rb` — a fallback to a different provider
  preprocesses history for the fallback's provider, dropping foreign
  `raw_content` from the request sent to it.

Every new cross-provider/cross-protocol/raise assertion above was verified
red before the fix and confirmed to fail again when the corresponding guard
is manually disabled (checked branch-by-branch during review), so none of
them pass vacuously. The full focused list and the complete `--tag "~live"`
suite were both run under Ruby 3.3.5 and Ruby 4.0.2 with identical results.

### Validation

| Check | Result |
|---|---|
| `bundle exec rspec spec/ruby_llm/chat_raw_content_replay_spec.rb spec/ruby_llm/message_spec.rb spec/ruby_llm/provider_spec.rb spec/ruby_llm/batch_spec.rb spec/ruby_llm/chat_fallbacks_spec.rb spec/ruby_llm/chat_thinking_replay_spec.rb spec/ruby_llm/active_record/chat_methods_thinking_replay_spec.rb` (Ruby 3.3.5 and 4.0.2) | 154 examples, 0 failures on each |
| `bundle exec rspec --tag "~live"` (Ruby 3.3.5 and 4.0.2) | 3362 examples, 0 failures on each |
| `bundle exec rubocop --cache false` | 706 files inspected, no offenses |
| `bundle exec archspec check` | 328 files, 1495 constants, 24400 facts checked |
| `overcommit --run` (RSpec hook skipped for the known ADC gap) | Gitleaks, TrailingWhitespace, RuboCop, ArchSpec, AppraisalUpdate, Flay all OK |
| `git diff --check` | clean |

## AI-generated code

## API changes

- [ ] Breaking change
- [x] No API changes

`Message#raw_content_protocol`, `#raw_content_for_storage`, and
`.raw_content_from_storage` are `:nodoc:`, mirroring `#raw_content` and
`#without_thinking` — internal, not part of the public API. A message built
by hand with `raw_content:` and no `raw_content_protocol:` behaves exactly as
before.

One user-visible behavior change worth flagging for maintainer review: a
provider-tool-only assistant turn (no visible text, no attachments, no
function tool calls) that previously replayed silently across a foreign
provider or protocol now raises `RubyLLM::Error` instead. This is intentional
(see "Raising instead of silently losing a turn" above) but is a new failure
mode for an edge case that previously "worked" by accident.


crmne:
Thanks but this is the wrong shape.

1. The information about provider and protocol is already in the message's model. Adding a new storage for it is not needed.
2. This shouldn't run on replay, should run at model change.

Closing.

PR #956, opened 2026-09-22
Render partials inside prompt templates

### LINKED ISSUE #955: [FEATURE] Partials in prompt templates
### Scope check

### Due diligence

Follow-up to #675.

In practice app/prompts templates share pieces: tone, safety rules, etc. The way to reuse them today is inline Ruby:

<%= RubyLLM.render_prompt("shared/tone", name: name) %>

It works, but it does not know where the current prompt lives, so an agent cannot keep partials next to its own instructions.txt.erb. And anyone who has written a Rails view reaches for render here.

### Proposed solution

A render helper inside templates, with ActionView rules:

<%= render "tone" %>
<%= render "shared/safety", product_name: product_name %>

- render `shared/safety` maps to `shared/_safety.txt.erb`. The underscore keeps partials out of agent default lookups.
- Lookup starts in the current prompt's directory, then goes through RubyLLM::Prompt.roots, so engine partials and app overrides work the same way prompts do now.
- Locals are explicit, nothing leaks from the outer template.
- Missing partial raises PromptNotFoundError.

The name is `render` because that is what it is called in views; it lives only inside templates and does not touch `Chat#render` or `render_prompt`.  If a bare `render` feels too overloaded, `partial "tone"` is my fallback, though Rails intuition reaches for `render` first.

Out of scope: collections and layouts.

Happy to send the PR.

The helper needs the current prompt's location and the lookup roots, and only RubyLLM:Prompt has both. Without it every app and engine writes its own wrapper for the same thing

kryzhovnik:
## What this does

Adds a `render` helper inside prompt templates so a template can insert a partial, with ActionView rules:

```erb
<%= render "tone" %>
<%= render "shared/safety", product_name: product_name %>
```

- `render "shared/safety"` maps to `shared/_safety.txt.erb`.
- Lookup starts in the current prompt's directory, then goes through `RubyLLM::Prompt.roots`, so engine partials work the same way prompts do.
- Locals are explicit, nothing leaks from the outer template.
- A missing partial raises `PromptNotFoundError`.

Templates now evaluate in `RubyLLM::Prompt::Context` instead of `ERB#result_with_hash`. The class is declared with its full name on purpose: a nested declaration would put `RubyLLM` in the template's lexical scope, and `Chat` in a template would resolve to `RubyLLM::Chat` instead of the application's model. There is a regression spec for that, and the `Style/ClassAndModuleChildren` cop is disabled on that one line.

No new methods on `RubyLLM` or `Prompt`. Docs are left out of this PR on purpose; happy to add a Partials section to the prompt rendering guide if you prefer it here.

Closes #955

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Performance improvement

## Scope check

## Required for new features

- [x] Linked issue: #955

## Quality check

## AI-generated code

## API changes

- [ ] Breaking change
- [ ] No API changes


crmne (changes requested):
Thanks, this is great work. The `Prompt::Context` design and the spec that keeps `Chat` pointing at the app's model are exactly right. Two things before we merge:

1. **Lookup order.** Only bare names should resolve relative to the current prompt, as in ActionView. `render "tone"` in `work_assistant/instructions.txt.erb` should find `work_assistant/_tone.txt.erb`. `render "shared/safety"` should resolve from the prompt roots only, as `shared/_safety.txt.erb`. Right now it tries `work_assistant/shared/_safety.txt.erb` first, which can silently pick the wrong file when both exist. A spec with both files present would pin this down.

2. **Docs.** Yes please to the Partials section in `docs/_core_features/prompt-rendering.md`, in this PR. Cover the naming convention, bare names vs. paths, explicit locals, and the missing-partial error.

kryzhovnik:
Path names now resolve from the prompt roots only, with a spec that has both `shared/_safety` and `work_assistant/shared/_safety` present. Added the Partials section to the prompt rendering guide.

